PR #11, opened 2017-01-20
Add math explanation

Atcold:
Add an explanation about `x.grad` values.

### DIFF
--- a/Deep Learning with PyTorch.ipynb
+++ b/Deep Learning with PyTorch.ipynb
@@ -424,6 +424,15 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "You should have got a matrix of `4.5`.\n",
+    "Let's call the `out` *Variable* \"$o$\".  \n",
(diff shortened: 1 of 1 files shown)


apaszke:
Well, these equalities are false. It's like writing `4 * 9 = 36 = 36/2 = 8`. The partial derivative is not equal to a number - either note that you apply it in some point right from the start, or don't put `=` between them.

Atcold:
Hmm... Not quite. `f(x) | x = 1` means `f(x)` computed for the point `x = 1` .
But I change the notation, if you don't like it.

apaszke:
Yeah I know. For example it is valid with definite integrals, because they're actually numbers, so you can write them as functions applied in different points. But partial derivatives are functions and you can't write that you apply them in the middle of equality, because you end up with a function equal to a number.

Atcold:
Fixed. It should be good to go.
Let me know.

Atcold:
Ping.

PR #12, opened 2017-01-21
Fix call in FFI tutorial

SeanNaren:
I had to change the function call (I assume the previous name is deprecated)!


soumith:
thanks!

apaszke:
It's still invalid, because you have to make it:
```python
ffi = create_extension(...)
ffi.build()
```

Issue #10, opened 2017-01-20
Reinforcement demo issue


brad0taylor:
When I run the following line in the cart-pole demo using pytorch 0.1.6 - I get the following error:

> plt.imshow(get_screen().squeeze(0).permute(1, 2, 0).numpy(), interpolation='none')

```
/home/brad/anaconda3/lib/python3.5/site-packages/gym/envs/classic_control/cartpole.py in _render(self, mode, close)
    135             self.viewer.add_geom(self.track)
    136 
--> 137         x = self.state
    138         cartx = x[0]*scale+screen_width/2.0 # MIDDLE OF CART
    139         self.carttrans.set_translation(cartx, carty)

AttributeError: 'CartPoleEnv' object has no attribute 'state'
```

however the rest of the demo seems to work ok 

Also I had to insert  'import torch' in order to avoid a downstream error. 

 Lovely fun demo

apaszke:
Ah! It's probably missing `env.reset()` before that cell. I'll push a fix soon. Thanks!

brad0taylor:
Yes that fixes it. Thanks much

PR #19, opened 2017-01-24
Fix a typo that breaks the file

ikostrikov:


soumith:
just fyi might encounter into this too: https://github.com/pytorch/tutorials/issues/10

PR #23, opened 2017-02-04
More clear and intuitive math explanation

ritchieng:
I believe this is more straightforward compared to the previous one. A small change but intuitive using basic differentiation.


soumith:
thanks!

apaszke:
Maybe clearer, but incorrect... You can't say that `z_i = 3/4(x + 2)^2` and then that `o = 1/4 sum_i z_i`, because you'll divide `x` by 16.

ritchieng:
@apaszke You're right. Actually the existing one is ok considering how you just simplified the fraction.

I revised mine and used the original fraction: https://github.com/ritchieng/tutorials/blob/master/Deep%20Learning%20with%20PyTorch.ipynb

Thanks.

PR #27, opened 2017-02-05
add missing env.reset() before plt.imshow

truskovskiyk:
fix issue https://github.com/pytorch/tutorials/issues/10


apaszke:
Thanks!

PR #28, opened 2017-02-05
Neural-Style tutorial

alexis-jacq:
PyTorch tutorial explaining how to implement the neural transfer of artistic style developed by Leon Gatys and AL. The code parts of the tutorial can be tested in a whole script available here: https://github.com/alexis-jacq/Pytorch-Tutorials/blob/master/Neural_Style.py.


chsasank:
Hi @alexis-jacq,

Can you please fix the issues [mentioned here](https://discuss.pytorch.org/t/pytorch-tutorial-for-neural-transfert-of-artistic-style/336) in your notebook. 

Also, can you please use a switch for cuda like in below, so that the notebook can be run seamlessly on all machines?

```python
use_cuda = torch.cuda.is_available()
dtype =  torch.cuda.FloatTensor if use_cuda else torch.FloatTensor

tens = torch.rand(30).type(dtype)

model = torchvision.models.alexnet()
if use_cuda:
     model  = model.cuda()
```

Once that is done, we will convert the notebook to sphinx gallery python file.

Sasank.

alexis-jacq:
Thank you for taking my tutorial into consideration!

I am doing the modification you suggest. 

I am also modifying the code (and the notebook) to use a "feature extractor" approach (as [suggested here](https://discuss.pytorch.org/t/how-to-extract-features-of-an-image-from-a-trained-model/119/13?u=alexis-jacq)).

alexis-jacq:
Please find [here](https://github.com/alexis-jacq/Pytorch-Tutorials) the corrected version (.py and .ipynb) of my tutorial. 

Finally, I preferred to keep my approach that is much easier to understand/explain in a beginner tutorial since there are no (so far) trivial ways to extract features properly. What I am doing is just a translation of https://github.com/jcjohnson/neural-style in Pytorch, using exactly the same approach of implementation.

chsasank:
Hi,

Just going through your notebook. You still have code blocks like 
```
cnn = models.alexnet(pretrained=True).features.cuda()
...
art_net = nn.Sequential().cuda() # the new Sequential module network
gram = GramMatrix().cuda() # we need a gram module in order to compute style targets
...
content_loss = ContentLoss(target, content_weight).cuda()
```
Which fail on my cpu only machine.

alexis-jacq:
Sorry, they slipped through my attention...
The [last version](https://github.com/alexis-jacq/Pytorch-Tutorials) should work on cpu now.

chsasank:
But there's still .cuda !
```
target = art_net.forward(content.cuda()).clone()
```

Please create a new environment without cuda like this:
```
conda create --name cpu jupyter numpy
source activate cpu
conda install pytorch torchvision -c south
```
Then make sure `which jupyter` is the right env path and run the jupyter notebook.

Please test in this environment.

alexis-jacq:
My mistake, I didn't update the notebook correctly. Now it is ok. 
Thanks for the cpu environment tips !

chsasank:
Thanks for fixing cpu issue. 
I have run the notebook. But the second run of the style transfer also gives me noise like this:
unknown

Some requests regarding style:
1. Please use pep8. Please make maximum line width 80. You can use `autopep8`.
2. Use python3 print function syntax.

alexis-jacq:
I used conv_3 for content loss, content_weight = 1 and style_weight = 100 to obtain the figure I show. 
I corrected it in both the notebook and the example code. Now you should obtain something similar to the figure in the notebook.

I also did the modifications you suggest (pep8, 80-max lines and python3 print)

chsasank:
Hi,

Here's the output I got now: 
unknown

I have run this on CPU. May be it's to do with that?

Sasank.

chsasank:
Would it be possible to adjust the parameters in your tutorial to get results as stunning as in [here](https://github.com/leongatys/PytorchNeuralStyleTransfer/blob/master/NeuralStyleTransfer.ipynb
)?

When the tutorial is added to site, it'll be executed and images will be shown. It'll be nice if results look great.

alexis-jacq:
This is weird, it works on my computer even on CPU. Is it also the second run? What do you obtain if you start from the input image instead of the gaussian noise? 

I can adjust parameters to get nicer images, but it in order to get results like [here](https://github.com/leongatys/PytorchNeuralStyleTransfer/blob/master/NeuralStyleTransfer.ipynb) it must be run with VGG as a model. I can add a line to explain how to get a pretrained VGG and to use it with the code of the tutorial (to get the final picture).

alexis-jacq:
And does it work if you run the .py code?

chsasank:
Here are the outputs of my runs:
```
$ python Neural_Style.py
```
figure_1

Running notebook as it is. I.e this cell remains as it is
```
input = image_loader("images/dancing.jpg").type(dtype)
# if we want to fill it with a white noise:
input.data = torch.randn(input.data.size()).type(dtype)

# add the original input image to the figure:
plt.subplot(223)
imshow(input.data)
```
First run
unknown-3

Second run

unknown-4

Next, input is content image. ie. cell looks like this:

```
input = image_loader("images/dancing.jpg").type(dtype)
# if we want to fill it with a white noise:
# input.data = torch.randn(input.data.size()).type(dtype)

# add the original input image to the figure:
plt.subplot(223)
imshow(input.data)
``` 

First run
unknown-2

Second run
unknown

I guess there's some bug in the notebook.

> I can adjust parameters to get nicer images, but it in order to get results like here it must be run with VGG as a model. I can add a line to explain how to get a pretrained VGG and to use it with the code of the tutorial (to get the final picture).

Then may be we should directly use VGG and get that final picture? It looks much more attractive that way. It shouldn't be much of a change in the code right?

alexis-jacq:
Done ! In fact, since pytorch now provide pretrained vgg, just changing the line by 
"cnn = models.vgg19(pretrained=True).features"  did it. 

I also moved to LBFGS optimizer as suggested by Leon Gatys. Now, whatever you start with white noise or content image, the result is nice. 

As you are going to test it with CPU, don't hesitate to decrease the number of iterations: it is much slower with VGG, but the result does not require several iterations (100 may be enough)

chsasank:
Hi, 

I figured the issue of bad outputs. I fixed that and made quite a few changes to the notebook. I have converted the notebook to format required by the repo and put it here: https://github.com/chsasank/tutorials/tree/neural-style

Here's a summary of changes I've made:
1. I converted chunks of codes to functions with proper parameters. This makes inputs and outputs of a given cell very clear.
2. I removed the run without clamping. Tutorial is already long and those two cells had almost same code. So better remove it.
3. Instead of subplots, I have 4 named figures. It looks good that way on web page.
4. Lot of cosmetic fixes here and there. 

Everything in my repo has been tested. 

What I propose for you to do:
1. Fork my repo, use neural-style branch
2. Preview your tutorial by building using `make html-noplot`
3. Make a few changes if any
4. Raise a new PR with my forked repo.


Cheers,
Sasank.

PR #30, opened 2017-02-07
Fixed clamp not being applied to parameters

farzadab:
I'm just learning about PyTorch, but I think the line `param.grad.data.clamp(-1, 1)` has no effect. I guess you wanted to do it *in-place*, which is `clamp_` if I'm not mistaken.


apaszke:
Yeah, my bad 😕 Thanks for the fix!

farzadab:
Thanks for the fast response! :)

PR #34, opened 2017-02-14
cannot call backward on one variable twice without turnning on retain…

eriche2016:
… varaible  flag


soumith:
i think added in other PR

Issue #36, opened 2017-02-20
Reinforcement Q-learning issue on GPU


yzli:
when running on GPU,  torch.cuda.is_available() == True, I run into the following problem, in training loop:

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-76-497fbef317fd> in <module>()
     50     for t in count():
     51         # Select and perform an action
---> 52         action = select_action(state)
     53         _, reward, done, _ = env.step(action[0,0])
     54         reward = torch.Tensor([reward])

<ipython-input-75-a53cf7c104da> in select_action(state)
     27     steps_done += 1
     28     if sample > eps_threshold:
---> 29         return model(Variable(state, volatile=True)).data.max(1)[1].cpu()
     30     else:
     31         return torch.LongTensor([[random.randrange(2)]])

/home/ubuntu/.local/lib/python2.7/site-packages/torch/nn/modules/module.pyc in __call__(self, *input, **kwargs)
    200 
    201     def __call__(self, *input, **kwargs):
--> 202         result = self.forward(*input, **kwargs)
    203         for hook in self._forward_hooks.values():
    204             hook_result = hook(self, input, result)

<ipython-input-73-58135f60e01c> in forward(self, x)
     11 
     12     def forward(self, x):
---> 13         x = F.relu(self.bn1(self.conv1(x)))
     14         x = F.relu(self.bn2(self.conv2(x)))
     15         x = F.relu(self.bn3(self.conv3(x)))

/home/ubuntu/.local/lib/python2.7/site-packages/torch/nn/modules/module.pyc in __call__(self, *input, **kwargs)
    200 
    201     def __call__(self, *input, **kwargs):
--> 202         result = self.forward(*input, **kwargs)
    203         for hook in self._forward_hooks.values():
    204             hook_result = hook(self, input, result)

/home/ubuntu/.local/lib/python2.7/site-packages/torch/nn/modules/conv.pyc in forward(self, input)
    235     def forward(self, input):
    236         return F.conv2d(input, self.weight, self.bias, self.stride,
--> 237                         self.padding, self.dilation, self.groups)
    238 
    239 

/home/ubuntu/.local/lib/python2.7/site-packages/torch/nn/functional.pyc in conv2d(input, weight, bias, stride, padding, dilation, groups)
     35     f = ConvNd(_pair(stride), _pair(padding), _pair(dilation), False,
     36                _pair(0), groups)
[...]

PR #39, opened 2017-02-20
Issue 36: fixing q-learning cuda problems

yzli:
#36 : @apaszke 
Thanks for the quick edits to master! But I still had some problems, typically after a few iterations and when the replay memory is sampled. I made a few more changes that appear to fix the problems; please review. Thanks.


mehdi-shiba:
Thanks for sharing this, it indeed fixed the problem with running this tutorial with CUDA.

soumith:
thanks!

Issue #51, opened 2017-03-26
Broken Simple Tutorial Example


sbodenstein:
I have just installed PyTorch, and this example doesn't work: http://pytorch.org/tutorials/beginner/pytorch_with_examples.html#pytorch-variables-and-autograd

Running the example returns:
```
0 26220132.0
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-44-6cc21ebb6010> in <module>()
     36 
     37   # Manually zero the gradients before running the backward pass
---> 38   w1.grad.data.zero_()
     39   w2.grad.data.zero_()
     40 

AttributeError: 'NoneType' object has no attribute 'data'
```
as `w1.grad.data` doesn't seem to be defined.

chsasank:
Weirdly enough, it runs fine for me. 
How did you run the example?

May be update pytorch and retry?

spro:
You're probably running into the same problem as https://github.com/jcjohnson/pytorch-examples/pull/3/commits/c7e345e30ea09b8800a11f22c16742c07a6a73dd

chsasank:
#52 fixes the issue.

PR #52, opened 2017-03-28
fix the issue #51

fehiepsi:
In PyTorch 1.10, I think that it is recommended to do zero gradients after updating weights?!


chsasank:
Thanks. Merged.

PR #58, opened 2017-04-10
Use __call__ instead of forward

darkstar112358:
Replacing usage of `forward()` with call operators like: `model(input)` which is the suggested usage.  [Code](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/module.py#L205-L222) for details.


soumith:
thanks abhishek!

PR #63, opened 2017-04-12
Fix LSTM model in sequence_models_tutorial.py

### LINKED ISSUE #55: unused hidden variables of LSTM on deep_learning_nlp_tutorial.py examples
Hi,

I am just curious about line [1197](https://github.com/pytorch/tutorials/blob/master/beginner_source/deep_learning_nlp_tutorial.py#L1197). 
```
self.hidden = self.init_hidden()
```

This value has not been used to update parameters of network in the program. But you are saying `also, we need to clear out the hidden state of the LSTM, detaching it from its history on the last instance`. Is it correct?

Best regards,

Varal7:
Fixes #55


antran89:
:+1:

soumith:
good find. thanks!

Issue #68, opened 2017-04-19
Indentation error 


euler16:
The code displayed on the site http://pytorch.org/tutorials/beginner/pytorch_with_examples.html for custom functions has some indentation error. I checked the code given in the repo and there its correct (proper indentation). How should I correct this .
PS I want to begin contributing to open source so maybe the issue is a trivial one. Kindly help me in finding a way to correct it.

chsasank:
Thanks for raising the issue :)

Can you please point out the errors in the page?

euler16:
Under the heading "PyTorch: Defining new autograd functions", inside the forward function of class MyRelu return statement is not in proper level. I found a few more while going through the docs

chsasank:
Thanks for pointing out.

I am also not completely sure why this is happening. 
But here are the two things that might fix the issue. Both are useful anyway though.

- [ ] Make beginners_source/examples_*/*.py pep8 complaint. Python packages `pep8` and `autopep8` should help you. For most of the part, this means changing from 2 spaces to standard 4 spaces. 
- [ ] Custom directive to include source file without top docstring is [here](https://github.com/pytorch/tutorials/blob/master/custom_directives.py#L18). I just realised that regex is substituting every docstring in the file. Two ways to fix this is to use `count=0` in `self.docstring_regex.sub()` or even better, figure out a regex to match only docstring at the top.

I welcome a PR with both the above changes. 

PS: If you are new to the PR business, read about PRs [here](https://help.github.com/articles/about-pull-requests/). Build instructions are in README.

PR #69, opened 2017-04-19
Fix for Indentation error #68

euler16:
This PR is with respect to Issue #68

### DIFF
--- a/beginner_source/nlp/sequence_models_tutorial.py
+++ b/beginner_source/nlp/sequence_models_tutorial.py
@@ -197,8 +197,7 @@ inputs = prepare_sequence(training_data[0][0], word_to_ix)
 tag_scores = model(inputs)
 print(tag_scores)
 
-for epoch in range(
-        300):  # again, normally you would NOT do 300 epochs, it is toy data
+for epoch in range(300):  # again, normally you would NOT do 300 epochs, it is toy data
     for sentence, tags in training_data:
         # Step 1. Remember that Pytorch accumulates gradients.
         # We need to clear them out before each instance
--- a/beginner_source/nlp/sequence_models_tutorial.py
+++ b/beginner_source/nlp/sequence_models_tutorial.py
@@ -75,12 +75,12 @@ for i in inputs:
     out, hidden = lstm(i.view(1, 1, -1), hidden)
 
 # alternatively, we can do the entire sequence all at once.
-# the first value returned by LSTM is all of the hidden states throughout 
-# the sequence. the second is just the most recent hidden state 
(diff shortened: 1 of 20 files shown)


euler16:
This PR is with respect to Issue #68

chsasank:
Does this fix #68 after building?

euler16:
@chsasank   I did what you suggested there (Issue #68 ). The build process was taking a lot of time. I will finish the build and get back to you.

chsasank:
from readme
> Then you can build using make docs. This will download the data, execute the tutorials and build the documentation to docs/ directory. However, this will take about 30-60 min based on your system.
> You can skip the execution by running **make html-noplot** to build html documentation to _build/html. This way, you can quickly preview your tutorial.

euler16:
@chsasank  it works

chsasank:
Thanks for pep8 fixes.

I built your fork using `make html-noplot`and this is how the output for the code in question looks now:

```
# -*- coding: utf-8 -*-
import torch
from torch.autograd import Variable


class MyReLU(torch.autograd.Function):
    
    def forward(self, input):
                self.save_for_backward(input)
        return input.clamp(min=0)

    def backward(self, grad_output):
                input, = self.saved_tensors
        grad_input = grad_output.clone()
        grad_input[input < 0] = 0
        return grad_input
...
```

Neither the indentation bug nor the multiple deletions of docstrings bug got fixed. Please recheck and build.

euler16:
screenshot from 2017-04-20 04-23-10
But in my computer they seem to be corrected

euler16:
I used make html-noplot

custom_directives.py
```diff
@@ -25,7 +25,7 @@ def run(self):
 
         try:
             text = open(filename).read()
-            text_no_docstring = self.docstring_regex.sub('', text)
+            text_no_docstring = self.docstring_regex.sub('', text, count = 0)
```
chsasank: Figured it! `count=1` is the right parameters. (no spaces around `=` because pep8)
Sorry that I've said it's count=0 in the issue.

beginner_source/nlp/sequence_models_tutorial.py
```diff
@@ -193,13 +197,14 @@ def forward(self, sentence):
 tag_scores = model(inputs)
 print(tag_scores)
 
-for epoch in range(300):  # again, normally you would NOT do 300 epochs, it is toy data
+for epoch in range(
+        300):  # again, normally you would NOT do 300 epochs, it is toy data
```
chsasank: Move the comment to above line rather than break range into two lines.

chsasank:
Thanks Nilay! :)

PR #70, opened 2017-04-21
Update requirements.txt

qiuwch:
Add sphinx as a dependency, otherwise, it will report sphinx-build not found.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
+sphinx
 sphinx-gallery
 numpy


qiuwch:
It looks like more dependencies are missing on my computer. I will make another more complete pull request.

chsasank:
Hi, I have not added matplotlib etc to requirements because sphinx-gallery had them in its [requirements.txt](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/requirements.txt). 

Looks like pip is not installing them when installing sphinx. So please add requirements from sphinx gallery to our requirements.txt

PR #71, opened 2017-04-21
Add missing dependencies: sphinx, sphinx_rtd_theme, matplotlib.

qiuwch:
Add missing dependencies.
Tested in a clean docker container.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ sphinx_rtd_theme
 sphinx-gallery
 numpy
 matplotlib
+torchvision
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,5 @@
+sphinx
+sphinx_rtd_theme
 sphinx-gallery
 numpy
+matplotlib


chsasank:
Please add `pillow` to requirements as well just in case.  

from sphinx gallery's read me:
> Sphinx-Gallery will not manage its dependencies when installing, thus you are required to install them manually. Our minimal dependencies are:
> 
> Sphinx
> Matplotlib
> Pillow

Thanks :)

qiuwch:
Pillow is installed during the installation of pytorch and torchvision. That's why I don't include it.

chsasank:
Oh right, then add `torchvision` to requirements. Ideally, torch should be added too, but there's no simple way to install it across OS without conda.

qiuwch:
Ok, done.

chsasank:
Thank a lot @qiuwch !

qiuwch:
My pleasure. This is a very well-organized tutorials repository.

PR #75, opened 2017-04-22
convert string to unicode for 2.7 compatibility

### LINKED ISSUE #72: Unicode error in python2.7
Hi, 
I got the following error while running char_rnn_classification_tutorial.py 
 ```python
print(unicodeToAscii('Ślusàrski'))
  File "src/char_rnn.py", line 17, in unicodeToAscii
    c for c in unicodedata.normalize('NFD', s)
TypeError: normalize() argument 2 must be unicode, not str
```

To circumvent the error I changed the code:
```python
def unicodeToAscii(s):
    try:
        s = unicode(s, 'utf-8')
    except NameError:
        pass
    s = unicodedata.normalize('NFD', s)
    s = s.encode('ascii', 'ignore')
    s = s.decode('utf-8')
    return str(s)
```

bajibabu:
Edit: Fix #72

### DIFF
--- a/intermediate_source/char_rnn_classification_tutorial.py
+++ b/intermediate_source/char_rnn_classification_tutorial.py
@@ -79,6 +79,10 @@ n_letters = len(all_letters)
 
 # Turn a Unicode string to plain ASCII, thanks to http://stackoverflow.com/a/518232/2809427
 def unicodeToAscii(s):
+    try:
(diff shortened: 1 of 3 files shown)


chsasank:
Thanks for the PR @bajibabu, but as it so turns out this doesn't really work. I will create a new PR with slight modified fix suggested by @spro.

PR #79, opened 2017-05-07
Fix errors in beginner/blitz

desktable:
A recent refactoring breaks the blitz tutorial. This PR fixes that.


chsasank:
Thanks @desktable for the PR! Sorry for the late reply. Merging.

PR #81, opened 2017-05-08
Replace changing volatile flag with Variable reconstruction

Kaixhin:
This PR is more of a question on best practices here - is it better to adjust the internal `volatile` flag (most efficient) or extract the `data` and create a new `Variable` (arguably more understandable)? I haven't looked at the new autograd changes, so perhaps there is more to be done to future-proof this as well.


apaszke:
Modifying Variable flags is ok - autograd will make sure you're not doing illegal things. Reusing the same tensor in multiple Variables is not supported (we depend on it in certain places).

PR #82, opened 2017-05-14
make some variable and function names more accurate in intermediate RNN tutorials

greaber:
Previously, processing a single training example was referred to as an "epoch".  I changed this to "step".  A downside of the change is that sometimes "step" is used to refer to a single step of the RNN in processing a training example.  But I don't think there will be too much confusion.

Previously, a function called "randomTrainingSet" in the generation tutorial returned the prepared tensors for a single training example.  I changed the name to "randomTrainingExample".

Previously, a function called "randomTrainingPair" in the classification tutorial returned the prepared tensors for a single training example.  I changed the name to "randomTrainingExample".  The old name also clashed with the use of the name "randomTrainingPair" in the generation tutorial.

A small remaining inconsistency is that both the classification and generation tutorials define functions called "train" but this function is invoked (after my name change) "train(*randomTrainingExample())" in the generation tutorial but cannot be invoked this way in the classification tutorial.

### DIFF
--- a/intermediate_source/char_rnn_classification_tutorial.py
+++ b/intermediate_source/char_rnn_classification_tutorial.py
@@ -274,7 +274,7 @@ import random
 def randomChoice(l):
     return l[random.randint(0, len(l) - 1)]
 
-def randomTrainingPair():                                                                                                               
+def randomTrainingExample():
     category = randomChoice(all_categories)
     line = randomChoice(category_lines[category])
     category_tensor = Variable(torch.LongTensor([all_categories.index(category)]))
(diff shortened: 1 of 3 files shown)


intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -364,19 +364,19 @@ def timeSince(since):
 
 start = time.time()
 
-for epoch in range(1, n_epochs + 1):
```
spro: `step` could get confusing considering we're using RNNs with multiple "time steps"... maybe `iter` and `n_iters`?

greaber: OK, I can change it to be that way.  Also, I was thinking to delete the comment that says you don't understand why initializing the loss to zero works since I don't think it is good to give people more things to feel confused about, and anyway I just assumed pytorch allows you to create a Variable by adding a python constant to a Variable, which doesn't seem mysterious, or am I missing something?

spro: Agreed

chsasank:
@greaber Is this PR ready to merge?

greaber:
Yes.

chsasank:
Thanks a lot @greaber

Issue #66, opened 2017-04-14
 'TimeLimit' object has no attribute 'x_threshold' error running reinforcement_q_learning.ipynb tutorial


NataliaDiaz:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-5-2f6091ee96f6> in <module>()
     37 
     38 env.reset()
---> 39 plt.imshow(get_screen().squeeze(0).permute(
     40     1, 2, 0).numpy(), interpolation='none')
     41 plt.show()

<ipython-input-5-2f6091ee96f6> in get_screen()
     19     screen = screen[:, 160:320]
     20     view_width = 320
---> 21     cart_location = get_cart_location()
     22     if cart_location < view_width // 2:
     23         slice_range = slice(view_width)

<ipython-input-5-2f6091ee96f6> in get_cart_location()
      8 
      9 def get_cart_location():
---> 10     world_width = env.x_threshold * 2
     11     scale = screen_width / world_width
     12     return int(env.state[0] * scale + screen_width / 2.0)  # MIDDLE OF CART

AttributeError: 'TimeLimit' object has no attribute 'x_threshold'
```

chsasank:
Duplicate of #29. Fixing.

chsasank:
Please download and report if you have any errors. Thanks.

NataliaDiaz:
Thanks, now I get : free(): invalid pointer error running both the notebook and reinforcement_q_learning.py (the kernels dies and restarts):

```
~/notebooks$ jupyter notebook
[I 09:22:24.158 NotebookApp] Serving notebooks from local directory: /home/natalia/notebooks
[I 09:22:24.158 NotebookApp] 0 active kernels 
[I 09:22:24.158 NotebookApp] The Jupyter Notebook is running at: http://localhost:8888/?token=e70854c1d5423883eda6325c1b503efbebafff06ad9e0177
[I 09:22:24.158 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[C 09:22:24.160 NotebookApp] 
    
    Copy/paste this URL into your browser when you connect for the first time,
    to login with a token:
        http://localhost:8888/?token=e70854c1d5423883eda6325c1b503efbebafff06ad9e0177
Created new window in existing browser session.
[I 09:22:24.727 NotebookApp] Accepting one-time-token-authenticated connection from 127.0.0.1
[...]

PR #83, opened 2017-05-15
fix bug when using cuda in ubuntu

WorldHellooo:
The Class Variable's __init__() doesn't return a cuda Variable because the 'data' would be released later. This bug results in free(): invalid pointer error.

### DIFF
--- a/intermediate_source/reinforcement_q_learning.py
+++ b/intermediate_source/reinforcement_q_learning.py
@@ -310,12 +310,12 @@ if USE_CUDA:
     model.cuda()
 
 
-class Variable(autograd.Variable):
+def Variable(data, volatile=False):
+    if USE_CUDA:
+        return autograd.Variable(data.cuda(),volatile=volatile)
+    else:
+        return autograd.Variable(data, volatile=volatile)
 
-    def __init__(self, data, *args, **kwargs):
-        if USE_CUDA:
-            data = data.cuda()
-        super(Variable, self).__init__(data, *args, **kwargs)
 
 
 steps_done = 0


apaszke:
How is that different from the previous solution (except that it always produces volatile variables which is incorrect)?

WorldHellooo:
The previous solution returns a cpu Variable even if the USE_CUDA is true. As a result, the data is in the cpu while the net model is in the gpu, causing a invalid pointer error.

apaszke:
No, it doesn't. The data is converted to CUDA

WorldHellooo:
Could you please tell me what's wrong with my experiment?
screenshot from 2017-05-15 20 25 14

WorldHellooo:
Could you tell me what's wrong with my experiment?
screenshot

apaszke:
Ah I now see what's wrong. Variables ignore `__init__`. In fact now I'd consider modifying these classes bad practice and the example should be modified. However your PR doesn't fix the problem.

WorldHellooo:
Ok, it doesn't matters.

chsasank:
Hi, 

@WorldHellooo does this PR fix #66 ?
@apaszke What changes do you recommend? I'll make those changes.

WorldHellooo:
In my test, it did fix #66.

apaszke:
It might fix it, but it's not a correct patch. We should use this hack but call `type`

chsasank:
Cool, I'll create a PR like this
```
use_cuda = torch.cuda.is_available()
dtype = torch.cuda.FloatTensor if use_cuda else torch.FloatTensor
.
.
.

var = Variable(data).type(dtype)
```

PR #90, opened 2017-05-26
rm useless init_hidden in BiLSTM_CRF

cmusjtuliuyuan:
These two init_hidden are useless. Because they will be called in the _get_lstm_features function.


soumith:
thanks!

PR #95, opened 2017-06-02
fix self.init_hidden never used

cmusjtuliuyuan:


beginner_source/nlp/advanced_tutorial.py
```diff
@@ -180,8 +180,8 @@ def __init__(self, vocab_size, tag_to_ix, embedding_dim, hidden_dim):
         self.hidden = self.init_hidden()
 
     def init_hidden(self):
-        return (autograd.Variable(torch.randn(2, 1, self.hidden_dim)),
-                autograd.Variable(torch.randn(2, 1, self.hidden_dim)))
+        return (autograd.Variable(torch.randn(2, 1, self.hidden_dim / 2)),
```
chsasank: Can you please use `self.hidden_dim // 2`?

beginner_source/nlp/advanced_tutorial.py
```diff
@@ -217,7 +217,7 @@ def _forward_alg(self, feats):
     def _get_lstm_features(self, sentence):
         self.hidden = self.init_hidden()
         embeds = self.word_embeds(sentence).view(len(sentence), 1, -1)
-        lstm_out, self.hidden = self.lstm(embeds)
+        lstm_out, self.hidden = self.lstm(embeds, self.hidden)
```
chsasank: Thanks for catching this! I wonder how it worked at all. Documentation says `nn.LSTM` expects `(input, (h_0, c_0))`. I should have raise an error, no?

beginner_source/nlp/advanced_tutorial.py
cmusjtuliuyuan: fix it

beginner_source/nlp/advanced_tutorial.py
cmusjtuliuyuan: Look at this line: https://github.com/pytorch/pytorch/blob/e694db0eeb3ff380b476e92710475efc4551c826/torch/nn/modules/rnn.py#L61

cmusjtuliuyuan: By this definition, lots of lines of code in "nlp/sequence_models_totorial" are useless too.
 
https://github.com/pytorch/tutorials/blob/master/beginner_source/nlp/sequence_models_tutorial.py#L170

jekbradbury: Yes, this was formerly required but PyTorch LSTMs have had automatic initialization with zeros since February or so.

chsasank: @cmusjtuliuyuan init_hidden is still required to reset the model. init_hidden can now just `return None`.

But I guess we'll leave it as it is to make things clear and explicit.

chsasank:
Thanks @cmusjtuliuyuan!

PR #99, opened 2017-06-15
Transfer learning: fix model visualization to report the predicted label correctly

okurisama:
The current tutorial reports the validation set label as though it was the predicted label

### DIFF
--- a/beginner_source/transfer_learning_tutorial.py
+++ b/beginner_source/transfer_learning_tutorial.py
@@ -253,9 +253,7 @@ def visualize_model(model, num_images=6):
             images_so_far += 1
             ax = plt.subplot(num_images//2, 2, images_so_far)
             ax.axis('off')
-            ax.set_title('predicted: {}, actual: {}'.format(
-                dset_classes[preds[j]], dset_classes[labels.data[j]]
-            ))
+            ax.set_title('predicted: {}'.format(dset_classes[preds[j]]))
             imshow(inputs.cpu().data[j])
 
             if images_so_far == num_images:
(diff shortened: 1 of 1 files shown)


beginner_source/transfer_learning_tutorial.py
```diff
@@ -253,7 +253,9 @@ def visualize_model(model, num_images=6):
             images_so_far += 1
             ax = plt.subplot(num_images//2, 2, images_so_far)
             ax.axis('off')
-            ax.set_title('predicted: {}'.format(dset_classes[labels.data[j]]))
+            ax.set_title('predicted: {}, actual: {}'.format(
```
chsasank: Thanks for catching that! Showing both predicted and actual takes too much space. Can you please just show predicted?

beginner_source/transfer_learning_tutorial.py
```diff
@@ -253,7 +253,7 @@ def visualize_model(model, num_images=6):
             images_so_far += 1
             ax = plt.subplot(num_images//2, 2, images_so_far)
             ax.axis('off')
-            ax.set_title('predicted: {}'.format(dset_classes[labels.data[j]]))
+            ax.set_title('predicted: {}'.format(dset_classes[preds[j]]))
```
chsasank: It has to be preds[j, 0] because torch.max leaves a dummy axis.

okurisama: Really?  It works fine like this on my machine, using current Pytorch master head.

chsasank: Yep, in the latest release, this does raise an error.

chsasank:
#119 Fixes the same issue.

chsasank:
Thanks for the PR though :)
I somehow missed the last commit.

PR #107, opened 2017-07-05
fix a  small mistake

zakizhou:
change `landmarks_frame` in the definition of class `FaceLandmarksDataset` to `self.landmarks_frame`


chsasank:
Thanks @zakizhou !

PR #114, opened 2017-07-07
Fix dim of action in q_learning tutorial

quanvuong:
Expected output of select_action is [torch.LongTensor of size 1x1]


soumith:
thanks!

PR #126, opened 2017-08-19
Removed unused references to encoder_output in the attention decoder

v1n337:
Tested with modified code. The tutorial works fine.


hunkim:
Good catch!

davidbp:
You changed the .py file but  the error is still in the documentation. 

http://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html

PR #152, opened 2017-10-08
Improved passage on immediates and their undefined behaviour.

seba-1511:
Checked locally with `html-noplot`, but didn't update http://seba-1511.github.io/tutorials/, since the changes are trivial.


chsasank:
Thanks, merging.

PR #167, opened 2017-10-24
update requirements.txt

raul-jr3:
added pillow to the requirements.txt as the 'neural_style_tutorial.py' imports PIL


chsasank:
Thanks!

PR #169, opened 2017-10-25
Fixed simple logic error in tutorial

dat-ai:
I found a simple logic error.

Changed `model(x) ` to `model.forward(x)` in line 56.


soumith:
.forward is the wrong thing to do. the existing code is correct

dat-ai:
Thanks, that is correct. I did not notice the `torch.nn.Module`.

PR #177, opened 2017-11-15
Fix softmax warnings in tutorials

zou3519:
These happen because of the 0.3 changes.


soumith:
will merge when 0.3 is live.

chsasank:
Merged because 0.3 is live

chsasank:
Thanks!

PR #187, opened 2017-12-13
Fix loss calculation in transfer learning tutorial

### LINKED ISSUE #179: Average loss in transfer learning tutorial calculated incorrectly
Hi! Correct me if I'm wrong, but it seems like the average loss for the epoch is calculated incorrectly the transfer learning tutorial.

The criterions used are initialized with `size_average=True` (the default). In addition, the code computes `dataset_sizes` for `train` and `val` sets, which represent the # of samples in each set. After each batch, we accumulate the output of the criterion in a variable called `total_loss`, and then divide by the appropriate `dataset_sizes` value for the phase we're in. However, this `total_loss` is not actually the total loss for each example, but the total loss for each batch.

Since the criterion as currently instantiated is returning the *average loss over the examples in the batch*, we should divide that by `num_batches` in order to get a proper value for the average loss over the epoch. 

Alternatively, to account for the fact that batches can be different sizes, it'd probably be best to make the criterion return the *summed loss over the examples in the batch*, then dividing by the # of samples in the dataset like we're currently doing.

Let me know if I'm missing something here, or if this is indeed a bug --- happy to submit a PR to fix if necessary.

Thanks for the tutorial -- was very informative!

cc: @chsasank

nelson-liu:
This fixes #179 by simply initializing the criterions with `size_average=False`.

### DIFF
--- a/beginner_source/transfer_learning_tutorial.py
+++ b/beginner_source/transfer_learning_tutorial.py
@@ -188,7 +188,7 @@ def train_model(model, criterion, optimizer, scheduler, num_epochs=25):
                     optimizer.step()
 
                 # statistics
-                running_loss += loss.data[0]
+                running_loss += loss.data[0] * inputs.size(0)
                 running_corrects += torch.sum(preds == labels.data)
 
             epoch_loss = running_loss / dataset_sizes[phase]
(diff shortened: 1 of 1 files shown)


chsasank:
Hey does this not mean lr is virtually changed because `size_average=False`?

chsasank:
I guess rather multiply loss with batch size while counting total loss

chsasank:
Thanks!

PR #188, opened 2017-12-16
Fixed deep copying model.state_dict()

b21527616:
Otherwise best_model_wts changes always to last epoch's state


chsasank:
Thanks!

Issue #192, opened 2018-01-05
Error in seq2seq tutorial


cedricdeboom:
I believe there is an error in the seq2seq tutorial. In the forward pass of the encoder, the returned hidden state is used to initialize the hidden state of the next GRU layer:

https://github.com/pytorch/tutorials/blob/7b35173c3f65d000c238e8199ac638a5ce6d40dd/intermediate_source/seq2seq_translation_tutorial.py#L347-L352

This is incorrect. The returned hidden state is the final hidden state of the processed sequence (i.e. at t=sequence_length), and is generally used to transfer hidden states between sequences. It should, however, **not** be used to initialize the hidden state of the upper layer.

chsasank:
please checkout #195.

PR #195, opened 2018-01-17
removed references to layers as they were not working correctly

holli:
Model made it look like it was easy to add new layers by giving parameter n_layers=2. The implementation didn't create extra layers. Instead it put the information through same layer multiple times.

Correct way would be to call `gru = nn.GRU(self.hidden_size, self.hidden_size, num_layers=n_layers)` so that the model would have more trainable parameters. But with extra layers the usage of hidden state would be a bit more complicated in the decoder with attention so in this tutorial its probably better just to leave the extra layers out.


holli:
seems like there was one existing issue about the same thing, this fixes also that https://github.com/pytorch/tutorials/issues/192

PR #202, opened 2018-01-26
Some suggestions regarding typos and style in the Blitz document

rasbt:
I started reading through the docs and had some tiny suggestions if useful.


chsasank:
Thanks!

PR #204, opened 2018-01-31
Fix a link linking to DataParallel

rasbt:
Fixes a hyperlink linking to `DataParallel` in the API documentation.


chsasank:
Thanks.

PR #207, opened 2018-02-03
Fix 'grad_output' with 'gradient'

xu-song:
`grad_output` has change to `gradient` in [PyTorch API](https://github.com/pytorch/pytorch/blob/master/torch/autograd/variable.py#L99). 
However this tutorial is still using `grad_output`.


soumith:
thanks!

PR #211, opened 2018-02-13
Use foo.new instead of torch.FloatTensor for GPU.

lucasb-eyer:
This replaces the calls to `torch.FloatTensor` by a call to `.new` on the input tensor, such that GPU types are respected.


chsasank:
Thanks, sorry for very late response.

PR #212, opened 2018-03-03
Changed "creator" to "grad_fn" in Variable.png

sdmonov:
This picture remained unchanged after .creator was renamed to .grad_fn.


soumith:
thanks! :)

PR #214, opened 2018-03-07
Cpp extension

goldsborough:
Moves the tutorial for using C++ extensions here, from https://github.com/pytorch/pytorch/pull/5541.

Reviews still welcome. @apaszke @zdevito @soumith 

I've also removed the C extensions tutorial, in order to promote the C++ extensions, which are a superset of functionality and easier to use. Let me know if this is not OK.


.gitignore
```diff
@@ -105,3 +105,4 @@ ENV/
 
 # Rope project settings
 .ropeproject
+intermediate_source/
```
chsasank: Why is this added to gitignore?

goldsborough: I can't remember right now, but it might have been temporary output from the docs build that I wanted to not commit.

advanced_source/cpp_extension.rst
```diff
@@ -0,0 +1,1033 @@
...
+
+More performance increases!
+
+Conclusion
+----------
+
+You should now be equipped with a good overview of PyTorch's C++ extension
+mechanism as well as a motivation for using them. You can find the code examples
+displayed in this note `here
+<https://github.com/goldsbborough/pytorch-cpp-extension>`_. If you have
```
t-vi: Is this intentionally not the link to https://github.com/pytorch/extension-cpp/ ?

RoyiAvital:
Could we leave option for C only extensions?
Some would prefer sticking with C.

advanced_source/cpp_extension.rst
goldsborough: Nope, this needs to be fixed.

goldsborough:
@RoyiAvital Not really, for one we're trying to phase out TH* things wherever we can, and for the other getting rid of `cffi` would be a blessing and make everyone's life much nicer. No one forces you to use classes with C++ extensions :)

advanced_source/cpp_extension.rst
```diff
@@ -0,0 +1,1033 @@
...
+
+  class LLTM(torch.nn.Module):
+      def __init__(self, input_features, state_size):
+          super(LLTM, self).__init__()
+          self.input_features = input_features
+          self.state_size = state_size
+          # 3 * state_size for input gate, output gate and candidate cell gate.
+          # input_features + state_size because we will multiply with [input, h].
+          self.weights = torch.nn.Parameter(
+              torch.Tensor(3 * state_size, input_features + state_size))
```
SsnL: Can we use `torch.empty` instead of `torch.Tensor`?

advanced_source/cpp_extension.rst
```diff
@@ -0,0 +1,1033 @@
...
+  import torch
+
+  assert torch.cuda.is_available()
+
+  batch_size = 16
+  input_features = 32
+  state_size = 128
+
+  # Note the .cuda() calls here
+  X = torch.randn(batch_size, input_features).cuda()
```
SsnL: Use the new `device=` option please.

advanced_source/cpp_extension.rst
```diff
@@ -0,0 +1,1033 @@
...
+
+  forward = 0
+  backward = 0
+  for _ in range(100000):
+      start = time.time()
+      new_h, new_C = rnn(X, (h, C))
+      forward += time.time() - start
+
+      start = time.time()
+      (new_h.sum() + new_C.sum()).backward()
```
SsnL: Don't we need synchronization to get correct timings?

goldsborough: what?

PR #224, opened 2018-04-24
update tutorials for 0.4 (#223)

soumith:
* Update beginner tutorials

* Update rest tutorials

* update q learning

* format

* rewrote part of fast neural transfer to add normalization and let it make much more sense

* improve fast neural transfer code again

### DIFF
--- a/beginner_source/former_torchies_tutorial.rst
+++ b/beginner_source/former_torchies_tutorial.rst
@@ -25,9 +25,6 @@ In this tutorial, you will learn the following:
 .. galleryitem:: /beginner/former_torchies/tensor_tutorial.py
     :figure: /_static/img/tensor_illustration_flat.png
 
-.. galleryitem:: /beginner/former_torchies/autograd_tutorial.py
-    :figure: /_static/img/Variable.png
-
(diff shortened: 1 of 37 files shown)


beginner_source/former_torchies_tutorial.rst
```diff
@@ -25,9 +25,6 @@ In this tutorial, you will learn the following:
 .. galleryitem:: /beginner/former_torchies/tensor_tutorial.py
     :figure: /_static/img/tensor_illustration_flat.png
 
-.. galleryitem:: /beginner/former_torchies/autograd_tutorial.py
```
chsasank: @soumith Why is this removed? former_torchies/autograd_tutorial.py itself is edited as a part of the PR. Should I add this back?

soumith: yes please, add this back. Not sure what happened here. cc: @SsnL

soumith: added back now.

chsasank: thanks! Will build the latest docs.

PR #227, opened 2018-04-25
Fix epoch accuracy (avoiding integer division)

Madder:
Fixes the constant 0 accuracy I get running this tutorial since v0.4
Epoch 0/24
train Loss: 0.8934 Acc: 0.0000
val Loss: 0.7645 Acc: 0.0000

Epoch 1/24
train Loss: 0.6792 Acc: 0.0000
val Loss: 0.2288 Acc: 0.0000


chsasank:
Thanks!

PR #228, opened 2018-04-26
add comment for possible RuntimeError

k0ala:
Since this is basically the first line of the first tutorial, users should be
warned that there is a good chance it will raise an error (otherwise users 
may conclude that their installation is faulty).


k0ala:
see for example this issue: https://github.com/pytorch/pytorch/issues/6339

chsasank:
I think this is an issue that needs to be fixed with a fix in the master. cc: @soumith

soumith:
yes, we're working on fixing it in master.

PR #232, opened 2018-04-30
fix #229 makes it work on Windows

### LINKED ISSUE #229: RNN Tutorials don't work on Windows
The tutorials [Classifying Names with a Character-Level RNN](http://pytorch.org/tutorials/intermediate/char_rnn_classification_tutorial.html) and [Generating Names with a Character-Level RNN](http://pytorch.org/tutorials/intermediate/char_rnn_generation_tutorial.html) don't work on Windows.

They attempt to extract the category from the filename using

    category = filename.split('/')[-1].split('.')[0]

which works on Linux where these look like `'data/names/Arabic.txt'` but not on Windows where they look like `'data/names\\Arabic.txt'`.

I suggest changing the code to

    category = os.path.splitext(os.path.basename(filename))[0]

(and `import os` of course)

no git installed on this machine, so no PR for now

jxcodetw:


zou3519:
Should fix #7033

soumith:
thank you!

PR #236, opened 2018-05-06
Tutorial on imbalanced datasets

ptrblck:
This tutorial deals with the problem of an imbalanced dataset and how to train a classifier on it.

After training a CNN on the original CIFAR10 dataset, we resample it to create an artificially imbalanced dataset. Since the CNN performs quite poorly on this new dataset, we use the ``WeightedRandomSampler`` in the first step and a weighted criterion afterwards to tackle the problem.

I've created the tutorial in the intermediate section, but I'm not sure if it's the right place.

Feedback regarding the text and code is very welcome!


intermediate_source/imbalanced_data_tutorial.py
```diff
@@ -0,0 +1,486 @@
...
+    pin_memory=torch.cuda.is_available())
+
+class_names = ('plane', 'car', 'bird', 'cat', 'deer', 'dog', 'frog', 'horse',
+               'ship', 'truck')
+nb_classes = len(class_names)
+
+###############################################################################
+# Let's have a look at the class distribution in the datasets.
+
+# Get all training targets and count the number of class instances
```
chsasank: This could go in a function as you use it elsewhere.

intermediate_source/imbalanced_data_tutorial.py
```diff
@@ -0,0 +1,486 @@
...
+# =================================
+#
+# Now let's simulate an imbalanced dataset.
+# We will resample the training and test dataset, so that some
+# classes will be undersampled. Let's reduce the amount of instances for the
+# first 5 classes by 90% leaving only the first 10% in the dataset.
+# The last 5 classes will keep their samples.
+
+# Create class proportions
+imbal_class_prop = imbal_class_prop = np.hstack(([0.1] * 5, [1.0] * 5))
```
chsasank: I think rather than relying on the original dataset's code, it's better to create a new Dataset class which creates something like `self.idxs` and in ``__getitem`` can return ``self.orig_dataset[self.idxs[i]]``. You can use this dataset for both train and test datasets.

chsasank: Current code to do this is a bit opaque. It's always best to have no code duplication.

intermediate_source/imbalanced_data_tutorial.py
ptrblck: Thanks for the suggestion! I've added a function named ``get_labels_and_class_counts``.

intermediate_source/imbalanced_data_tutorial.py
ptrblck: Yeah, I've had the same feeling. I created a ``Dataset`` with a function to resample the data.
However, since I need the targets and class counts for the weights, I've added a helper class to get these members. It's better, but maybe still not really clean.

ptrblck:
@chsasank Thank you for the review! I've added your suggestions. Let me know, what you think about the changes.

intermediate_source/imbalanced_data_tutorial.py
```diff
@@ -0,0 +1,512 @@
...
+
+class_names = ('plane', 'car', 'bird', 'cat', 'deer', 'dog', 'frog', 'horse',
+               'ship', 'truck')
+nb_classes = len(class_names)
+
+###############################################################################
+# Let's have a look at the class distribution in the datasets.
+
+
+def get_labels_and_class_counts(labels_list):
```
chsasank: Can you somehow move plotting code to one function as well? `get_labels_and_class_counts` is also used as class method, so this function is also confusing.

chsasank: function could be named visualize_class_imbalance with dataset args for example.

ptrblck: I use the barplot for the original distribution and the imbalanced one.
So I created a functon named `plot_class_distributions`.

`get_labels_and_class_counts` is created as a function and as a class method. 
I'm not sure, if I can get rid of the class method, since I need the targets and counts from the imbalanced dataset. Since in the last review I created an own `Dataset` with indices for the imbalanced classes, I have to get these indices back. The CIFAR10 dataset stored in `ImbalancedCIFAR10` has still its original targets and therefore counts.

intermediate_source/imbalanced_data_tutorial.py
```diff
@@ -0,0 +1,486 @@
...
+
+class_names = ('plane', 'car', 'bird', 'cat', 'deer', 'dog', 'frog', 'horse',
+               'ship', 'truck')
+nb_classes = len(class_names)
+
+###############################################################################
+# Let's have a look at the class distribution in the datasets.
+
+# Get all training targets and count the number of class instances
+train_targets = np.array(train_dataset.train_labels)
```
Angelina1996: Getting the following error on this line:
AttributeError: 'CIFAR10' object has no attribute 'train_labels'
Could you please help with this

svekars:
Closing this as it's been quite some time since it was created and no longer relevant.

PR #244, opened 2018-05-17
Use variable instead of magic number in data_parallel_tutorial.py

fthoele:
Uses the variable that was defined already in the header, but not used in the source code.


soumith:
thanks!

PR #265, opened 2018-06-23
Fix broken pipe error in cifar10_tutorial.py

krikru:


peterjc123:
@soumith Should we merge this?

soumith:
@peterjc123 is the issue fixed on pytorch master? we are planning a 0.4.1 release for July 15th, and if it's fixed on master, the fix will be in there.

peterjc123:
@soumith This is not a specific issue for PyTorch. It is just a general problem for Python on Windows. In interactive mode, the multiprocessing can't be used. And good news for a new release.

brianjo:
As we've reorganized and updated a number these tutorials, we're going to close most of the older PRs. If you would like to recreate this PR, please rebase on the latest build. Thanks for our understanding.

PR #268, opened 2018-06-29
corrected the example for cross-correlation with scipy

adam-dziedzic:
Fully correct gradients for the scipy example (checked with torch.autograd.gradcheck).


soumith:
thanks @adam-dziedzic !

PR #271, opened 2018-07-10
Hybrid Frontend Tutorials

inkawhich:
@orionr @goodlux 

Initial commit for both hybrid frontend tutorials and caffe2 onnx primer tutorial. Both added in beginner section.


inkawhich:
It may be a bit tricky to get this to build depending on what version of pytorch is being used. Some of the hybrid frontend code is not available in 0.4 so we were developing in the developer version 0.5. Also, I know we talked about not including the caffe2 onnx primer because of the requirements but in the advanced tutorial 'transferring model from pytorch to C2 with ONNX', there is a required ONNX import.

inkawhich:
@matthewinkawhich

orionr:
@jamesr66a and @zdevito as an FYI

orionr:
@inkawhich would you be able to split this into two pull requests - one for the hybrid tutorial and another for the Caffe2 and ONNX tutorial? I'm not sure if Caffe2 and ONNX is going to land here (still talking with @goodlux and @soumith), but definitely want the hybrid frontend tutorial. Thanks.

beginner_source/hybrid_frontend/introduction_to_hybrid_frontend_tutorial.py
```diff
@@ -0,0 +1,526 @@
...
+# PyTorch's eager mode is the familiar PyTorch interface. Eager mode is
+# easy because it is simply idiomatic Python. Because of the imperative
+# nature of the mode, we can use our favorite Python debugging tools such
+# as ``pdb`` and ``print`` statements. This mode is designed for rapid
+# model prototyping and ambitious research efforts.
+#
+# However, it is difficult to optimize models represented in eager mode
+# for large-scale production and deployment.
+#
+# Graph Mode
```
zdevito: "Script" mode is more consistent with the names in our API, where anything in this mode turns into a torch.jit.ScriptModule. So I think we should use "Script" or "Torch Script" consistently for this mode and not Graph. 

The terminology we have been using internally is that both `torch.jit.trace` and `@torch.jit.script` produce Torch Script functions.

beginner_source/hybrid_frontend/introduction_to_hybrid_frontend_tutorial.py
```diff
@@ -0,0 +1,526 @@
...
+# ----------
+#
+# The graph mode creates a deferred model representation, which is a
+# common methodology in many deep learning frameworks. In graph mode, you
+# write a model definition in Python, which is run later on in the Caffe2
```
zdevito: "Caffe2" -- this is not accurate. It should just be C++-only runtime environment.

beginner_source/hybrid_frontend/introduction_to_hybrid_frontend_tutorial.py
```diff
@@ -0,0 +1,526 @@
...
+
+
+
+######################################################################
+# JIT Compiler
+# ------------
+#
+# To accomodate to as many workflows as possible, PyTorch not only
+# provides both an eager mode and a graph mode, but a transition mechanism
+# between them: ``torch.jit``. This enables the use of the newest
```
zdevito: All of the script mode is contained under ``torch.jit`` so calling it the transition mechanism is not very accurate. We can say,  

""PyTorch provides mechanisms for incrementally converting eager-mode code into Torch Script."

beginner_source/hybrid_frontend/introduction_to_hybrid_frontend_tutorial.py
```diff
@@ -0,0 +1,526 @@
...
+#
+# Tracing mode is a non-invasive tool which traces your Python module or
+# function to create an internal representation that is compiled to graph
+# mode. The ``torch.jit.trace`` function takes your module or function and
+# an example data input, and traces the computational steps that the data
+# encounters as it progresses through the model.
+#
+# An advantage of this approach is that it is an unobtrusive process that
+# will work regardless of how your Python code is structured. However,
+# because of the way the tracer works, data-dependent control flows (if
```
zdevito: "control flow" (singular)

beginner_source/hybrid_frontend/introduction_to_hybrid_frontend_tutorial.py
```diff
@@ -0,0 +1,526 @@
...
+# ~~~~~~~~~~~
+#
+# To compile models with data-dependent control flow elements, such as
+# RNNs, we simply add a ``@torch.jit.script`` decorator to our function.
+# This annotation lets PyTorch know that the function may contain
+# data-dependent control flows. Now, our loops, conditionals, and
+# iterative data mutations are handled explicitly.
+#
+# The one constraint for users compiling with script mode is that it
+# currently only supports a restricted subset of Python. As of now,
```
zdevito: "currently" - this makes it seem like it will eventually be all of python, which is not true.

beginner_source/hybrid_frontend/introduction_to_hybrid_frontend_tutorial.py
```diff
@@ -0,0 +1,526 @@
...
+# features such as generators, defs, and Python data structures are not
+# supported. To remedy this, you can invoke traced modules from script
+# modules (and vice-versa), and you can call pure Python functions and
+# modules from script modules. However, the operations done in the pure
+# Python functions will not be compiled, and will run as-is.
```
zdevito: "will not be exportable" <- this is probably the biggest distinction here since if you want to run the model in production you will need to remove all the python calls.

beginner_source/hybrid_frontend/introduction_to_hybrid_frontend_tutorial.py
```diff
@@ -0,0 +1,526 @@
...
+########## Example ##########
+# Non-compiled
+print("Euclidean distance between train_set[0] and train_set[1]:")
+print("train_set[0]:", train_set[0])
+print("train_set[1]:", train_set[1])
+distance = getDistance(train_set[0], train_set[1])
+print("Non-compiled output:", distance)
+
+# Compiled
+getDistance_traced = trace(train_set[0], train_set[1])(getDistance)
```
zdevito: we updated the API to look nicer here:

`trace(getDistance, (train_set[0], train_set[1]))`

beginner_source/hybrid_frontend/introduction_to_hybrid_frontend_tutorial.py
```diff
@@ -0,0 +1,526 @@
...
+# does this by iterating over every instance in the ``train_set``,
+# recording the distance between the train instance and the
+# ``test_instance``, and using our ``getSortedLabels`` function to sort
+# the corresbonding labels based on these distances.
+#
+# --------------
+#
+# **Script mode**
+#
+# Notice that this function contains a data-dependent control flow (our
```
zdevito: "contains data-dependent control flow"

beginner_source/hybrid_frontend/introduction_to_hybrid_frontend_tutorial.py
```diff
@@ -0,0 +1,526 @@
...
+# functions such as ``addEntries``.
+#
+# Also, notice in the graph output that the loop is captured
+# (``prim::Loop``). The **script** mode enables us to convert the Python
+# AST directly to graph mode, rather than tracing an example input through
+# the control flow and simply capturing the unrolled loop.
+#
+
+# Add elements to distances and labels tensors at index i
+def addEntries(i, d, l, distances, labels):
```
zdevito: I am not sure this is the best example, unless you follow it up with an example that shows how to eliminate the Python. As it stands, it is not possible to call `.save` on this model because it has a call.

PR #276, opened 2018-07-24
Question: requires_grad defaults False?

imunolion:
The example seems want to say if not specifies, it points to False.


soumith:
thanks!

imunolion:
Good to know that I was right. Thank you for confirmation and the nice tutorial!

PR #292, opened 2018-08-29
Adds note on adding extra compile args for building cuda extensions

syed-ahmed:
When building CUDA extensions, we have to pass `extra_compile_args` now to avoid cuda header collisions with half operator overloading (that happens through implicit half conversions) details: https://github.com/pytorch/pytorch/pull/10301#issuecomment-416773333.


syed-ahmed:
Thanks @goldsborough !  I totally agree with you! I'll close this PR since you got this. Till then, people can just refer to https://github.com/pytorch/pytorch/pull/10301#issuecomment-416773333 if they get the operator overloading errors.

PR #300, opened 2018-09-13
DO NOT MERGE: Posting 'Deploying Seq2Seq Model w/ Hybrid Frontend' Jupyter Notebook for review comments

MatthewInkawhich:
@jamesr66a


Deploying_Seq2Seq_Model_with_Hybrid_Frontend.ipynb
```diff
@@ -0,0 +1,720 @@
...
+    "        output = F.softmax(output, dim=1)\n",
+    "        # Return output and final hidden state\n",
+    "        return output, hidden"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Data Handling\n",
```
jamesr66a: Feels weird having this here and not earlier on. I don't think it depends on anything that's been covered so far, and understanding what all this "embedding" stuff is doing requires understanding how we're representing our sequences.

Deploying_Seq2Seq_Model_with_Hybrid_Frontend.ipynb
```diff
@@ -0,0 +1,720 @@
...
+    "\n",
+    "Although our models conceptually deal with sequences of words, in reality, they deal with numbers like all machine learning models do. In this case, every word in the model's vocabulary, which was established before training, is mapped to an integer index. We use a `Voc` object to contain the mappings from word to index, as well as the total number of words in the vocabulary. We will load the object later before we run the model.\n",
```
jamesr66a: In general "words" -> "tokens", since we can use any linguistic unit, not just words

Deploying_Seq2Seq_Model_with_Hybrid_Frontend.ipynb
```diff
@@ -0,0 +1,720 @@
...
+    "\n",
+    "### Hybrid Frontend Notes:\n",
+    "\n",
+    "The `forward` method of this module involves iterating over the range of $[0, max\\_length)$ when decoding an output sequence one word at a time. Because of this, we should use **scripting** to convert this module to Torch Script. Unlike with our encoder and decoder models, which we can trace, we must make some necessary changes to the `GreedySearchDecoder` module in order to initialize an object without error. In other words, we must ensure that our module adheres to the rules of the scripting mechanism, and does not utilize any language features outside of the subset of Python that Torch Script includes.\n",
+    "\n",
+    "To get an idea of some manipulations that may be required, we will go over the diffs between the `GreedySearchDecoder` implementation from the chatbot tutorial and the implementation that we use in the cell below. Note that the lines highlighted in red are lines removed from the original implementation and the lines highlighted in green are new.\n",
+    "\n",
+    "![diff](images/gsd_diff.png)\n",
```
jamesr66a: I don't like this diff, since using `device` and `dtype` as attributes on the tensor constructors (e.g. `torch.ones`) likely should have been done in the original code to prevent cross-device copies and unnecessary casts

Deploying_Seq2Seq_Model_with_Hybrid_Frontend.ipynb
```diff
@@ -0,0 +1,720 @@
...
+    "\n",
+    "\n",
+    "- Added `decoder_n_layers` to the constructor arguments\n",
+    "\n",
+    "    - This change stems from the fact that the encoder and decoder models that we pass to this module will be a child of `TracedModule` (not `Module`). Therefore, we cannot access the decoder's number of layers with `decoder.n_layers`. Instead, we plan for this, and pass this value in during module construction.\n",
+    "\n",
+    "\n",
+    "- Store away new attributes as constants\n",
+    "\n",
+    "    - In the original implementation, we were free to use variables from the surrounding (global) scope in our `GreedySearchDecoder`'s `forward` method. However, now that we are using scripting, we do not have this freedom. An easy solution to this is to store these values from the global scope as attributes to the module in the constructor, and add them to a special list called `__constants__` so that they can be used as literal values when constructing the graph in the `forward` method. An example of this usage is on NEW line 19, where instead of using the `device` and `SOS_token` global values, we use our constant attributes `self._device` and `self._SOS_token`.\n",
```
jamesr66a: Mention that we cannot do this because the assumption with scripting is that we cannot necessarily hold on to the Python objects, especially if it goes into export (which hasn't been introduced, but oh well)

Deploying_Seq2Seq_Model_with_Hybrid_Frontend.ipynb
```diff
@@ -0,0 +1,720 @@
...
+    "\n",
+    "\n",
+    "### Decoder\n",
+    "\n",
+    "We perform the same process for tracing the decoder as we did for the encoder. Notice that we call forward on a set of random inputs to the traced_encoder to get the output that we need for the decoder. This is not required, as we could also simply create a tensor of the correct shape, type, and value range. We trace the model in the exact fashion that we traced the encoder.\n",
```
jamesr66a: - "This is not required, as we could also simply create a tensor of the correct shape, type, and value range" probably need to elaborate on why this is the case (we do not have any constraints on the values of the tensor because we don't have any ops that could fault on out-of-range inputs)

Deploying_Seq2Seq_Model_with_Hybrid_Frontend.ipynb
```diff
@@ -0,0 +1,720 @@
...
+    "print('decoder graph', traced_decoder.graph)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print('searcher graph', scripted_searcher.graph)"
```
jamesr66a: You could get away with only showing the graph for the `scripted_searcher` instance, because the graphs for the encoder and decoder will be directly inlined.

Deploying_Seq2Seq_Model_with_Hybrid_Frontend.ipynb
```diff
@@ -0,0 +1,750 @@
...
+    "Now that we have successfully converted our model to Torch Script, we will serialize it for use in a non-Python deployment environment. To do this, we can simply save our `scripted_searcher` module, as this is the user-facing interface for running inference against the chatbot model."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 14,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "torch.save(scripted_searcher.state_dict(), \"scripted_chatbot.pth\")"
```
zdevito: This is wrong save. This should be using `scripted_searcher.save("scripted_chatbot.pth")`.

MatthewInkawhich: Thanks, I'll make this change.

PR #302, opened 2018-09-14
Tutorial on loading a model in C++

goldsborough:


advanced_source/cpp_export.rst
```diff
@@ -0,0 +1,366 @@
...
+model that can be understood, compiled and serialized by the Torch Script
+compiler. If you are starting out from an existing PyTorch model written in the
+vanilla "eager" API, you must first convert your model to Torch Script. In the
+most common cases, exemplified below, this requires only little effort. If you
+already have a *scripted* module (a module written in Torch Script), you can
+skip to the next section of this tutorial.
+
+There are two primary ways of converting a PyTorch model to Torch Script. The
+first is known as "tracing", a mechanism in which the structure of the model is
+captured by evaluating it once using examplary inputs, and recording the flow
```
jamesr66a: "examplary" -> "exemplary", although i think that has a different connotation than "example" (i.e. "exemplary" implies there exist desiderata to be copied, but "example" is just an instance of something). "Example inputs" sounds right to me

advanced_source/cpp_export.rst
```diff
@@ -0,0 +1,366 @@
...
+    include/
+    lib/
+    share/
+
+- The ``lib/`` folder contains the shared libraries you must link against,
+- The ``include/`` folder contains header files your program will need to include,
+- The ``share/`` folder contains the necessary CMake configuration to enable the simple `find_package(Torch)` command above.
+
+The last step is building the application. For this, assume our example
+directory is layed out like this:
```
jamesr66a: "layed out" -> "laid out"

advanced_source/cpp_export.rst
```diff
@@ -0,0 +1,366 @@
...
+----------------------------------------
+
+Having succesfully loaded our serialized ``ResNet18`` in C++, we are now just a
+couple lines of code away from executing it! Let's add those lines to our C++
+application's ``main()`` function:
+
+.. code-block:: cpp
+
+    // Create a vector of inputs.
+    std::vector<torch::jit::IValue> inputs;
```
apaszke: I think IValue is being moved to ATen core, so this might need an update in the future

apaszke:
Also, didn't you forget to add this tutorial to the TOC? I can't see it in the preview.

advanced_source/cpp_export.rst
```diff
@@ -0,0 +1,364 @@
...
+
+  find_package(Torch REQUIRED)
+
+  add_executable(example-app example-app.cpp)
+  target_include_directories(example-app PRIVATE "${TORCH_INCLUDE_DIRS}")
+  target_link_libraries(example-app "${TORCH_LIBRARIES}")
+  target_compile_definitions(example-app PRIVATE -D_GLIBCXX_USE_CXX11_ABI=0)
+  set_property(TARGET example-app PROPERTY CXX_STANDARD 11)
+
+The last thing we need to build the example application is the LibTorch
```
zdevito: Add a warning that the C++ version and the conda version you exported with need to be matched. In the future we will provide a better BC story. Not in preview though.

advanced_source/cpp_export.rst
zdevito: We should keep a consistent alias to it at either `torch::IValue` or `torch::jit::IValue` just like we do for `torch::Tensor`

advanced_source/cpp_export.rst
```diff
@@ -0,0 +1,364 @@
...
+    auto output = module->forward(inputs).toTensor();
+
+    std::cout << output.slice(/*dim=*/1, /*start=*/0, /*end=*/5) << '\n';
+
+The first two lines set up the inputs to our model. We create a vector of
+``torch::jit::IValue`` (a type-erased value type ``script::Module`` methods
+accept and return) and add a single input. To create the input tensor, we use
+``torch::ones()``, the equivalent to ``torch.ones`` in the C++ API.  We then
+run the ``script::Module``'s ``forward`` method, passing it the input vector we
+created. In return we get a new ``IValue``, which we convert to a tensor.
```
zdevito: "We need to call `toValue()` on the result because not all models will return a single tensor. Some might return a Tuple or a lists of tensors. IValue has methods to extract each from the result."

advanced_source/cpp_export.rst
```diff
@@ -0,0 +1,364 @@
...
+  root@4b5a67132e81:/example-app/build# ./example-app model.pt
+  -0.2698 -0.0381  0.4023 -0.3010 -0.0448
+  [ Variable[CPUFloatType]{1,5} ]
+
+
+For reference, the output in Python previously was::
+
+  tensor([-0.2698, -0.0381,  0.4023, -0.3010, -0.0448], grad_fn=<SliceBackward>)
+
+Looks like a good match!
```
zdevito: Plug the C++ API here: if you need to do any preprocessing, basically all of PyTorch is available as a C++ API on these torch::Tensor objects.

PR #306, opened 2018-09-16
Updated Chatbot and Added Deploying Seq2Seq w/ Hybrid Frontend tutorial

MatthewInkawhich:
Changes:
1) Added new Deploying Seq2Seq model w/ hybrid frontend tutorial
2) Minor updates to chatbot tutorial explanations
3) Changes to _static/img/chatbot/ images
4) Removed Introduction to Hybrid Frontend tutorial for now... I have incorporated the important info here into the deploying seq2seq w/ h.f. tutorial, so this one is no longer necessary, as Nate has an introductory tutorial already


lezhang-thu:
> > Hey, I saw the code on Chatbot published at [https://pytorch.org/tutorials/beginner/chatbot_tutorial.html](url)
> > In the function `maskNLLLoss`, when the following is executed:
> > `loss = crossEntropy.masked_select(mask).mean()`
> > I noticed that `crossEntropy` is of size `(batch_size, 1)`, but `mask` is of size `(batch_size)`. It will cause a problem, won't it? These two tensors are of different sizes. @MatthewInkawhich
> 
> Hey @lezhang-thu, this should not cause a problem because the shapes of the input tensor and the mask tensor for the `masked_select` operation do not need to match, but they need to be [broadcastable](https://pytorch.org/docs/stable/notes/broadcasting.html#broadcasting-semantics).

Hey @MatthewInkawhich , thanks for the reply. I feel it might cause a problem, as shown in the following example.
```
>>> x = torch.randn(4,1)
>>> x
tensor([[-0.3857],
        [ 0.5516],
        [ 0.9389],
        [-0.6940]])
>>> mask = torch.tensor([0,1,1,0], dtype=torch.uint8)
>>> x.masked_select(mask)
tensor([-0.3857, -0.3857,  0.5516,  0.5516,  0.9389,  0.9389, -0.6940, -0.6940])
>>> x.masked_select(mask.view(-1,1))
tensor([0.5516, 0.9389])
```

MatthewInkawhich:
@lezhang-thu Hmm. I see what you mean. I borrowed this function from https://github.com/ywk991112/pytorch-chatbot. I made your suggested changes on my local machine, and the chatbot output quality is not good. I will investigate this further, but I will leave the implementation as-is until we figure out what implementation is best.

Good catch though, I appreciate it!

PR #336, opened 2018-10-09
Fix line endings in csvwriter for windows machine

abhishekraok:
On windows machines the default line endings are \r\n instead of \n. This causes code to fail in the filterPair function. Fix details here https://stackoverflow.com/questions/8746908/why-does-csv-file-contain-a-blank-line-in-between-each-data-line-when-outputting


soumith:
thanks @abhishekraok

PR #338, opened 2018-10-11
Fix maskNLLLoss in chatbot_tutorial.py

manipopopo:
The original `crossEntropy` has shape `[batch_size, 1]` and `mask` has shape `[batch_size]`. Calling `crossEntropy.masked_select(mask)` will get `crossEntropy.expand(-1, nTotal).flatten()`.

This patch flatten `crossEntropy` so that `crossEntropy.masked_select(mask)` will only contain the elements corresponding to 1 in the `mask`.


manipopopo:
Addressed in #382.

PR #346, opened 2018-10-24
Adds autograd tutorial to 60 minute blitz. 

brianjo:
Also includes a draft article still under review. I'll add that to the TOC when we're ready to call it complete.


brianjo:
Joel, yes on the article. I'll update the TOC as soon as I have more feedback. Thanks!

PR #348, opened 2018-10-29
Removed onnx_caffe2 and related docs, update to caffe2.python.onnx

zrphercule:
onnx_caffe2 and related things are not be used anymore, now we are using caffe2.python.onnx.

### DIFF
--- a/advanced_source/super_resolution_with_caffe2.py
+++ b/advanced_source/super_resolution_with_caffe2.py
@@ -8,10 +8,10 @@ Caffe2, we can run the model to double-check it was exported correctly,
 and we then show how to use Caffe2 features such as mobile exporter for
 executing the model on mobile devices.
 
-For this tutorial, you will need to install `onnx <https://github.com/onnx/onnx>`__,
-`onnx-caffe2 <https://github.com/onnx/onnx-caffe2>`__ and `Caffe2 <https://caffe2.ai/>`__.
-You can get binary builds of onnx and onnx-caffe2 with
-``conda install -c ezyang onnx onnx-caffe2``.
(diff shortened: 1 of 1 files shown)


advanced_source/super_resolution_with_caffe2.py
```diff
@@ -8,10 +8,10 @@
...
 executing the model on mobile devices.
 
-For this tutorial, you will need to install `onnx <https://github.com/onnx/onnx>`__,
-`onnx-caffe2 <https://github.com/onnx/onnx-caffe2>`__ and `Caffe2 <https://caffe2.ai/>`__.
-You can get binary builds of onnx and onnx-caffe2 with
-``conda install -c ezyang onnx onnx-caffe2``.
+For this tutorial, you will need to install `onnx <https://github.com/onnx/onnx>`__
+and `Caffe2 <https://caffe2.ai/>`__.
+You can get binary builds of onnx with
+``conda install -c ezyang onnx``.
```
dzhulgakov: are we still guiding people toward conda?

also - since C2 and PyTorch are now part of the same package - do we actually need to install C2 separately?

zrphercule: @dzhulgakov We can install C2 separately without installing pytorch, I guess that would be faster. Although guiding people to caffe2.ai is not a good idea I believe, I would guide them to github of pytorch.

As for conda, I dont think it is a good idea to use it. Should we change to use pip here?

houseroad: I would suggest `pip install onnx` :-)

conda package is out of date, too

PR #358, opened 2018-11-19
Fix wrong parameter.

tczhangzhi:
The second parameter was wrong which means the name (for error messages) of the kernal. It is repeated with the name of the forward kernel and should be backward kernal.


soumith:
thank you @tczhangzhi !

PR #365, opened 2018-11-30
Remove unnecessary js files in _static

brsoff:
These files are already included/generated in the sphinx theme.

Files in '_static' in this repo will overwrite the corresponding file in the sphinx theme, keeping the build from using the up-to-date files.

Compare the corresponding _static folder in the docs: https://github.com/pytorch/pytorch/tree/master/docs/source/_static

(If the build is still going to take a long time after merging this you can quickly fix the javascript on the live tutorials site in the meantime by replacing the current `theme.js` in the `gh-pages` branch with the latest `theme.js` build file (https://github.com/pytorch/pytorch.github.io/blob/site/docs/master/_static/js/theme.js). 

Example here: https://github.com/pytorch/tutorials/commit/dc10b659161cc0ca96636535810b75ba742bc2de, preview at https://brsoff.github.io/tutorials/


JoelMarcey:
@brsoff Hi. Thanks for this. I have a couple of questions:

> If the build is still going to take a long time after merging this you can quickly fix the javascript on the live tutorials site in the meantime by replacing the current theme.js in the gh-pages branch with the latest theme.js build file 

What about the `vendor` .js files that you deleted here, but would still remain until the build finishes? Do those files need to be manually replaced too?

> These files are already included/generated in the sphinx theme.

I assume this file is ok? https://github.com/pytorch/pytorch.github.io/blob/site/docs/master/_static/js/modernizr.min.js

brsoff:
@JoelMarcey Apologies, I should have mentioned that the current vendor files on `gh-pages` are still the same as those in the build so don't need to be replaced. And yes the `modernizr.min.js` file can stay, that's not included in the sphinx theme.

PR #368, opened 2018-12-05
[UPDATED] AWS distributed training tutorial

teng-li:
cc: @inkawhich @JoelMarcey @jspisak


jspisak:
Thanks @teng-li!!

JoelMarcey:
@orionr Depends on your definition of expected. 😄  In theory, no. But we know the builds are a bit of a mess now, so I would say yes, short-term.

cc: @brianjo

PR #378, opened 2018-12-08
Fix aws_distributed_training_tutorial

yf225:


yf225:
@teng-li I found a bug in the generated HTML :(  I am going to open another PR for this

PR #382, opened 2018-12-13
fix CrossEntropy size in chatbot to align with the mask

aobo-y:
I feel how the following filter loss through mask is queationable

assume `batch_size` is `64`, then
`mask` size is `(64)`
`crossEntropy` size is `(64, 1)`

I know you can take use of **[Broadcasting](https://pytorch.org/docs/stable/notes/broadcasting.html#broadcasting-semantics)**. That's why it can run without exceptions.
However, i guess Broadcasting changes `mask` to `(1, 64)` to work.

Below is sample test how it's wrong
screenshot 2018-12-13 at 12 38 31 am

Here I simply squeeze the added dim for `gather`, so their sizes are both `(64)`


soumith:
nice catch! thank you

PR #397, opened 2018-12-19
Change format of note about types in custom TorchScript ops

goldsborough:
@soumith


goldsborough:
@soumith can we merge? The build failures look unrelated

PR #399, opened 2018-12-21
Add Pytorch Sphinx theme as proper submodule

JoelMarcey:
Let's see if this allows Netlify to build without breaking our build


JoelMarcey:
The question we will need to answer, and maybe the PR checks will tell us, is if we need to update the any of the Circle/Jenkins/etc build scripts because of this. 

Maybe everything just passes 😄

JoelMarcey:
Things are looking good thus far between Netlify and Circle. I will wait until Circle is done and then merge. I will probably remove some of the Netlify notifications, the mixed content particularly.

JoelMarcey:
I think we are good here. I am going to merge.

PR #406, opened 2019-01-08
Missing space character for an example

eonr:
All the other apostrophe-containing prefixes have a space after the last character except for "she s" (should be "she s ").


soumith:
thanks!

PR #415, opened 2019-01-15
change the get_screen method so that it does not fill the gpu memory

andreistirb:
I changed the get_screen method so that it does not copy the tensor to GPU device at the return point of the method. If we copy the tensor each and every time to the device at some point we would fill the memory with unnecessary data. 
Also, in order for this to work I changed the forward method from the DQN object so that it copies the input to device before doing any other stuff.


soumith:
cc: @mike7ant

holly1238:
Hi @andreistirb! Can you sign the CLA agreement so that I can merge the pull request? Thanks!

andreistirb:
Hello @holly1238, I have completed the form, thank you for accepting my contribution!

PR #418, opened 2019-01-15
add windows tip to cpp tutorials

suo:
Users following the tutorial on windows had issues when building in debug mode while trying to link to the release DLL. I've added a tip that recommends building from source if they want debug mode.


suo:
ping @soumith

PR #420, opened 2019-01-17
Fix broken https URL back to http

seungwonpark:
- http link works: http://www.cs.columbia.edu/%7Emcollins/crf.pdf 
- but https doesn't: https://www.cs.columbia.edu/%7Emcollins/crf.pdf


soumith:
thanks!

PR #430, opened 2019-01-31
Update cpp_frontend.rst

fakerbrother:
Can't use linear(input) if u don't reload operator() method, so use linear->forward(input).


soumith:
cc: @yf225 to merge or close

PR #434, opened 2019-02-12
Commented suggested example variables

polplop:
Shifted the "where for example X is ..." to comments. 

Found it confusing due to sphinx applying bold i.e. ""where **for** example X **is** ..."


soumith:
thank you!

PR #438, opened 2019-03-01
Fix circle build

JoelMarcey:
I was told this line added to our config should fix the problem


JoelMarcey:
Hasn't fixed it yet. Looking.

.circleci/config.yml
```diff
@@ -29,7 +29,8 @@ pytorch_tutorial_build_defaults: &pytorch_tutorial_build_defaults
           linux-image-generic \
           moreutils \
           nvidia-docker2 \
```
yf225: We likely also need to pin the version of `nvidia-docker2`. For example:
```
nvidia-docker2=2.0.3+docker18.06.2-1
```

PR #440, opened 2019-03-04
typo fix & fix hyper link

youkaichao:
the hyperlink in [restructured text](http://docutils.sourceforge.net/docs/user/rst/quickref.html#hyperlink-targets)  is not the same as markdown. Currently, the rendering is wrong. ( check [here](https://pytorch.org/tutorials/advanced/numpy_extensions_tutorial.html))


soumith:
thanks!

PR #445, opened 2019-03-07
Update parallelism_tutorial.py

keunwoochoi:
Regarding https://github.com/pytorch/pytorch/issues/16885 .

### DIFF
--- a/beginner_source/former_torchies/parallelism_tutorial.py
+++ b/beginner_source/former_torchies/parallelism_tutorial.py
@@ -11,6 +11,7 @@ Data Parallelism is implemented using ``torch.nn.DataParallel``.
 One can wrap a Module in ``DataParallel`` and it will be parallelized
 over multiple GPUs in the batch dimension.
 
+
 DataParallel
 -------------
 """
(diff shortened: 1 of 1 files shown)


soumith:
thanks @keunwoochoi !

keunwoochoi:
😊👍

wangg12:
@keunwoochoi I got an error by adding this method:
```
    return getattr(self.module, name)
  [Previous line repeated 329 more times]
RecursionError: maximum recursion depth exceeded while calling a Python object
```
What is the correct way of accessing the updated module attributes?

keunwoochoi:
@wangg12 Alright, not sure if it's related but could you try this?
```python
class OpenDataParallel(nn.DataParallel):
    """custom class so that I can have other attributes than modules."""

    def __init__(self, module, **kwargs):
        super(OpenDataParallel, self).__init__(module, **kwargs)

    def __getattr__(self, name):
        if name is not 'module':
            try:
                return getattr(self.module, name)
            except AttributeError:
                pass

        return super(nn.DataParallel, self).__getattr__(name)


```

PR #450, opened 2019-03-22
Making loops a bit more pythonic

MrinalJain17:
**Replaced**

```
for i in range(len(data)):
    sample = data[i]
    ...
```

**with**

```
for i, sample in enumerate(data):
    ...
```


MrinalJain17:
Hmm... Looks like the `Dataset` does not support *`enumerate`ing*.

https://github.com/pytorch/pytorch/blob/77a7285764334cc50eb573c3d2614ffefba4a29f/torch/utils/data/dataset.py#L8-L23

PR #452, opened 2019-03-25
reinforcement_q_learning: Remove hard-coded entries for action space.

edowson:
This commit removes hard-coded entries for the output action space
and gets the value directly from the gym environment.

Signed-off-by: Elvis Dowson <elvis.dowson@gmail.com>


soumith:
thank you @edowson !

PR #471, opened 2019-04-08
fix broken link

tomguluson92:
@inkawhich, it strike to me that the original link of dataset celeA is broken. So this PR is to fix it.


tomguluson92:
Please review, thank you~

soumith:
thank you!

PR #472, opened 2019-04-08
 Include GoogLeNet aka Inception v1 

rajasekharponakala:


soumith:
it complicates the tutorial (tutorials need to be explanatory and simple) and isn't really necessary to get the point across. So I'd prefer we dont merge this.

PR #475, opened 2019-04-09
Replace redundant module names

ludweeg:
Provided that: 
```python 
import torch.nn as nn
```
`torch.nn` module name can be replaced with `nn`


soumith:
thank you!

PR #479, opened 2019-04-14
Update cifar10_tutorial.py

fabiofumarola:
- use the classes in the trainset and not declaring them statically
- iterating over the labels


soumith:
i disagree that this is clearer. showing that classes are a list of strings is worth it for the tutorial imo

PR #480, opened 2019-04-14
Update data_loading_tutorial.py

fabiofumarola:
fix error on the call of the function to apply the transformations. Add a remark in case the first image in the dataset returns an error.


soumith:
@fabiofumarola i cant reproduce the issue of the first image returning error, we have a continuous build where all tutorials are run on a clean machine. i think the error might be when your machine downloaded the dataset, maybe?

holly1238:
Closing this PR because the issue cannot be reproduced.

PR #483, opened 2019-04-17
Upgrade CPP extension tutorial

ClementPinard:
Along with more up to date function calls,
a tutorial for packed tensor accessors is proposed.

See the corresponding PR in cpp-extension [here](https://github.com/pytorch/extension-cpp/pull/33)


soumith:
thanks @ClementPinard , looks great

PR #486, opened 2019-04-22
Fix link to Sphinx-Gallery

Trabing:
The link does not work because Sphinx-Gallery's documentation is no longer hosted on Read the Docs.


soumith:
thank you!

wAuner:
the seems to be broken again.

PR #487, opened 2019-04-23
moving input data to device

ShishirPatil:
1. Fix: Moving input "data" to device. 
2. This in-turn benefits from describing what "data" list is composed of.


soumith:
thanks!

PR #496, opened 2019-05-06
 Fix a broken link for STN image

GHamrouni:
This merge fixes a bug causing STN tutorial to disappear from gallery items


soumith:
thank you!

PR #499, opened 2019-05-09
[nn_tutorial.py] remove line `var` before line `print(var)`

Oktosha:
It's really a typo-level fix. I suggest to remove the line 

```
x_train, x_train.shape, y_train.min(), y_train.max()
```

which comes right before the lines

```
print(x_train, y_train)
print(x_train.shape)
print(y_train.min(), y_train.max())
```

It looks like the line I suggest to remove was used to make the notebook to display the variables. But there are print statements below that print the very same variables. So, if I get it right, the line I suggest to remove serves for nothing. That's why I suggest to remove it for clarity.


holly1238:
Hi @Oktosha! Can you sign the CLA agreement so that I can merge the pull request? Thanks!

PR #509, opened 2019-05-25
Fixed a bug in data_loading_tutorial

lewha0:
Border lines of the grids are not considered; The coordinates of the landmarks should be shifted accordingly.


soumith:
thanks!

Issue #432, opened 2019-02-05
Caffe2 Mobile Exporter link broken


alejo9719:
The link for the Caffe2 mobile exporter in the section "Running the model on mobile devices" of https://pytorch.org/tutorials/advanced/super_resolution_with_caffe2.html is outdated. I think the correct link is https://github.com/pytorch/pytorch/blob/master/caffe2/python/predictor/mobile_exporter.py.

jspisak:
This was fixed in the above commits. Closing the issue..

PR #512, opened 2019-05-27
Fix url for mobile_exporter.py

sakaia:
This patch is based on #432.
This patch considers the github of caffe2 is moved to its pytorch/caffe2.


soumith:
thank you!

Issue #503, opened 2019-05-20
Pedagogically confusing image size in "NEURAL NETWORKS"


WillRNaylor:
At the start of the tutorial:
https://pytorch.org/tutorials/beginner/blitz/neural_networks_tutorial.html

You have 5x5 convolutions and in both of the Conv layers and then you add the next linear layer explicitly writing that it has size (16 * 5 * 5, 120). Clearly the 16 comes from the number of filters in the prev. layer, and it seems at first obvious that the hard coded 5 * 5 might come from the filter size, but it does not - it should come from the dimension of the input of the image (which is 32 x 32). You could make a note of this, or simply change the size of the filters to 3 x 3 or 7 x 7 or something.

Of course the code currently runs fine, and I see you added the nice pedagogical function `def num_flat_features(self, x)`, which was what helped me realise something was amiss in the code.

Below is the init of the `Net()` class which confused me for reference.

```python
    def __init__(self):
        super(Net, self).__init__()
        # 1 input image channel, 6 output channels, 5x5 square convolution
        # kernel
        self.conv1 = nn.Conv2d(1, 6, 5)
        self.conv2 = nn.Conv2d(6, 16, 5)
        # an affine operation: y = Wx + b
        self.fc1 = nn.Linear(16 * 5 * 5, 120)
        self.fc2 = nn.Linear(120, 84)
        self.fc3 = nn.Linear(84, 10)
```

There is also an issue here, which is maybe related.

https://github.com/pytorch/tutorials/issues/498

Apologies if this is well written, and thanks to all who made the great PyTorch intro pages.

soumith:
this seems very reasonable to attempt to fix, so that future readers wont get affected by it.

Would you be interested in sending a PR to fix the sample code, by editing this file: https://github.com/pytorch/tutorials/blob/master/beginner_source/blitz/neural_networks_tutorial.py ?

WillRNaylor:
So, I just went ahead and made the simplest pull request:

https://github.com/pytorch/tutorials/pull/515

Note that I went for 3x3 filters over 7x7 as the input images are already small (32x32), so with 7x7 we would end up with a 4x4 encoded "image", after the conv. layers. It's 6x6 as you see in the code with the 3x3 filters.

Thanks, and I will take no offence if you think of a better solution!

soumith:
closed via https://github.com/pytorch/tutorials/pull/515

saper0:
[...]

PR #515, opened 2019-05-27
Change to 3x3 conv filters, for ease of reading dimension in net

WillRNaylor:
See here for the original comment.

https://github.com/pytorch/tutorials/issues/503

With 5x5 filters and 5*5 dimension from the number of filters that "span" the 32x32 image (after two layers with maxpooling) those numbers were confusing. Hence the filters are changed to 3x3 size.

Possible downsides of this solution:
 - We do not cover the image as well, as on the final 2x2 maxpool layer goes over an image of size 13x13, and hence cuts it to 12x12. 
 - the comment on line 54 isn't terribly illuminating (but is it kept to a single line).

### DIFF
--- a/beginner_source/blitz/neural_networks_tutorial.py
+++ b/beginner_source/blitz/neural_networks_tutorial.py
@@ -46,12 +46,12 @@ class Net(nn.Module):
 
     def __init__(self):
         super(Net, self).__init__()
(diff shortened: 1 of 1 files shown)


soumith:
thanks @WillRNaylor !

PR #528, opened 2019-06-10
Add test_seq.to(device) in the hybrid frontend tutorial

shhshn:
which is needed when using
```python
device = torch.device("cuda:0")
```
instead of CPU as in the tutorial as
```python
device = torch.device("cpu")
```


soumith:
thank you!

PR #558, opened 2019-07-17
remove Variable which is deprecated

9bow:
Removed mention of `Variable` which is deprecated.
Also, modified a link to the `Function` more detail.


soumith:
thanks!

PR #559, opened 2019-07-17
Update Super Resolution Tutorial to Use ONNXRuntime

lara-hdr:


advanced_source/super_resolution_with_onnxruntime.py
```diff
@@ -0,0 +1,285 @@
...
+                               export_params=True,        # store the trained parameter weights inside the model file
+                               opset_version=10,          # the onnx version to export the model to
+                               do_constant_folding=True,  # wether to execute constant folding for optimization
+                               input_names = ['input'],   # the model's input names
+                               output_names = ['output'], # the model's output names
+                               dynamic_axes={'input' : {0 : 'batch_size'},    # variable lenght axes
+                                             'output' : {0 : 'batch_size'}})
+
+######################################################################
+# ``torch_out`` is the output after executing the model. Normally you can
```
houseroad: After switching to torch.onnx.export, you can call the model with the same input and get example output, which can be used to verify the exported model.

advanced_source/super_resolution_with_onnxruntime.py
```diff
@@ -0,0 +1,285 @@
...
+batch_size = 1    # just a random number
+
+# Initialize model with the pretrained weights
+map_location = lambda storage, loc: storage
+if torch.cuda.is_available():
+    map_location = None
+torch_model.load_state_dict(model_zoo.load_url(model_url, map_location=map_location))
+
+# set the train mode to false since we will only run the forward pass.
+torch_model.train(False)
```
houseroad: you can also mention torch_model.eval() has the same affect as torch_model.train(False)

houseroad: Also explain why calling this method is important, because it will turn model to inference mode. Otherwise, ops like dropout, batchnorm may behave differently.

advanced_source/super_resolution_with_onnxruntime.py
```diff
@@ -0,0 +1,285 @@
...
+
+ONNXRuntime is a performance-focused engine for ONNX models,
+which inferences efficiently across multiple platforms and hardware
+(Windows, Linux, and Mac and on both CPUs and GPUs).
+ONNXRuntime has proved to considerably increase performance over
+multiple models as explained `here
+<https://cloudblogs.microsoft.com/opensource/2019/05/22/onnx-runtime-machine-learning-inferencing-0-4-release>`__
+
+For this tutorial, you will need to install `onnx <https://github.com/onnx/onnx>`__
+and `onnxruntime <https://github.com/microsoft/onnxruntime>`__.
```
houseroad: Do we want to mention onnxruntime requires at least python 3.5?

advanced_source/super_resolution_with_onnxruntime.py
```diff
@@ -0,0 +1,285 @@
...
+
+from PIL import Image
+import torchvision.transforms as transforms
+
+img = Image.open("./_static/img/cat.jpg")
+
+resize = transforms.Resize([224, 224])
+img = resize(img)
+
+img_ycbcr = img.convert('YCbCr')
```
houseroad: Provide more comments about the preprocessing part?

custom_directives.py
```diff
@@ -92,7 +92,7 @@ def run(self):
                 intro, _ = sphinx_gallery.gen_rst.extract_intro_and_title(abs_fname, blocks[0][1])
 
             thumbnail_rst = sphinx_gallery.backreferences._thumbnail_div(
-                dirname, basename, intro)
+                dirname, dirname, basename, intro)
```
houseroad: is this change necessary?

advanced_source/super_resolution_with_onnxruntime.py
```diff
@@ -0,0 +1,285 @@
...
+#
+# To learn more details about PyTorch's export interface, check out the
+# `torch.onnx documentation <https://pytorch.org/docs/master/onnx.html>`__.
+#
+
+# Input to the model
+x = torch.randn(batch_size, 1, 224, 224, requires_grad=True)
+
+# Export the model
+torch_out = torch.onnx._export(torch_model,               # model being run
```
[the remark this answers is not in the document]
spandantiwari: I agree that we should try to use `export` as much as possible here.

advanced_source/super_resolution_with_onnxruntime.py
spandantiwari: +1

advanced_source/super_resolution_with_onnxruntime.py
```diff
@@ -0,0 +1,285 @@
...
+# and confirm that the model has a valid schema. 
+
+import onnx
+
+onnx_model = onnx.load("super_resolution.onnx")
+onnx.checker.check_model(onnx_model)
+
+
+######################################################################
+# Now let's create an onnxruntime session. This part can normally be
```
spandantiwari: Maybe we can expand on this a little, saying a few different things:
1) We are using ORT's Python APIs.
2) In ORT we need to create an inference session for the model with chosen configuration parameters (here we use the default config).
3) Once the session is created, we evaluate the model using the run() api.

custom_directives.py
lara-hdr: this was required to generate the html files locally. I was testing to see if it was the reason why the CI was failing.
I reverted the change now.

advanced_source/super_resolution_with_onnxruntime.py
```diff
@@ -0,0 +1,285 @@
...
+
+######################################################################
+# Using the same process as above, we also transferred an interesting new
+# model "SRResNet" for super-resolution presented in `this
+# paper <https://arxiv.org/pdf/1609.04802.pdf>`__ (thanks to the authors
+# at Twitter for providing us code and pretrained parameters for the
+# purpose of this tutorial). The model definition and a pre-trained model
+# can be found
+# `here <https://gist.github.com/prigoyal/b245776903efbac00ee89699e001c9bd>`__.
+# Below is what SRResNet model input, output looks like. |SRResNet|
```
lara-hdr: This model was only mentioned in the previous version. 

We are able to export it as well and match numbers with ORT.
Should I add the exported .onnx model or just mention it like before?

advanced_source/super_resolution_with_onnxruntime.py
lara-hdr: good point, I didn't notice this comment before.
Adding to the next commit.

advanced_source/super_resolution_with_onnxruntime.py
houseroad: I don't think model will help us show more details, let's keep it.

lara-hdr: you mean remove the whole section?

lara-hdr:
> What is the ouptut size we use in this example, I saw the new generated output is much smaller than the original one.

The output size is actually bigger, it is 672x672 (the previous is 402x402).
I computed the output with torch_model and got the same size as well (672x672).

advanced_source/super_resolution_with_onnxruntime.py
```diff
@@ -0,0 +1,285 @@
...
+# ignore this output, but here we will use it to verify that the model we
+# exported computes the same values when run in onnxruntime.
+#
+# But before verifying the model's output with onnxruntime, we will check
+# the onnx model with onnx's API. This will verify the model's structure
+# and confirm that the model has a valid schema. 
+
+import onnx
+
+onnx_model = onnx.load("super_resolution.onnx")
```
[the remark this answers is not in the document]
lara-hdr: lmk if it is too much or not enough details

advanced_source/super_resolution_with_onnxruntime.py
houseroad: Yeah... two models are quite similar.. I didn't see any benefits here... Just remove it... of course, follow the same process as above, we will be able to transfer tons of models :-)

advanced_source/super_resolution_with_onnxruntime.py
```diff
@@ -0,0 +1,305 @@
...
+# and is widely used in image processing or video editing. For this
+# tutorial, we will first use a small super-resolution model with a dummy
+# input.
+#
+# First, let's create a SuperResolution model in PyTorch. `This
+# model <https://github.com/pytorch/examples/blob/master/super_resolution/model.py>`__
+# comes directly from PyTorch's examples without modification:
+#
+
+# Super Resolution model definition in PyTorch
```
houseroad: Add a comment about the expect input and output content (such as size, data layout, required pre/post-processing etc)

jspisak:
@lara-hdr - thanks so much for pulling this update together!

jspisak:
cc @brianjo - please take a look as well before we merge prior to the v1.2 release. Thanks!!

brianjo:
Thanks for your work on this @lara-hdr.

PR #560, opened 2019-07-17
Update our cpp export tutorial

Krovatkin:


advanced_source/cpp_export.rst
```diff
@@ -92,28 +92,33 @@ vanilla Pytorch model::
 
 Because the ``forward`` method of this module uses control flow that is
 dependent on the input, it is not suitable for tracing. Instead, we can convert
-it to a ``ScriptModule`` by subclassing it from ``torch.jit.ScriptModule`` and
-adding a ``@torch.jit.script_method`` annotation to the model's ``forward``
-method::
+it to a ``ScriptModule``. 
+In order to convert the module to the ``ScriptModule``, one needs to 
+enable ``torch.jit._enable_recursive_script``
```
driazati: This flag is gone in nightly and recursive mode is always on

Krovatkin: oopsie, I'll fix this.

brianjo:
@Krovatkin Nick, can you rebase this? Thanks!

Krovatkin:
@brianjo done!

suo:
@Krovatkin is this ready for merge?

Krovatkin:
@suo looks it like it's out of date again. Lemme rebase it again

SethHWeidman:
@brianjo @Krovatkin running the code in the tutorial [here](https://deploy-preview-560--pytorch-tutorials-preview.netlify.com/advanced/cpp_export.html) did not work. See [here](https://nbviewer.jupyter.org/github/SethHWeidman/tutorials/blob/cpp_tutorial/advanced_source/torchscript_tutorial/New_Tutorial.ipynb).

Krovatkin:
@SethHWeidman I'll take a look!

Krovatkin:
@SethHWeidman looks like you are using 1.1.0. This tutorial was updated for **1.2** or **1.1.1** whichever it will be called.
I tried the scripting part of the tutorial against our nightly and it seems to work: https://colab.research.google.com/drive/1BRE0hbKoW9jYwmNtxh4X93ol1xNB5304
I couldn't test tracing `resnet` since it seems to require GPU, but I'll double check that it works in a little bit.

SethHWeidman:
@Krovatkin cool - that is why I noted that I used 1.1.0 🙂. I will check that this works with nightly as well!

SethHWeidman:
@Krovatkin still getting an error using 1.2.0. See [here](https://github.com/SethHWeidman/tutorials/blob/exploratory/advanced_source/cpp_test/CPP%20test.ipynb) for how I exported the model into C++ (doesn't explicitly say how to do so in the tutorial).

When I save the model and try to load into C++, as in the tutorial, I get an error; it simply says "error loading the model". (Step 3 of the tutorial)

Also, the tutorial contains this line: `traced_script_module.save("model.pt")`, which doesn't make sense since we haven't defined an object called `traced_script_module`. 

This is all on `torch` 1.2.0 by the way. Can you please try testing the C++ load sections and see if they work?

Krovatkin:
I'll take a look!

SethHWeidman:
Thanks @Krovatkin.

Update:

* Reading the model into C++ works - you just have to point the command to the path of `model.pt`. Perhaps we could say "run the command `./example-app <PATH-TO-model.pt>`" in the tutorial so people don't get confused if they're just following along.

* This is a minor point; the major point is: the second part, where we try to pass a Tensor of 1s through our model, generated the following error when I ran `./example-app ../model.pt` from within `build`:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  matrix and vector expected, got 2D, 4D at /pytorch/aten/src/TH/generic/THTensorMath.cpp:626
The above operation failed in interpreter, with the following stack trace:
at code/model.py:6:14
op_version_set = 1
def forward(self,
    input: Tensor) -> Tensor:
  _0 = torch.gt(torch.sum(input, dtype=None), 0)
  if bool(_0):
    output = torch.mv(self.weight, input)
             ~~~~~~~~ <--- HERE
  else:
    output = torch.add(self.weight, input, alpha=1)
  return output
Compiled from code at <ipython-input-4-0d89e318784e>:8:20
    def forward(self, input):
        if input.sum() > 0:
          output = self.weight.mv(input)
                   ~~~~~~~~~~~~~~ <--- HERE
        else:
          output = self.weight + input
        return output

Aborted (core dumped)
```

See my `.cpp` file [here](https://github.com/SethHWeidman/tutorials/blob/exploratory/advanced_source/cpp_test/example-app/example-app.cpp). 

Hope this helps.

Krovatkin:
@SethHWeidman  

> Reading the model into C++ works - you just have to point the command to the path of model.pt. Perhaps we could say "run the command ./example-app <PATH-TO-model.pt>" in the tutorial so people don't get confused if they're just following along. 


I'll make the correction!

> where we try to pass a Tensor of 1s through our model, generated the following error when I ran ./example-app ../model.pt from within build:

What's `input`'s shape you are passing in? `mv` expects a vector of shape `(m)` where `m` should match `m` of  weight's shape `(n, m)`

SethHWeidman:
Hey @Krovatkin, I see what you mean, you're right: the problem is that we save two models during the tutorial, and unless people are paying close attention - which, I acknowledge, they should be! - they might try to run `example-app` with the wrong one.

How about we call the models different things to clarify - see [here](https://github.com/SethHWeidman/tutorials/blob/exploratory/advanced_source/cpp_test/CPP_test.ipynb). I think just calling the ResNet model `traced_resnet_model.pt` could help quite a bit.

Tutorial is great by the way - very clear and to-the-point. Thanks for looking into this 🙂

Krovatkin:
@SethHWeidman , Thank you very much for your very detailed feedback! I'll tried to address all the points you brought up in https://github.com/pytorch/tutorials/pull/609 . Please let me know if I missed anything.

PR #568, opened 2019-07-22
Update cifar10_tutorial.py

jspisak:
Some spelling errors fixed..

"That looks waaay better than chance"
"Because your network is realllly small"


soumith:
they were intentional dramatic language, but we're no longer small enough to do silly things ;-)

PR #573, opened 2019-07-24
Update chatbot_tutorial.py

brianjo:
Updated some code to remove eval in the code.


brianjo:
@soumith I think we can merge this unless we need more changes.

PR #577, opened 2019-07-25
chatbot_tutorial.py:   Solve the optimizer cuda call problem

jiangzhonglian:
If you don't configure this string of code, you will get an error when you iterate over the update from 4000_checkpoint.tar:

```
encoder_optimizer.step()  
```

Error message:

```
exp_avg.mul_(beta1).add_(1 - beta1, grad)
RuntimeError: expected backend CPU and dtype Float but got backend CUDA and dtype Float
```

Fix it: https://github.com/pytorch/pytorch/issues/2830

```
model = Model()
model.load_state_dict(checkpoint['model'])
model.cuda()
optimizer = optim.Adam(model.parameters())
optimizer.load_state_dict(checkpoint['optimizer'])
for state in optimizer.state.values():
    for k, v in state.items():
        if isinstance(v, torch.Tensor):
            state[k] = v.cuda()
```


vincentqb:
@brianjo -- can you merge this?

brianjo:
> @brianjo -- can you merge this 

Sure thing. It might be a while as we need to merge some other fixes first. Thanks!

jiangzhonglian:
Cool !  I am so happy to help you a little!
Hahaha, in fact, we also translated [Pytorch Chinese Documentation](https://pytorch.apachecn.org), the pytorch tutorial is the best in-depth tutorial I have learned.

jspisak:
Thank you!!! @jiangzhonglian

PR #580, opened 2019-07-25
Fix some minor issues in Custom C++ and CUDA Extensions

rgommers:
These are all just names that were inconsistently defined.


soumith:
thanks!

PR #537, opened 2019-06-20
Add a new tutorial: Building a Flask API server with PyTorch

avinassh:
I was writing this for my blog, however, I noticed that PyTorch Tutorials does not have one on building an API. So I am submitting here. My apologies, since I did not discuss this here first. I think PyTorch Tutorial could really use this 😄

PR #583, opened 2019-08-01
Fix a typo

avinassh:
From the discussion - https://github.com/pytorch/tutorials/pull/537#discussion_r309808913


brianjo:
Thanks @avinassh. I added this to the fix I did in #585.

PR #587, opened 2019-08-04
Add TensorBoard Tutorial

SethHWeidman:
Adds new Tensorboard tutorial to the 60 minute blitz. Source file [here](https://github.com/SethHWeidman/tutorials/blob/master/beginner_source/blitz/tensorboard_tutorial.py).


SethHWeidman:
Agreed @soumith - especially now that this is a full-length tutorial and not just a quick add-on to the CIFAR-10 one. I like the idea of linking to it at the end of CIFAR-10 now. 

I'll work with @brianjo to figure out what exactly needs to be changed in `tutorials` to add it in.

PR #591, opened 2019-08-05
Mention a quirk with jupyter notebooks

smessmer:


jspisak:
@brianjo @soumith - I was going to add reviewers but doesnt seem like i can either.

PR #599, opened 2019-08-08
Tensorboard tutorial

SethHWeidman:
Adding TensorBoard tutorial off of a new fork.


brianjo:
Let's see how it goes. :)

brianjo:
Don't merge until after I merge the audio tutorial. Thanks!

brianjo:
This is close. Looks like it's just spinning on something in your tutorial.

SethHWeidman:
@brianjo recreated the tutorial as an `.rst`, it now passes all checks. This was never intended to be a tutorial whose code runs and produces output, so we should have created it this way from the start. Ok to merge?

PR #609, opened 2019-08-12
fix confusion w.r.t. which model to serialize and use in cpp_export tutorial

Krovatkin:
Addressing feedback we got from @SethHWeidman .
Mainly, making sure there's no confusion which serialized model should be loaded in a c++ app.

### DIFF
--- a/advanced_source/cpp_export.rst
+++ b/advanced_source/cpp_export.rst
@@ -1,6 +1,8 @@
 Loading a PyTorch Model in C++
 ==============================
 
+**This tutorial was updated to work with PyTorch 1.2**
+
 As its name suggests, the primary interface to PyTorch is the Python
 programming language. While Python is a suitable and preferred language for
 many scenarios requiring dynamism and ease of iteration, there are equally many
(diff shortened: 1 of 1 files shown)


Krovatkin:
@brianjo would it be possible to merge a small update to this tutorial?

@SethHWeidman gave it a try and, unfortunately, it wasn't as clear as we would've hoped 😢

SethHWeidman:
Appreciate you making these changes @Krovatkin.

Could I request one more change? In the line near the end of the tutorial:

```
root@4b5a67132e81:/example-app/build# ./example-app model.pt
```

Could we change this to:

```
root@4b5a67132e81:/example-app/build# ./example-app traced_resnet_model.pt
```

Part of my confusion in going through this tutorial was that the `example-app` we have only works with the ResNet model since we are passing in a 4D Tensor, so if someone tries to run this with `custom_model.pt`, they'll get a confusing error.

Adding a note to this effect - that we are pushing a 4D Tensor through our model here, so this `example-app` will only work with a CNN model - would be even better 🙂

Krovatkin:
> Could we change this to:

Arrgh, I thought `example.cpp` was supposed to use `MyModule.` Fixed it.

brianjo:
I'll merge this after it rebases. Thanks!

SethHWeidman:
Hey @Krovatkin, don't mean to be a pest about this, but I meant the very last time we call `example-app`, where it prints a slice of the first five elements of the result of feeding the Tensor through the model. It still says 

`root@4b5a67132e81:/example-app/build# ./example-app model.pt` 

Could we change that line to:

`root@4b5a67132e81:/example-app/build# ./example-app <path_to_model>/traced_resnet_model.pt` 

Or Brian or I can change directly on `master` after this merges. Thanks!

Krovatkin:
@SethHWeidman ah, you meant the output of a CMake script. Nice catch! Thanks for the update!

PR #611, opened 2019-08-13
add serving batching for flask using service_streamer

Meteorix:
Following the original tutorial, readers will write a server with these two serious issues:
 - One request is served at a time, it is much slower compared to local batch prediction 
 - It will cause cuda out-of-memory error on GPU when there are large concurrent requests

So I add one paragraph to describe how to use [service_streamer](https://github.com/ShannonAI/service-streamer) middleware. It will solve these two issues with a couple of lines of code.

### DIFF
--- a/intermediate_source/flask_rest_api_tutorial.py
+++ b/intermediate_source/flask_rest_api_tutorial.py
@@ -225,7 +225,7 @@ with open("../_static/img/sample_file.jpeg", 'rb') as f:
 # Integrating the model in our API Server
 # ---------------------------------------
 #
-# In this final part we will add our model to our Flask API server. Since
+# In this part we will add our model to our Flask API server. Since
 # our API server is supposed to take an image file, we will update our ``predict``
 # method to read files from the requests:
 #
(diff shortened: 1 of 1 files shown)


jeffreyksmithjr:
Thanks for proposing this addition to the tutorial.

I agree that a production grade model server might need to have batching capabilities as well as support for a high number of concurrent requests. That said, I don't think that this tutorial should be expanded to consider the scope of such production-scale concerns. The tutorial already has some fairly explicit language about the limitations of the example and the differences that production deployment would require. Expanding out into that scope doesn't seem to server the original goal: demonstrate a minimal inference service scenario.

Beyond that, the Service Streamer technology you developed and propose adding to this tutorial is, while cool looking, very young and not widely adopted. I think it's a useful project, and I hope you are successful with it. But I think we should be keeping our scope for a tutorial to widely deployed technologies with a large amount of established production usage, which Flask clearly satisfies but other, younger technologies would not.

My suggestion is that you adapt this material into a blog post elsewhere, and then we can help point others to it, when they are attempting to address the concerns that you raise.

Meteorix:
> Thanks for proposing this addition to the tutorial.
> 
> I agree that a production grade model server might need to have batching capabilities as well as support for a high number of concurrent requests. That said, I don't think that this tutorial should be expanded to consider the scope of such production-scale concerns. The tutorial already has some fairly explicit language about the limitations of the example and the differences that production deployment would require. Expanding out into that scope doesn't seem to server the original goal: demonstrate a minimal inference service scenario.
> 
> Beyond that, the Service Streamer technology you developed and propose adding to this tutorial is, while cool looking, very young and not widely adopted. I think it's a useful project, and I hope you are successful with it. But I think we should be keeping our scope for a tutorial to widely deployed technologies with a large amount of established production usage, which Flask clearly satisfies but other, younger technologies would not.
> 
> My suggestion is that you adapt this material into a blog post elsewhere, and then we can help point others to it, when they are attempting to address the concerns that you raise.

Thanks for your reply. How about raising the concerns in `Next steps` section, and point others to `service_streamer`?  In my opinion, readers should be aware of these 2 issues when they read this tutorial. `service_streamer` is indeed a young repo. But it exactly solves the issues for PyTorch users and it has been well tested and benchmarked.

SethHWeidman:
Hi @Meteorix, we really appreciate you taking the time to incorporate `service_streamer` into this tutorial in a clear, understandable way.

I agree with @jeffreyksmithjr that we should keep the current tutorial clean and focused on widely-used and understood Flask functionality. 

I'd love to see the example presented here turned into a blog post, and then we could point readers to that post in either the "Next Steps" section of this tutorial or another appropriate place. Keep up the great work!

Meteorix:
@jeffreyksmithjr  @SethHWeidman thanks for your advice. I wrote a blog post [Develop Vision Service with Flask and service streamer](https://github.com/ShannonAI/service-streamer/wiki/Develop-Vision-Service-with-Flask-and-service-streamer). Please take a look : )

jeffreyksmithjr:
FYI: I think [this](https://github.com/ShannonAI/service-streamer/wiki/Vision-Recognition-Service-with-Flask-and-service-streamer) is the correct link.

Yep, your tutorial looks good. I haven't had the chance to test it, but I agree with @SethHWeidman  that this could make sense in the Next Steps section. Feel free to add it, in a reopened version of this PR or in a new PR.

Meteorix:
> FYI: I think [this](https://github.com/ShannonAI/service-streamer/wiki/Vision-Recognition-Service-with-Flask-and-service-streamer) is the correct link.
> 
> Yep, your tutorial looks good. I haven't had the chance to test it, but I agree with @SethHWeidman that this could make sense in the Next Steps section. Feel free to add it, in a reopened version of this PR or in a new PR.

It seems I can't reopen this PR, so I create a new one https://github.com/pytorch/tutorials/pull/614

PR #614, opened 2019-08-16
Add serving batching: point to service-streamer tutorial

Meteorix:
As discussed in the previous PR https://github.com/pytorch/tutorials/pull/611

### DIFF
--- a/intermediate_source/flask_rest_api_tutorial.py
+++ b/intermediate_source/flask_rest_api_tutorial.py
@@ -326,21 +326,6 @@ with open("../_static/img/sample_file.jpeg", 'rb') as f:
 # Next steps
 # --------------
 #
-# Before putting the server into production, we need to solve two issues:
-# 
-# - One request is served at a time, it is much slower compared to a local batch prediction 
-# - It will cause CUDA out-of-memory error on GPU when there are large concurrent requests
-#
-# We can cache user requests in batches and schedule the prediction process. 
-# Follow `service-streamer tutorial <https://github.com/ShannonAI/service-streamer/wiki/Vision-Recognition-Service-with-Flask-and-service-streamer>`_
-# you will solve these issues with a few lines of code.
-#
-# .. Note ::
-#    `service-streamer <https://github.com/ShannonAI/service-streamer>`_ is a middleware for web service
-#    of machine learning applications. Queued requests from users are sampled into mini-batches. Service-streamer 
-#    can significantly enhance the overall performance of the web server by improving GPU utilization.
-#
-#
 # The server we wrote is quite trivial and and may not do everything
 # you need for your production application. So, here are some things you
 # can do to make it better:
@@ -365,3 +350,9 @@ with open("../_static/img/sample_file.jpeg", 'rb') as f:
 # - You can also add a UI by creating a page with a form which takes the image and
 #   displays the prediction. Check out the `demo <https://pytorch-imagenet.herokuapp.com/>`_
 #   of a similar project and its `source code <https://github.com/avinassh/pytorch-flask-api-heroku>`_. 
+#
+# - In this tutorial, we only showed how to build a service that could return predictions for 
+#   a single image at a time. We could modify our service to be able to return predictions for 
+#   multiple images at once. In addition, the `service-streamer <https://github.com/ShannonAI/service-streamer>`_
+#   library automatically queues requests to your service and samples them into mini-batches 
+#   that can be fed into your model. You can check out `this tutorial <https://github.com/ShannonAI/service-streamer/wiki/Vision-Recognition-Service-with-Flask-and-service-streamer>`_.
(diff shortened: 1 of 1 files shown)


SethHWeidman:
Thanks @Meteorix. As-written this is to "marketing-y" for our tastes. How about the following, as a final bullet under "Next Steps" rather than a separate Note:

In this tutorial, we only showed how to build a service that could return predictions for a single image at a time. We could modify our service to be able to return predictions for multiple images at once; in addition, the [service-streamer](https://github.com/ShannonAI/service-streamer) library automatically queues requests to your service and samples them into mini-batches that can be fed into your model.

No need to add anything more - interested folks can click the link and see more info.

Meteorix:
> Thanks @Meteorix. As-written this is to "marketing-y" for our tastes. How about the following, as a final bullet under "Next Steps" rather than a separate Note:
> 
> In this tutorial, we only showed how to build a service that could return predictions for a single image at a time. We could modify our service to be able to return predictions for multiple images at once; in addition, the [service-streamer](https://github.com/ShannonAI/service-streamer) library automatically queues requests to your service and samples them into mini-batches that can be fed into your model.
> 
> No need to add anything more - interested folks can click the link and see more info.

@SethHWeidman I don't agree with this point. 

IMO, readers should be aware of the 2 issues I wrote when they read this tutorial. Most PyTorch users are researchers or students, who lack the experience of web development. They follow this tutorial and write this flask server, put it online. They don't even have the idea that this server is defective.

For other bullets under "Next Steps", they are more like **`features`**. Users will come back to those links when they need to add those features. But for the 2 **`issues`** I wrote, users won't know they need to solve them, because of a lack of background knowledge. It's this tutorial's responsibility to let users know: take care of these 2 issues and here is a solution.

BTW, I don't know what this word "marketing-y" means. The word "marketing" makes me feel offended. I am the author of several [larger opensource projects](https://github.com/Meteorix) and have no need to "sell" anything in `service-streamer`. The reason I wrote `service-streamer` is as described above. I found most PyTorch users around me have no idea why their flask services are slow and sometimes crash. So I wrote `service-streamer` to solve their problems and opensource to share.

jeffreyksmithjr:
I see both of your points, @SethHWeidman and @Meteorix . Let me see if I can sketch out where I think we should be trying to collectively get to.

**Productionization**
I 100% agree with you @Meteorix , that building production capable ML infrastructure is not a common skill. I think more people should be learning more about this topic. And I've put my time where my mouth is, spending a couple of years writing a [book about production AI](https://www.manning.com/books/machine-learning-systems). We as a field of ML engineering and specifically within the PyTorch ecosystem need more resources like the one you're trying to develop.

**Tutorial Scope**
So, what we do with the above productionization material needs to be guided by our overall strategy around structuring materials for developers attempting to use PyTorch for different purposes. Internally to the core PyTorch team, we identified a need for a very simple, focused tutorial around how to use common infrastructure to serve models in simplistic use cases. Thankfully, @avinassh , a community member, decided to fill exactly the hole we had identified. It covers a minimal service that could be useful for exploratory work in academic, prototyping, and research settings. Given that scope, it fulfills its original goals as written and is no way "defective." Depending on the users goals, the limitations of this simple service design may not be "issues" requiring a "solution." Productionization-type hardening is only justified if you are attempting to take something to production. I make a longer version of this point in my book on production AI architectures, and I'm happy to send you a copy if you want to review the rationale.

**Production Serving Resources**
What you've written, I will characterize as "opinionated," which is basically all I think @SethHWeidman  was trying to say. It is one of many ways to take a model to a more production capable serving solution. We in the PyTorch ecosystem would typically try to not say that there is only one way, when the community is building multiple solutions. In this case, you've chosen to show how to incrementally evolve a simple service. That's a valid route to production. There are also other routes such as OSS model servers like MLFlow and Kubeflow or cloud serving solutions from AWS, GCP, and others. All are things we need more supporting documentation around.

**Next Steps** 
I think something like @SethHWeidman 's original next steps still seem like the right thing to do: link out to the resource you wrote with less opinionated language clarifying when these additional steps might be useful. Does that sound reasonable @Meteorix ?

Meteorix:
@jeffreyksmithjr thanks for your detailed explanation, it makes senses to me. I'll push a new commit based on @SethHWeidman 's text. 

I am focusing on ML engineering now and will be for future years. It will be my pleasure to review your book on production AI architectures, please send me lxhustauto@gmail.com. 

I like the PyTorch ecosystem and am willing to make contributions to the ecosystem. That's what I am doing here and in several other PRs to PyTorch repos. So could you give me advice on where to spread the word about `service-streamer`?

SethHWeidman:
Thanks for making these changes, @Meteorix!

To answer your question about where else you can spread the word about `service-streamer`: we are starting to work on reorganizing our `tutorials` and `examples` repos - specifically either adding a section within examples or another repo called `recipes` where we can store short, copy-pastable snippets of both PyTorch functionality and additional libraries. Your extra `service-streamer` material could fit well there, so stay tuned!

jeffreyksmithjr:
@Meteorix thanks for your email. I'll follow up on topics beyond the scope of this issue outside of email.

Issue #540, opened 2019-06-21
small fix needed to make the dataset compatible with random_split


barthelemymp:
Hello,

the __getitem__ of the custom dataset in this tutorial uses idx as int or list. However if we want to use the function random_split in order to get to train and test dataset will not work. Indeed it will feed the __getitem__ with a tensor.
adding :

```
    def __getitem__(self, idx):
        if torch.is_tensor(idx):
            idx = idx.tolist()
```
solve this problem. 

regards b

vincentqb:
Which tutorial?

barthelemymp:
this tutorial:
https://pytorch.org/tutorials/beginner/data_loading_tutorial.html

jlin27:
Thanks @barthelemymp  and @vincentqb !

PR #616, opened 2019-08-16
Enabling tensor index from random_split

vincentqb:
Converting the index from `random_split` given in tensor form requested in #540. It seems like a good place to show how a user could enable tensor index.


brianjo:
Will merge this right after it rebases. Thanks @vincentqb!

PR #633, opened 2019-08-23

standbyme:


standbyme:
@jspisak Does this PR look good?
Thanks~

SethHWeidman:
Hey @standbyme, I see where you're coming from with this; the output of an encoder that is fed into a decoder in a Seq2Seq model _is_ analogous to a hidden state. However, I think it is still accurate to call it an "output", and doing so is also consistent with the rest of this tutorial, so I'd rather keep it as-is.

What do you think? Do you mind if I close this?

standbyme:
@SethHWeidman Thanks for your reply. I read the rest code more carefully and find `hidden state` is accurate.
: )

PR #644, opened 2019-09-04
Use PackedTensorAccessor32 in cpp extension tutorial

peterbell10:
Closes pytorch/pytorch#19268

See pytorch/pytorch#25654 for the corresponding renaming of `PackedTensorAccessor`.


TimZaman:
Nit - the tutorial refers to code that's different from the tutorial as it was changed in this PR; https://github.com/pytorch/extension-cpp/blob/master/cuda/lltm_cuda_kernel.cu eg in my copy-pasta i suddenly ended up with the legacy `PackedTensorAccessor`

PR #648, opened 2019-09-06
Reorganize Production Usage section

SethHWeidman:
Reorganized the production tutorials based on discussions with @mrshenli, @suo, @jamesr66a. 

In particular, the previous "Production Usage" section is now split into two sections:

1. Deploying PyTorch Models in Production
2. Parallel and Distributed Training

The tutorials within each of these sections are now numbered.

Also, added text to the Flask tutorial explaining that it is the first in a series of tutorials on deploying to production.

Finally, deleted references to the "TCP" and "MPI" backends in the "Writing Distributed Applications..." tutorial.


intermediate_source/model_parallel_tutorial.py
```diff
@@ -23,11 +23,22 @@
 of model parallel. It is up to the readers to apply the ideas to real-world
 applications.
 
-**Recommended Reading:**
-
--  https://pytorch.org/ For installation instructions
--  :doc:`/beginner/blitz/data_parallel_tutorial` Single-Machine Data Parallel
--  :doc:`/intermediate/ddp_tutorial` Combine Distributed Data Parallel and Model Parallel
+**Note**: you may be wondering when you should use `DataParallel <https://pytorch.org/docs/stable/nn.html#torch.nn.DataParallel>`_
```
mrshenli: Both `DataParallel` and `DistributedDataParallel` implement data parallel training. The difference is that `DataParallel` is single-process multi-thread, and only works on a single machine, while `DistributedDataParallel` is multi-process and works for both single- and multi- machine training. Even for single machine training `DistributedDataParallel` is expected to be faster than `DataParallel`.

So, the text here will need some change, e.g., sth like:

you may be wondering when you should use model parallel vs data parallel.

- If your **model** is too large to fit on one GPU, use model parallel.
- If your **data** is too large and too slow to train, use data parallel (`DataParallel` or `DistributedDataParallel`).
- If both your **model** and your **data** are too large, combine model parallel with `DistributedDataParallel`. More specifically, each `DistributedDataParallel` process uses model parallel, and all processes collectively use data parallel.

intermediate_source/dist_tuto.rst
```diff
@@ -631,7 +568,7 @@ with rank 0 and follow the procedure described above.
 
 .. code:: python
 
-    dist.init_process_group(init_method='tcp://10.1.1.20:23456', rank=args.rank, world_size=4)
+    dist.init_process_group(init_method='gloo://10.1.1.20:23456', rank=args.rank, world_size=4)
```
mrshenli: I am not sure if the `init_method` for `tcp` and `gloo` accept the same format. @pietern could you please help to take a look?

SethHWeidman: Thanks Shen. @pietern, context here is that this tutorial didn't work for me when I ran it with either the TCP or MPI backend.

We can make the edits to this tutorial in a separate commit.

intermediate_source/model_parallel_tutorial.py
```diff
@@ -23,11 +23,22 @@
...
+- If your **model** is small enough to fit on one GPU, don't use either; using ``DataParallel``
+  in this case will actually make your training go slower.
+- If your model is too large for one GPU, use ``DataParallel``
+- If your **data** is too large for one machine, use ``DistributedDataParallel`` (DDP)
+- DDP also allows for multiple processes to be run at once,
+  where each process uses multiple GPUs; if you are comfortable with multiprocessing,
+  using DDP even if your data is just on one machine can still be faster than using
+  ``DataParallel``.
+
+We exclusively cover `DataParallel <https://pytorch.org/docs/stable/nn.html#torch.nn.DataParallel>`_ here;
```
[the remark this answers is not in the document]
SethHWeidman: @mrshenli, what is the difference between Model Parallel and `DataParallel`?

SethHWeidman:
@mrshenli updated the wording in the Model Parallel tutorial, please take a look. In particular, make sure I got the difference between "model parallel" and "data parallel" right. Are you sure this tutorial should be called "Model Parallel Best Practices" instead of "Model Parallel and DataParallel Best Practices"? Up to you.

Also, I reverted the distributed training tutorial to its previous version. I'll create a separate issue to fix that one.

jamesr66a:
Looks great! My only comment is i'm torn over whether the "Loading a TorchScript Model in C++" tutorial should keep the existing C++ logo or switch to the TorchScript logo. Otherwise, LGTM

intermediate_source/model_parallel_tutorial.py
```diff
@@ -23,11 +23,33 @@
...
 
-**Recommended Reading:**
-
--  https://pytorch.org/ For installation instructions
--  :doc:`/beginner/blitz/data_parallel_tutorial` Single-Machine Data Parallel
--  :doc:`/intermediate/ddp_tutorial` Combine Distributed Data Parallel and Model Parallel
+**Comparison between DataParallel and DistributedDataParallel**
+
+You may be wondering what the differences between
+`DataParallel <https://pytorch.org/docs/stable/nn.html#torch.nn.DataParallel>`_
```
mrshenli: This tutorial does not cover `DataParallel`. `DataParallel` **replicates** model to different GPUs, model parallel **splits** model to different GPUs. Say a model `m` contains 10 layers, when using `DataParallel`, each GPU will have a replica of those 10 layers. When using model parallel on two GPUs, each GPU could host 5 layers.

intermediate_source/model_parallel_tutorial.py
```diff
@@ -23,11 +23,33 @@
...
+(covered in this tutorial) and
+`DistributedDataParallel <https://pytorch.org/docs/stable/_modules/torch/nn/parallel/distributed.html#DistributedDataParallel>`_
+(covered in the next tutorial) are. First, the similarity: both ``DataParallel``
+and ``DistributedDataParallel``` implement data parallel training.
+
+- The difference is that ``DataParallel`` is single-process multi-thread,
+  and only works on a single machine, while ``DistributedDataParallel`` is
+  multi-process and works for both single- and multi- machine training.
+  Thus, even for single machine training, ``DistributedDataParallel``
+  is expected to be faster than ``DataParallel``.
```
mrshenli: Shall we move the comparison above (DataParallel vs DistributedDataparallel) to `DistributedDataparallel`'s tutorial? This model only focuses on model parallel. It might be easier to understand to mention less concepts.

intermediate_source/model_parallel_tutorial.py
```diff
@@ -23,11 +23,33 @@
...
+
+Here's our guideline on when to use each:
+
+- If your **model** is too large to fit on one GPU, use model parallel -
+  that is, splitting your model across multiple GPUs as described
+  `here <https://pytorch.org/tutorials/beginner/blitz/data_parallel_tutorial.html>`_
```
mrshenli: link is wrong. It points to `DataParallel` tutorial. Model parallel is covered in this tutorial.

intermediate_source/model_parallel_tutorial.py
```diff
@@ -23,11 +23,33 @@
...
+- If your data is too large and thus makes your model take too long to
+  train, use data parallel (DataParallel or DistributedDataParallel).
+- If both your model and your data are too large, combine model parallel
```
mrshenli: If both your model and your data are too large, combine model parallel (splitting the model on multiple GPUs in the same process) with `DistributedDataParallel` (splitting data across multiple processes). ...

intermediate_source/model_parallel_tutorial.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """
-Model Parallel Best Practices
-*************************************************************
+1. Model Parallel Best Practices
+================================
 **Author**: `Shen Li <https://mrshenli.github.io/>`_
 
 Data parallel and model parallel are widely-used in distributed training
```
mrshenli: It's my bad to mention data parallel in the first sentence in this tutorial. This might confuse people and make them think this tutorial is talking about data parallel. No, this tutorial only talks about model parallel.

intermediate_source/model_parallel_tutorial.py
SethHWeidman: Got it! 

How about we cover this difference, between model parallel and data parallel, in this tutorial, and `DataParallel` vs. DDP in the next tutorial? Will do this in the revised version.

intermediate_source/model_parallel_tutorial.py
SethHWeidman: Yes, I agree!

SethHWeidman:
@jamesr66a I got you 🙂
torchscript_to_cpp

@mrshenli believe all the wording is now resolved. Let me know of any other requested changes. Thanks for your help working through this!

intermediate_source/dist_tuto.rst
pietern: This is the initialization method. There is no `gloo://` initialization method, so I would assume running this will yield an error. The `backend` kwarg may be specified and can be set to `gloo`, `nccl`, or `mpi`. The initialization method is unrelated to which of the backends you pick.

intermediate_source/ddp_tutorial.rst
```diff
@@ -19,13 +19,33 @@ every module replica, i.e., no module replication within a process. The code in
...
+Before we dive in, let's clarify why, despite the added complexity, you would
+consider using ``DistributedDataParallel`` over ``DataParallel``, remembering
+that **model parallel** (covered in the
+`prior tutorial <https://pytorch.org/tutorials/intermediate/model_parallel_tutorial.html>`__)
+is necessary to use in either case if your model is too large to fit on a single
+GPU.
+
+- ``DataParallel`` is single-process, multi-thread, and only works on a single
+  machine, while ``DistributedDataParallel`` is multi-process and works for both
+  single- and multi- machine training. Thus, even for single machine training,
```
mrshenli: nit: Even for single machine traning, where your **data** is small enough to fit on a single machine, ``DistributedDataParallel`` is expected to be faster than ``DataParallel``, as ``DistributedDataParallel`` replicates models upfront instead of in every iteration and Global Interpreter Lock is no longer in the way.

intermediate_source/ddp_tutorial.rst
```diff
@@ -19,13 +19,33 @@ every module replica, i.e., no module replication within a process. The code in
 this tutorial runs on an 8-GPU server, but it can be easily generalized to
 other environments.
 
+Comparison between ``DataParallel`` and ``DistributedDataParallel``
+-------------------------------------------------------------------
+
+Before we dive in, let's clarify why, despite the added complexity, you would
+consider using ``DistributedDataParallel`` over ``DataParallel``, remembering
+that **model parallel** (covered in the
```
mrshenli: `DataParallel` does not work with model parallel yet. For now, model parallel can only be combined with ``DistributedDataParallel``.

PR #651, opened 2019-09-08
Torchtext translation tutorial

SethHWeidman:
George, re-added after prior branch got irrelevant commits mixed up in it (my fault).

Key change is that, responding to your prior comment, I made it clear that the model in this tutorial is different from `nn.Transformer`, and is just an example model using an architecture that is commonly used for language translation.


zhangguanheng66:
@SethHWeidman It would be great if someone approves the PR before landing.

SethHWeidman:
@zhangguanheng66, apologies! I interpreted your prior comment, my trying to address your comment, and your +1 as approval. Did you have any other feedback? I'll incorporate right away.

PR #660, opened 2019-09-16
torch.ByteTensor is depricated; switched to torch.BoolTensor

jakec888:
requirements.txt torch version: `torch==1.2.0`

`mask = torch.ByteTensor(mask)`

returns the following error:

`UserWarning: masked_select received a mask with dtype torch.uint8, this behavior is now deprecated,please use a mask with dtype torch.bool instead.`

to fix, use the following:

`mask = torch.BoolTensor(mask)`


holly1238:
Closing this PR. The current tutorial has this issue fixed so this PR is no longer relevant.

PR #671, opened 2019-09-25
Add tutorials for network analysis and weight transfer

SebastienEske:
Hello,

This adds two tutorials:
- one to analyze the weights, activations and gradients of a network. This can be helpful for debugging or other in depth analysis.
- one to easily transfer weights between two networks even when layer names vary and when layer shapes change. I don't think that it would be worth including in pytorch as a set of functions but it is still very useful so here it is.

I put two tutorials in one pull request to save time for the CI system but I can always remove one if needed.

Signed-off-by: Sebastien ESKENAZI <s.eskenaz@pixelz.com>


SebastienEske:
Hello, any update?

PR #674, opened 2019-09-26
Remove numbers from tutorial names

SethHWeidman:
Remove numbers from tutorials names in new sections, per @jspisak.


SethHWeidman:
Numbers are gone from Deploying PyTorch Models in Production and Parallel and Distributed Training sections.

PR #676, opened 2019-09-30
Add named tensor tutorial

SethHWeidman:
Adding a tutorial by @zou3519 on named tensors for the 1.3 release.

**Note on position of tutorial on landing page**:

To reduce the number of sections with just one tutorial (this PR adds another), I'll create a separate PR after this one lands that groups the Audio, Named Tensor, Reinforcement Learning, and C++ tutorials into a "Special Topics" section. So, I know it looks odd to add a new "Named Tensor" _section_ to the landing page, but let's just leave that for now.


intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,469 @@
...
+# - JIT support; for example, the following will error
+
+@torch.jit.script
+def fn(x):
+    return x
+
+catch_error(lambda: fn(named_tensor))
+
+######################################################################
+# Longer example: Multi-headed attention
```
zou3519: I've been contemplating whether or not to change this example to something shorter. Let me think about this for a bit and get back to you.

SethHWeidman: I personally think it's great, especially since:

1. We have a side-by-side comparison of this example with the code from ParlAI
2. It brings together many of the concepts you've introduced above (e.g. `refine_names` and `align_to`)

zou3519: Resolved: we're keeping the multihead attention example

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+this tutorial is intended as a guide to the functionality that will
+be included with the 1.3 launch. By the end of it, you will be able to:
+
+- Initiate a ``Tensor`` with named dimensions, as well as removing or renmaing those dimensions
+- Understand the basics of how dimension names are propagated through operations
+- See how naming dimensions enables clearer code in two key areas:
+    - Broadcasting operations
+    - Flattening and unflattening dimensions
+
+Finally, we'll put this into practice by writing a multi-headed attention module
```
zou3519: Let's and call it "multi-head attention" rather than "multi-headed attention". I think more people call it the former (googling "multi-headed attention" leads to search results for "multi-head attention")

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+**Author**: `Richard Zou <https://github.com/zou3519>`_
+
+Named Tensors aim to make tensors easier to use by allowing users to associate explicit names
+with tensor dimensions. In most cases, operations that take dimension parameters will accept
+dimension names, avoiding the need to track dimensions by position. In addition, named tensors
+use names to automatically check that APIs are being used correctly at runtime, providing extra
+safety. Names can also be used to rearrange dimensions, for example, to support
+"broadcasting by name" rather than "broadcasting by position".
+
+this tutorial is intended as a guide to the functionality that will
```
zou3519: Capital T

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
+# -*- coding: utf-8 -*-
+"""
+Introduction to Named Tensors in PyTorch
```
zou3519: We should note that it is experimental somewhere. I'm not sure where the best place for that is though.

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+imgs = torch.randn(1, 2, 2, 3 , names=('N', 'C', 'H', 'W'))
+
+# Unlike in
+# `the original named tensors blogpost <http://nlp.seas.harvard.edu/NamedTensor>`_,
+# named dimensions are ordered: `tensor.names[i]` is the name of the `i`th dimension of `tensor`.
+#
+# There are two ways rename a ``Tensor``'s dimensions:
+#
+
+print(imgs.names)
```
zou3519: Maybe we should move this to directly under the `imgs = torch.randn(1, 2, 2, 3 , names=('N', 'C', 'H', 'W'))` line. This sections should be talking about renaming a tensor's dimensions.

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+
+# Method #1: set .names attribute
+imgs.names = ['batch', 'channel', 'width', 'height']
+print(imgs.names)
+
+# Method #2: specify new names:
+imgs.rename(channel='C', width='W', height='H')
```
zou3519: tensor.rename performs the rename out-of-place. This means that `imgs` won't actually have the updated names, one has to do `imgs = imgs.rename(channel='C', width='W', height='H')`.

This is an interesting tidbit in pytorch's API; only functions marked with a trailing underscore modify tensors in-place.

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+print(imgs.names)
+
+######################################################################
+# The preferred way to remove names is to call ``tensor.rename(None)``:
+
+imgs.rename(None)
```
zou3519: Ditto, should be `imgs = imgs.rename(None)`, or we can provide a new name for the new tensor.

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+print(unnamed.names)
+
+######################################################################
+# Named tensors do not require that all dimensions be named.
+
+imgs = torch.randn(3, 1, 1, 2, names=('N', None, None, None))
+print(imgs.names)
+
+######################################################################
+# Because named tensors coexist with unnamed tensors, we need a nice way to write named-tensor-aware
```
zou3519: nit: It might be good to stay within 80 characters per line so that someone reading this tutorial directly in the code won't see weird line breaks.

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+# code that works with both named and unnamed tensors. Use ``tensor.refine_names(*names)`` to refine
+# dimensions and lift unnamed dims to named dims. Refining a dimension is defined as a "rename" with
+# the following constraints:
+#
+# - A ``None`` dim can be refined to have any name
+# - A named dim can only be refined to have the same name.
+
+imgs = torch.randn(3, 1, 1, 2)
+named_imgs= imgs.refine_names('N', 'C', 'H', 'W')
```
zou3519: space between named_imgs and equals

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+print(imgs.refine_names(..., 'H', 'W').names)
+
+def catch_error(fn):
+    try:
+        fn()
+    except RuntimeError as err:
+        print(err)
+
+# Tried to refine an existing name to a different name
+print(catch_error(lambda: imgs.refine_names('batch', 'channel', 'height', 'width')))
```
zou3519: Don't need the print because catch_error does the printing already.

Also, this won't throw an error because `refine_names` performs refining out-of-place. `imgs` still is unnamed here.

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+
+######################################################################
+# Most simple operations propagate names. The ultimate goal for named tensors is
+# for all operations to propagate names in a reasonable, intuitive manner. Many
+# common operations have been added at the time of the 1.3 release; here,
```
zou3519: Would be good to link to the docs, but this is a chicken and egg problem...

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+print(output.names)
+
+img0 = named_imgs.select('N', 0)  # get one image
+print(img0.names)
+
+######################################################################
+# Name inference
+# --------------
+#
+# Names are propagated on operations in a two step process called **name inference**. It
```
zou3519: We don't need the "It works as follows:"

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+# works as follows:
+#
+# 1. **Check names**: an operator may check that certain dimensions must match.
```
zou3519: Let's use the wording from the updated notebook. It should be:
```
Check names: an operator may perform automatic checks at runtime that check that certain dimension names must match.
Propagate names: name inference propagates output names to output tensors.
```

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+
+# **Check names**: first, we will check whether the names of these two tensors
+# match. Two names match if and only if they are equal (string equality) or at
+# least one is ``None`` (``None``s are essentially a special wildcard name).
+# The only one of these three that will error, therefore, is ``x+z``:
+
+catch_error(lambda: x + z)
+
+# **Propagate names**: _unify_ the two names by returning the most refined name of
+# the two. With ``x + y``,  ``X`` is more specific than ``None``.
```
zou3519: Let's use the same word for both sentences and change "specific" to "refined"

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+
+print((x + y).names)
+
+######################################################################
+# Most name inference rules are straightforward but some of them (the dot
```
zou3519: Change "the dot product ones" to "matrix multiplication in particular"

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+# product ones) can have unexpected semantics. Let's go through a few more of
+# them.
+#
+# Broadcasting
+# ------------
+#
+# Named tensors do not change broadcasting behavior; they still broadcast by
+# position. However, when checking two dimensions for if they can be
+# broadcasted, the names of those dimensions must match. Two names match if and
+# only if they are equal (string equality), or if one is None.
```
zou3519: Remove "Two names match if and only if they are equal (string equality), or if one is None."

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+
+# Automatic broadcasting: expected to fail
+imgs = torch.randn(6, 6, 6, 6, names=('N', 'C', 'H', 'W'))
+per_batch_scale = torch.rand(6, names=('N',))
+catch_error(lambda: imgs * per_batch_scale)
+
+# Explicit broadcasting: the names check out and the more refined names are propagated.
+imgs = torch.randn(6, 6, 6, 6, names=('N', 'C', 'H', 'W'))
+per_batch_scale_4d = torch.rand(6, 1, 1, 1, names=('N', None, None, None))
+print((imgs * per_batch_scale_4d).names)
```
zou3519: None of these are "explicit broadcasting by names". They're both "automatic broadcasting" in the traditional (positional sense); the first example demonstrates where names prevent incorrect behavior and the second example demonstrates where names don't impede a correct result. Maybe we can clarify this point?

intermediate_source/named_tensor_tutorial.py
SethHWeidman: Hmm, I'll ping Joe Spisak & co. about whether to include the word experimental in the title here.

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+catch_error(lambda: fn(imgs_named))
+
+######################################################################
+# As a workaround, please drop names via `tensor = tensor.rename(None)`
+# before using anything that does not yet support named tensors.
+#
+# Longer example: Multi-headed attention
+# --------------------------------------
+#
+# Now we'll go through a complete example of implementing a common advanced
```
zou3519: Is it advanced?

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+
+        # (II)
+        def prepare_head(tensor):
+            tensor = tensor.refine_names(..., 'T', 'D')
+            return (tensor.unflatten('D', [('H', n_heads), ('D_head', dim_per_head)])
+                          .align_to(..., 'H', 'T', 'D_head'))
+
+        if self_attn:
+            key = value = query
+        elif value is None:
```
zou3519: should probably remove the value is None now... and `assert value is None`

intermediate_source/named_tensor_tutorial.py
SethHWeidman: Update: Soumith feels we should have "experimental" in the title - good call.

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+        v = prepare_head(self.v_lin(value)).rename(T='T_key')
+        q = prepare_head(self.q_lin(query))
+
+        dot_prod = q.div_(scale).matmul(k.align_to(..., 'D_head', 'T_key'))
+        dot_prod.refine_names(..., 'H', 'T', 'T_key')  # just a check
+
+        # (III)
+        # Named tensors doesn't support `==` yet; the following is a workaround.
+        attn_mask = (mask.rename(None) == 0).refine_names(*mask.names)
+        attn_mask = attn_mask.align_as(dot_prod)
```
zou3519: We actually do support == now! Yay! (It got into the release branch)

Could we change
```
# Named tensors doesn't support `==` yet; the following is a workaround.
attn_mask = (mask.rename(None) == 0).refine_names(*mask.names)
attn_mask = attn_mask.align_as(dot_prod)
```

to: 
```
attn_mask = (attn_mask == 0).align_as(dot_prod)
```

zou3519: (And also change the code in III)

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+#
+# The second thing to note is how clearly the code describes what is going on.
+# prepare_head takes the key, query, and value and splits the embedding dim into
+# multiple heads, finally rearranging the dim order to be `[..., 'H', 'T', 'D_head']`.
+# ParlAI implements prepare_head as the following, using `view` and `transpose`
+# operations:
+#
+# **(III) Explicit broadcasting by names**
+
+def prepare_head(tensor):
```
zou3519: Needs to be changed as per https://github.com/pytorch/tutorials/pull/676/files#r332304115

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,518 @@
...
+
+    dot_prod.masked_fill_(attn_mask, -float(1e20))
+
+######################################################################
+# ``mask`` usually has dims ``[N, T]`` (in the case of self attention) or
+# ``[N, T, T_key]`` (in the case of encoder attention) while ``dot_prod``
+# has dims ``[N, H, T, T_key]``. To make ``mask`` broadcast correctly with
+# ``dot_prod``, we would usually `unsqueeze` dims `1` and `-1` in the case of self
+# attention or `unsqueeze` dim `1` in the case of encoder attention. Using
+# named tensors, we can simply align the two tensors and stop worrying about
```
zou3519: Sentence repeated twice

SethHWeidman: Which one? I am blind apparently and don't see it.

intermediate_source/named_tensor_tutorial.py
SethHWeidman: Here, `value is None` is `True`, so we want to run the line `key = value = query`, right?

So we want to remove the `elif` block? That's what I'll do in the next version.

intermediate_source/named_tensor_tutorial.py
SethHWeidman: Perhaps, but that word is unnecessary :)

index.rst
```diff
@@ -98,6 +98,17 @@ Audio
 
     <div style='clear:both'></div>
 
+Named Tensor
+----------------------
+
+.. customgalleryitem::
+   :figure: /_static/img/named_tensor.png
```
zou3519: What is the image?

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,519 @@
...
+# `the original named tensors blogpost <http://nlp.seas.harvard.edu/NamedTensor>`_,
+# named dimensions are ordered: ``tensor.names[i]`` is the name of the ``i`` th dimension of ``tensor``.
+#
+# There are two ways rename a ``Tensor``'s dimensions:
+
+# Method #1: set .names attribute
+imgs.names = ['batch', 'channel', 'width', 'height']
+print(imgs.names)
+
+# Method #2: specify new names:
```
zou3519: Could we add an extra line here that is something like "Note: this changes names out-of-place"?

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,519 @@
...
+#
+# - A ``None`` dim can be refined to have any name
+# - A named dim can only be refined to have the same name.
+
+imgs = torch.randn(3, 1, 1, 2)
+imgs = imgs.refine_names('N', 'C', 'H', 'W')
+print(imgs.names)
+
+# Coerces the last two dims to 'H' and 'W'. In Python 2, use the string '...' instead of ...
+print(imgs.refine_names(..., 'H', 'W').names)
```
zou3519: It would be better if this line acted on unnamed imgs. So something like:
```
imgs = torch.randn(3, 1, 1, 2)
print(imgs.refine_names(..., 'H', 'W').names)
```

Otherwise, if we used the `imgs` from the line before, this line doesn't actually do anything.

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,519 @@
...
+# for all operations to propagate names in a reasonable, intuitive manner. Many
+# common operations have been added at the time of the 1.3 release; here,
+# for example, is `.abs()`:
+
+named_imgs = imgs.refine_names('N', 'C', 'H', 'W')
+print(named_imgs.abs().names)
+
+######################################################################
+# Speaking of operations propogating names, let's quickly cover one
+# of the most important operations in PyTorch.
```
zou3519: Why is matrix multiply one of the "most important operations"? Let's not use value statements and instead leave it to the user to judge if an operator is important to them or not.

Also, can we move this back to the name inference section? It's important to see that there is a check names step and a propagate names step and to understand what matrix multiply does for both. For something like `abs`, that has really simple a name inference rule this isn't necessary.

zou3519: The original order of these sections was:
- Accessors / reduction
- Name inference
  - basic 1D example
  - broadcasting
  - matrix multiply
- explicit broadcasting by names

I understand that it is difficult to go automatic broadcasting to matrix multiply and then to explicit broadcasting by names, but we shouldn't introduce matrix multiply before name inference.

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,519 @@
...
+# `per_batch_scale` to the last dimension of `imgs`.
+
+# Automatic broadcasting: expected to fail
+imgs = torch.randn(6, 6, 6, 6, names=('N', 'C', 'H', 'W'))
+per_batch_scale = torch.rand(6, names=('N',))
+catch_error(lambda: imgs * per_batch_scale)
+
+######################################################################
+# How `should` we perform this broadcasting operation along the first
+# dimension? One way, involving names, would be to explicitly initialize
```
zou3519: Let's not call it "explicit" so that we don't confuse it with the "explicit broadcasting by names" section below. Can we change this line to:

"One way, involving names, would be to name the ``per_batch_scale`` tensor such that it matches ``imgs.names``, as shown below."

And then let's change `per_batch_scale_4d`'s names to `('N', 'C', 'H', 'W')` instead of being partially named.

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,519 @@
...
+# dimensions) by using names. We provide a new ``tensor.align_as(other)`` operation
+# that permutes the dimensions of tensor to match the order specified in
+# ``other.names``, adding one-sized dimensions where appropriate
+# (``tensor.align_to(*names)`` works as well):
+
+imgs = imgs.refine_names('N', 'C', 'H', 'W')
+per_batch_scale = per_batch_scale.refine_names('N')
+
+named_result = imgs * per_batch_scale.align_as(imgs)
+assert torch.allclose(named_result.rename(None), correct_result)
```
zou3519: Add comment that named tensors don't yet work with allclose?

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,519 @@
...
+
+print(weight.grad)  # still unnamed
+
+######################################################################
+# Other supported (and unsupported) features
+# ------------------------------------------
+#
+# See here (link to be included) for a detailed breakdown of what is
+# supported with the 1.3 release, what is on the roadmap to be supported soon,
+# and what will be supported in the future but not soon.
```
zou3519: The link in the documentation that I'm writing doesn't have either of "what is on the roadmap to be supported soon,
# and what will be supported in the future but not soon." so maybe we should remove those two clauses?

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,519 @@
...
+def fn(x):
+    return x
+
+catch_error(lambda: fn(imgs_named))
+
+######################################################################
+# As a workaround, please drop names via ``tensor = tensor.rename(None)``
+# before using anything that does not yet support named tensors.
+#
+# Longer example: Multi-headed attention
```
zou3519: Rename to "multi-head attention"

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,519 @@
...
+# --------------------------------------
+#
+# Now we'll go through a complete example of implementing a common
+# PyTorch ``nn.Module``: multi-headed attention. We assume the reader is already
```
zou3519: This section needs renames from multi-headed -> multi-head

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,519 @@
...
+# Running the example
+# -------------------
+
+n, t, d, h = 7, 5, 2 * 3, 3
+query = torch.randn(n, t, d, names=('N', 'T', 'D'))
+mask = torch.ones(n, t, names=('N', 'T'))
+attn = MultiHeadAttention(h, d)
+output = attn(query, mask=mask)
+print(output.names)
+
```
zou3519: We're missing the last section from the updated tutorial I posted: 
```
Notice how in the code we did not mention the name of the batch dimension at all. In fact, the code is agnostic to the existence of batch dimensions; we can run the following example-level code:

query = torch.randn(t, d, names=('T', 'D'))
mask = torch.ones(t, names=('T',))
output = attn(query, mask=mask)
output.names
('T', 'D')
```
(https://our.internmc.facebook.com/intern/anp/view/?id=152205 scroll to the end)

zou3519:
The tutorial builds are failing -- should we be concerned?

intermediate_source/named_tensor_tutorial.py
SethHWeidman: Got it. I re-ordered this because I thought the broadcasting stuff all made sense together. I will order it back. Thanks!

SethHWeidman:
@zou3519 the build is failing because named tensor is not yet in the version of `torch` that is `pip install`ed. See [here](https://circleci.com/gh/pytorch/tutorials/17375).

Going to try fixing this by changing `torch` to `git+https://github.com/pytorch/pytorch.git@master` in `requirements.txt`.

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,519 @@
...
+# ------------------------------------------
+#
+# See here (link to be included) for a detailed breakdown of what is
+# supported with the 1.3 release, what is on the roadmap to be supported soon,
+# and what will be supported in the future but not soon.
+#
+# In particular, we want to call out three important features that are not
+# currently supported:
+#
+# - Retaining names when serializing or loading a serialized ``Tensor`` via
```
zou3519: "Retaining names when ..." is a little misleading because it implies that `torch.save` will work but just not retain names. `torch.save` errors out gracefully when passed a named tensor

index.rst
SethHWeidman: named_tensor

intermediate_source/named_tensor_tutorial.py
SethHWeidman: done

intermediate_source/named_tensor_tutorial.py
SethHWeidman: good point, done

intermediate_source/named_tensor_tutorial.py
SethHWeidman: Reading through this again: I see what you were going for: broadcasting and matrix multiply are examples of where name inference can have unexpected semantics. 

I made broadcasting and matmul subheadings, rather than main headings, in the next version, to reinforce this.

intermediate_source/named_tensor_tutorial.py
SethHWeidman: Much better 😀

index.rst
SethHWeidman: What _should_ the logo be? I'm at a loss here

intermediate_source/named_tensor_tutorial.py
SethHWeidman: We'll keep as a to-do to do this after the docs land.

zou3519:
The heading on "broadcasting" is not distinguishable from bold text: 
image

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,525 @@
...
+
+# Apply one transition
+new_state = markov_states @ transition_matrix
+print(new_state.names)
+
+######################################################################
+# As you can see, matrix multiply does not check if the contracted dimensions
+# have the same name.
+#
+# Next, we'll cover two new behaviors that named tensors enable: explicit
```
zou3519: Only one new behavior is listed here

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,525 @@
...
+print(imgs.refine_names(..., 'H', 'W').names)
+
+def catch_error(fn):
+    try:
+        fn()
+    except RuntimeError as err:
+        print(err)
+
+# Tried to refine an existing name to a different name
+catch_error(lambda: imgs.refine_names('batch', 'channel', 'height', 'width'))
```
zou3519: This is not going to work because `imgs` is unnamed. We should name it first:
```
named_imgs = imgs.refine_names('N', 'C', 'H', 'W')
# Tried to refine an existing name to a different name
catch_error(lambda: named_imgs.refine_names('batch', 'channel', 'height', 'width'))
```

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,525 @@
...
+
+imgs = torch.randn(3, 1, 1, 2)
+print(imgs.refine_names('N', 'C', 'H', 'W'))
+
+# Coerces the last two dims to 'H' and 'W'. In Python 2, use the string '...' instead of ...
+print(imgs.refine_names(..., 'H', 'W').names)
+
+def catch_error(fn):
+    try:
+        fn()
```
zou3519: Let's change this to 

```
fn()
assert False
```
Because whenever we run `catch_error`, there should be an error

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,525 @@
...
+######################################################################
+# Accessors and Reduction
+# -----------------------
+#
+# One can use dimension names to refer to dimensions instead of the positional
+# dimension. These operations also propagate names. Indexing (basic and
+# advanced) has not been implemented yet but is on the roadmap. Using the ``named_imgs``
+# tensor from above, we can do:
+
+output = named_imgs.sum(['C'])  # Perform a sum over the channel dimension
```
zou3519: Let's change this to `output = named_imgs.sum('C')`; it is cleaner and the [] are unnecessary

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,525 @@
...
+
+        # (II)
+        def prepare_head(tensor):
+            tensor = tensor.refine_names(..., 'T', 'D')
+            return (tensor.unflatten('D', [('H', n_heads), ('D_head', dim_per_head)])
+                          .align_to(..., 'H', 'T', 'D_head'))
+
+        assert value is None
+        if self_attn:
+            key = value = query
```
zou3519: Sorry, I wasn't clear about what to remove. This should be:
```
if self_attn:
            key = value = query
else:
            assert value is None
            # key and value are the same, but query differs
            key = key.refine_names(..., 'T', 'D')
            value = key
```

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,525 @@
...
+        # Distinguish between query_len (T) and key_len (T_key) dims.
+        k = prepare_head(self.k_lin(key)).rename(T='T_key')
+        v = prepare_head(self.v_lin(value)).rename(T='T_key')
+        q = prepare_head(self.q_lin(query))
+
+        dot_prod = q.div_(scale).matmul(k.align_to(..., 'D_head', 'T_key'))
+        dot_prod.refine_names(..., 'H', 'T', 'T_key')  # just a check
+
+        # (III)
+        attn_mask = (attn_mask == 0).align_as(dot_prod)
```
zou3519: This should be `attn_mask = (mask == 0).align_as(dot_prod)`

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,525 @@
...
+######################################################################
+# Our named tensor variant uses ops that, though more verbose, also have
+# more semantic meaning than ``view`` and ``transpose`` and include enforcable
+# documentation in the form of names.
+#
+# **(III) Explicit broadcasting by names**
+
+def ignore():
+    # (III)
+    attn_mask = (attn_mask == 0).align_as(dot_prod)
```
zou3519: Should be `attn_mask = (mask == 0).align_as(dot_prod)`

Also, there is an extra whitespace here that doesn't match the above code

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,525 @@
...
+attn = MultiHeadAttention(h, d)
+output = attn(query, mask=mask)
+# works as expected!
+print(output.names)
+
+######################################################################
+# The above works as expected. Furthermore, note that in the code we
+# did not mention the name of the batch dimension at all. In fact,
+# the code is agnostic to the existence of the batch dimensions, so
+# that we can run the following example-level code:
```
zou3519: We're using the word "code" a lot here, could we change this to:
```
# The above works as expected. Furthermore, note that in the code we
# did not mention the name of the batch dimension at all. In fact,
# our ``MultiHeadAttention`` module is agnostic to the existence of batch
# dimensions. 
```

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,525 @@
...
+# the two. With ``x + y``,  ``X`` is more refined than ``None``.
+
+print((x + y).names)
+
+######################################################################
+# Most name inference rules are straightforward but some of them can have
+# unexpected semantics. Let's go through a couple you're likely to encounter:
+# broadcasting and matrix multiply.
+#
+# **Broadcasting**
```
zou3519: In markdown, one can use "^^^^^^^^^^^" as a smaller heading I think. It's really easy to confuse this heading with the bold "propagate names" and "check names" above.

SethHWeidman: Thanks, I did not know this!!

intermediate_source/named_tensor_tutorial.py
zou3519: Resolved.

intermediate_source/named_tensor_tutorial.py
zou3519: Resolved

intermediate_source/named_tensor_tutorial.py
```diff
@@ -0,0 +1,530 @@
...
+# - A ``None`` dim can be refined to have any name
+# - A named dim can only be refined to have the same name.
+
+imgs = torch.randn(3, 1, 1, 2)
+print(imgs.refine_names('N', 'C', 'H', 'W'))
+
+# Coerces the last two dims to 'H' and 'W'. In Python 2, use the string '...' instead of ...
+print(imgs.refine_names(..., 'H', 'W').names)
+
+def catch_error(fn):
```
zou3519: Sorry, this is what I meant. This should be:
```
def catch_error(fn):
    try:
        fn()
        assert False
    except RuntimeError as err:
        print(err)
```

intermediate_source/named_tensor_tutorial.py
zou3519: Resolved

PR #686, opened 2019-10-10
Static quantization tutorial

SethHWeidman:
Static quantization tutorial, covering:

- Post-training static quantization
- Quantization-aware training

Based on @raghuramank100's Bento notebook.


advanced_source/static_quantization_tutorial.py
```diff
@@ -0,0 +1,621 @@
...
+
+######################################################################
+# Next, we'll "fuse modules"; this can both make the model faster by saving on memory access
+# while also improving numerical accuracy. While this can be used with any model, this is
+# especially common with quantized models.
+
+print('\n Inverted Residual Block: Before fusion \n\n', float_model.features[1].conv)
+float_model.eval()
+
+# Fusion is optional
```
raghuramank100: Lets remove this comment, it is required for quantization.

advanced_source/static_quantization_tutorial.py
```diff
@@ -0,0 +1,621 @@
...
+
+#####################################################################
+# More on quantization-aware training:
+#
+# - QAT is a super-set of post training quant techniques that allows for more debugging.
+#   For example, we can analyze if the accuracy of the model is limited by weight or activation
+#   quantization.
+# - We can also simulate the accuracy of a quantized model in floating point since
+#   we are using fake-quantization to model the numerics of actual quantized arithmetic.
+# - We can mimic post training quantization easily too.
```
raghuramank100: We can also state that quantization aware training yields an accuracy of over 71% on the entire imagenet dataset, which is close to the floating point accuracy of 71.9%. In addition, we can also state that the simple quantization technique that we first try gives us 63% accuracy and the per-channel technique boosts it to 67%

PR #307, opened 2018-09-17
Updated TOC and tutorial notes about PT1 preview

brianjo:

PR #687, opened 2019-10-10
Reinstate torchaudio tutorial

vincentqb:
Let's see if the tutorial loads correctly with current binaries. pytorch/audio#307 

This partly reverts commit 9cee82507629c5f557dcef0ed508dc973935ad92 from #685.

CC @SethHWeidman @ezyang

### DIFF
--- a/beginner_source/audio_preprocessing_tutorial.py
+++ b/beginner_source/audio_preprocessing_tutorial.py
@@ -1,0 +1,237 @@
+"""
+torchaudio Tutorial
+===================
+
+PyTorch is an open source deep learning platform that provides a
+seamless path from research prototyping to production deployment with
+GPU support.
+
+Significant effort in solving machine learning problems goes into data
+preparation. torchaudio leverages PyTorch’s GPU support, and provides
+many tools to make data loading easy and more readable. In this
+tutorial, we will see how to load and preprocess data from a simple
+dataset.
+
+For this tutorial, please make sure the ``matplotlib`` package is
+installed for easier visualization.
+
+"""
+
+import torch
+import torchaudio
+import matplotlib.pyplot as plt
+
+
+######################################################################
+# Opening a dataset
+# -----------------
+# 
+
+
+######################################################################
+# torchaudio supports loading sound files in the wav and mp3 format. We
+# call waveform the resulting raw audio signal.
+# 
+
+filename = "../_static/img/steam-train-whistle-daniel_simon-converted-from-mp3.wav"
+waveform, sample_rate = torchaudio.load(filename)
+
+print("Shape of waveform: {}".format(waveform.size()))
+print("Sample rate of waveform: {}".format(sample_rate))
+
+plt.figure()
+plt.plot(waveform.t().numpy())
+
+
+######################################################################
+# Transformations
+# ---------------
+# 
+# torchaudio supports a growing list of
+# `transformations <https://pytorch.org/audio/transforms.html>`_.
+# 
+# -  **Resample**: Resample waveform to a different sample rate.
+# -  **Spectrogram**: Create a spectrogram from a waveform.
+# -  **MelScale**: This turns a normal STFT into a Mel-frequency STFT,
+#    using a conversion matrix.
+# -  **AmplitudeToDB**: This turns a spectrogram from the
+#    power/amplitude scale to the decibel scale.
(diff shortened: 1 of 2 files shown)


SethHWeidman:
Thanks @vincentqb! I'll make sure this is merged as soon as the `torchaudio` issue is fixed. Glad you were able to reproduce that one.

Will hold off on merging until the new tutorials we're releasing today are merged, though.

vincentqb:
@SethHWeidman -- I'd like to see if the latest version of torchaudio fixes the original issue. However, I see a lot of tests with "ci/circleci: pytorch_tutorial_build_manager Expected — Waiting for status to be reported" that are not running. Is that expected?

SethHWeidman:
Hey @vincentqb, I can re-run those build scripts in Circle since I'm an admin of this repo, no prob. 

However, I'm still getting the same error when I `import torchaudio` if I have version >=1.3.0 of `torch` installed. Are you able to reproduce this? If not I can walk through with you on a VC.

vincentqb:
> Hey @vincentqb, I can re-run those build scripts in Circle since I'm an admin of this repo, no prob.
> 
> However, I'm still getting the same error when I `import torchaudio` if I have version >=1.3.0 of `torch` installed. Are you able to reproduce this? If not I can walk through with you on a VC.

As mentioned in #307, I am unable to reproduce the error.

SethHWeidman:
Ok, trying this - I do hope it works!! Will re-add the audio tutorial if so.

I suspect that it won't though: the reason is our tutorials `pip install` all the libraries in `requirements.txt` before running the code in the tutorials; and when I `pip install torchaudio` locally (or on my devserver), `pip` still force uninstalls `torch 1.3.0` and installs `torch 1.2.0` instead. 

Is it possible you've fixed the issue on a development version of `torchaudio`, but not the version that is on PyPi? Seems to me like that version is still incompatible with `torch 1.3.0`.

vincentqb:
> Ok, trying this - I do hope it works!! Will re-add the audio tutorial if so.
> 
> I suspect that it won't though: the reason is our tutorials `pip install` all the libraries in `requirements.txt` before running the code in the tutorials; and when I `pip install torchaudio` locally (or on my devserver), `pip` still force uninstalls `torch 1.3.0` and installs `torch 1.2.0` instead.
> 
> Is it possible we've fixed the issue on a development version of `torchaudio`, but not the version that is on PyPi? Seems to me like that version is still incompatible with `torch 1.3.0`.

Indeed. The version on [pip](https://pypi.org/project/torchaudio/#history) needs to be updated to 0.3.1, [here](https://github.com/pytorch/audio/pull/308#issuecomment-542822662) @ezyang. If this is indeed the issue, I would expect an error in the logs about incompatible versions though, as I'm getting locally.

vincentqb:
The version seems to be the problem. In linux:
```
conda create -n pytorch-pip python=3.7
conda activate pytorch-pip
pip install torchaudio
python -c "import torchaudio"
# success

pip install torch==1.3.0
# ERROR: torchaudio 0.3.0 has requirement torch==1.2.0, but you'll have torch 1.3.0 which is incompatible.
python -c "import torchaudio"

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/private/home/vincentqb/anaconda3/envs/pytorch-pip/lib/python3.7/site-packages/torchaudio/__init__.py", line 5, in <module>
    import _torch_sox
ImportError: /private/home/vincentqb/anaconda3/envs/pytorch-pip/lib/python3.7/site-packages/_torch_sox.cpython-37m-x86_64-linux-gnu.so: undefined symbol: _ZN3c1012CUDATensorIdEv
```

SethHWeidman:
Great, @vincentqb, then at least we know the fix: let's get version 0.3.1 on PyPi!

vincentqb:
The code snippet now works.
```
conda create -n pytorch-pip python=3.7
conda activate pytorch-pip
pip install torchaudio
python -c "import torchaudio"
# success

pip install torch==1.3.0
# already satisfied
```
Let's see if the tests pass.

SethHWeidman:
@vincentqb merging!

@ezyang, to prevent this kind of thing from happening in the future, let's make sure we release versions of the domain libraries to PyPi along with our major releases.

vincentqb:
As mentioned in pytorch/audio#308, the new binaries need to be uploaded to pypi and conda. The former is done, but not the latter.

PR #692, opened 2019-10-10
Update dynamic_quantization_tutorial.py

brianjo:
Tweak to quantization script.


SethHWeidman:
Going with #693 instead. Thanks @brianjo !

PR #713, opened 2019-10-29
Add paths for downloaded files to static quantization tutorial

SethHWeidman:
@raghuramank100 adding language making it clear for readers how to download the files associated with this tutorial.

### DIFF
--- a/advanced_source/static_quantization_tutorial.py
+++ b/advanced_source/static_quantization_tutorial.py
@@ -304,7 +304,11 @@ def print_size_of_model(model):
 # ----------------------------------
 #
 # As our last major setup step, we define our dataloaders for our training and testing set.
-# The specific dataset we've created for this tutorial contains just 1000 images, one from
+#
+# ImageNet Data
+# ^^^^^^^^^^^^^
+#
+# The specific dataset we've created for this tutorial contains just 1000 images from the ImageNet data, one from
 # each class (this dataset, at just over 250 MB, is small enough that it can be downloaded
 # relatively easily). The URL for this custom dataset is:
 #
@@ -312,7 +316,7 @@ def print_size_of_model(model):
 #
 #     https://s3.amazonaws.com/pytorch-tutorial-assets/imagenet_1k.zip
 #
-# To download this data locally using Python, then, you could use:
+# To download this data locally using Python, you could use:
 #
 # .. code:: python
 #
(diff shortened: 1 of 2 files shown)


SethHWeidman:
@raghuramank100 merging!

Closes PyTorch issue #28844.

advanced_source/static_quantization_tutorial.py
```diff
@@ -348,7 +374,8 @@ def prepare_data_loaders(data_path):
     return data_loader, data_loader_test
 
 ######################################################################
-# Next, we'll load in the pre-trained MobileNetV2 model
+# Next, we'll load in the pre-trained MobileNetV2 model. Similarly to the data about, the file with the pre-trained
+# weights is stored at ``https://s3.amazonaws.com/pytorch-tutorial-assets/mobilenet_quantization.pth``:
```
raghuramank100: This is confusing to end users. Let us use the file from torchvision at: https://github.com/pytorch/vision/blob/master/torchvision/models/mobilenet.py#L9

Also, lets call this file something else: not mobilenet_quantization.pth, might be better to call it mobilenet_pretrained_float.pth'

advanced_source/static_quantization_tutorial.py
```diff
@@ -306,7 +306,33 @@ def print_size_of_model(model):
 # As our last major setup step, we define our dataloaders for our training and testing set.
 # The specific dataset we've created for this tutorial contains just 1000 images, one from
 # each class (this dataset, at just over 250 MB, is small enough that it can be downloaded
-# relatively easily). These functions mostly come from
+# relatively easily). The URL for this custom dataset is:
```
raghuramank100: We should also add information on how one can download imagenet for pytorch here. We could say that:
If you want to download the entire imagenet dataset, please follow instructions at:

https://pytorch.org/docs/stable/torchvision/datasets.html#imagenet

SethHWeidman:
@raghuramank100 tried to address these changes, see [here](https://deploy-preview-713--pytorch-tutorials-preview.netlify.com/advanced/static_quantization_tutorial.html)

z-a-f:
@SethHWeidman There is an issue with showing the tutorial version (#28938), which means the users will not see it until the next major release.. Is there a way to force this change to either the current branch or to the 1.3.1?

SethHWeidman:
@z-a-f sorry, users will not see _what_ until the next major release? The tutorial version? Meaning if we update the version to 1.3.1, it will still show as 1.3.0?

In case there's any confusion, these changes are live; see [here](https://pytorch.org/tutorials/advanced/static_quantization_tutorial.html).

Help me understand what the issue is and what you think we should do. Thanks!

PR #733, opened 2019-11-25
Use `cmake --build .` instead of `make`

peterjc123:
Targets https://github.com/pytorch/pytorch/issues/30379.


peterjc123:
cc @ezyang

peterjc123:
cc @soumith

PR #736, opened 2019-11-27
C++11 -> C++14

smessmer:


_static/torchvision_finetuning_instance_segmentation.ipynb
```diff
@@ -348,8 +348,8 @@
...
+            "x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -DWITH_CUDA -I/content/vision/torchvision/csrc -I/usr/local/lib/python3.6/dist-packages/torch/include -I/usr/local/lib/python3.6/dist-packages/torch/include/torch/csrc/api/include -I/usr/local/lib/python3.6/dist-packages/torch/include/TH -I/usr/local/lib/python3.6/dist-packages/torch/include/THC -I/usr/local/cuda/include -I/usr/include/python3.6m -c /content/vision/torchvision/csrc/vision.cpp -o build/temp.linux-x86_64-3.6/content/vision/torchvision/csrc/vision.o -O0 -DTORCH_API_INCLUDE_EXTENSION_H -DTORCH_EXTENSION_NAME=_C -D_GLIBCXX_USE_CXX11_ABI=0 -std=c++14\n",
```
nairbv: D_GLIBCXX_USE_CXX11_ABI looks like it's still referencing CXX11. Is the 14 equivalent just `D_GLIBCXX_USE_CXX14_ABI` ?

nairbv: same in a few places below as well

nairbv:
I'm not sure if the change to `torchvision_finetuning_instance_segmentation.ipynb` is correct. Earlier in the output I see `-std=c99` as well. It looks like this is the actual output of building torchvision, which maybe has other dependencies? We might want to actually run the build and copy/paste the actual output instead of attempting to change the existing one.

_static/torchvision_finetuning_instance_segmentation.ipynb
smessmer: No, there was an ABI change in gcc between c++03 and c++11 and this flag allows you to select which ABI you want. There was no such change and there is no such flag between c++11 and c++14

advanced_source/cpp_frontend.rst
```diff
@@ -5,7 +5,7 @@ The PyTorch C++ frontend is a pure C++ interface to the PyTorch machine learning
 framework. While the primary interface to PyTorch naturally is Python, this
 Python API sits atop a substantial C++ codebase providing foundational data
 structures and functionality such as tensors and automatic differentiation. The
-C++ frontend exposes a pure C++11 API that extends this underlying C++ codebase
+C++ frontend exposes a pure C++14 API that extends this underlying C++ codebase
```
jamesr66a: There's also the CMake property at line 147

holly1238:
@nairbv @jamesr66a is this PR good to be merged or are there additional updates that need to be made?

nairbv (approved):
looks fine to me

PR #744, opened 2019-12-03
Fix decoder dropout value type

XingwenZhang:
Decoder's param dropout should be float rather than int


holly1238:
Closing this PR. The tutorial had been updated such that this PR is no longer relevant.

PR #795, opened 2019-12-17
Update neural_style_tutorial.py; allow diff sizes

jsimonlane:
This pull requests allows users to use images of different sizes -- ie, it removes the constraint that the style image and the content image need to be the same size. This is helpful as it makes playing around with the code easier.

We previously needed this line, due to the way the optimization model was constructed. Namely, the MSE error (content loss) was calculated on 'style' images' in the initial creation of the optimization model, triggering an error. However, such a calculation was unnecessary in the style case.


holly1238:
Hi @jsimonlane! Can you sign the CLA agreement so that I can merge the pull request? Thanks!

PR #801, opened 2019-12-29
Changing the DCGAN Face Notebook to be able to run it on Google colab.

JonathanSum:
The original DCGAN Face Notebook Colab notebook was not written for running it on Colab, but it was written for running locally on ubuntu . For example, the data root is under the ubuntu local directory.

I think it is important to make it able to run for the colab since I know some students that they do not have GPU. But they can run it on colab to use Google's GPU.

I have also checked. It can now use the DC GAN face to make many fake faces on my laptop. 

Watch it on google drive: https://colab.research.google.com/drive/1MqChDo1qgHvRBtMNM0nS8q9ZCmDIQ8bL#scrollTo=PT4LZIAttGNV


JonathanSum:
Another code added in text_sentiment_ngrams_tutorial.ipynb:
First is fixing the typo that makes the notebook fails to run. #803 
Second is adding the missing library Torch Text.

JonathanSum:
adding the missing file: torchtext_translation_tutorial.ipynb
I have also written some code for adding the missing library in the notebook.

albanD:
@jlin27 is checking this and will get back to you!

JonathanSum:
@albanD or @jlin27, I just want to remind any of you that the reinforcement Colab notebook still has an issue. It seems Colab uses HTML instead of OpenGL to capture the screen of gym. 
I tried to fix it, there is some hacks that at least print out the screen of gym. But it seems the google has their official solution to solve that. Thus, for that, I suggest to google search the solution on StackOverflow(at least I saw some official solutions) and fix that reinforcement Colab notebook

PR #805, opened 2020-01-02
Fix multiple typos in beginner tutorial: nn_tutorial.py

SimonCW:
Fixed multiple minor typos in the beginner tutorial _What is torch.nn really?_.


brianjo:
Thanks for those fixed @SimonCW!

PR #478, opened 2019-04-11
[Don't merge] Adding DistributedDataParallel Tutorial

mrshenli:
Requires pytorch/pytorch#18953 and pytorch/pytorch#19271 to support DDP wrapping model parallel. Putting this PR up for early content review.

cc @pietern 

To avoid the sphinx/multiprocessing/ipython error, Python file in this PR has been converted to rst format in #488. Leaving this PR here for testing purpose.

PR #807, opened 2020-01-03
Adding an RPC Tutorial

mrshenli:


intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+This section describes steps to build a toy distributed reinforcement learning
+model using RPC to solve CartPole-v1 from `OpenAI Gym <https://gym.openai.com>`__.
+The policy code is mostly borrowed from the existing single-thread
+`example <https://github.com/pytorch/examples/blob/master/reinforcement_learning>`__
+as shown below. We will skip details of the ``Policy`` design, and focus on RPC
+usages.
+
+.. code:: python
+
+    class Policy(nn.Module):
```
rohan-varma: general q: Should the tutorials be "copy pastable", i.e. should have all the imports included and some instructions on what's needed to actually be able to run the code? I've found it useful in some pytorch tutorials where I can just copy + paste the code and it works, but not all tutorials are setup this way.

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+            x = F.relu(x)
+            action_scores = self.affine2(x)
+            return F.softmax(action_scores, dim=1)
+
+Let's first prepare a helper function to call a function on a local ``RRef``. It
+might look unnecessary at the first glance, as you could simply do
+``rref.local_value().some_func(args)`` to run the target function. The reason
+for adding this helper function is because there is no way to get a reference
+of a remote value, and ``local_value`` is only available on the owner of the
+``RRef``.
```
rohan-varma: Are we eventually going to change this and provide a better design to call a method on an RRef? If so, might be worth mentioning so someone reading in the future uses the better API

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+                print("Training epoch {}".format(epoch))
+
+
+Finally, let's add some glue code to launch the parameter server and the trainer
+processes.
+
+
+.. code:: python
+
+    def run_ps():
```
rohan-varma: Don't we need to `rpc.init_rpc` on the node that hosts the param server?

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+        rpc.shutdown()
+
+    mp.set_start_method('spawn')
+    ps = mp.Process(target=run_worker, args=("ps", 0, run_ps, 2))
+    ps.start()
+
+    trainer = mp.Process(target=run_worker, args=("trainer", 1, run_trainer, 2))
+    trainer.start()
+    ps.join()
+    trainer.join()
```
rohan-varma: Would it be useful to show some output of the training process?

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+        pass
+
+    def run_worker(name, rank, func, world_size):
+        os.environ['MASTER_ADDR'] = 'localhost'
+        os.environ['MASTER_PORT'] = '29500'
+        rpc.init_rpc(name, rank=rank, world_size=world_size)
+
+        func()
+
+        # block until all rpcs finish
```
rohan-varma: Mention that this will also shutdown the RPC layer on this node?

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+        opt = DistributedOptimizer(
+            optim.SGD,
+            model.parameter_rrefs(),
+            lr=0.05,
+        )
+
+        # train for 10 iterations
+        for epoch in range(10):
+            # create distributed autograd context
+            with dist_autograd.context():
```
rohan-varma: Should we provide more details about dist autograd or link to the docs?

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+        nprocs=args.world_size,
+        join=True
+    )
+
+
+In this example, we show how to use RPC as the communication vehicle to pass
+date across workers, and how to use RRef to reference remote objects. It is true
+that you could build the entire structure directly on top of ``ProcessGroup``
+``send`` and ``recv`` APIs or use other communication/RPC libraries. However,
+by using `torch.dstributed.rpc`, you can get the native support plus
```
rohan-varma: nit: spelling

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+
+.. code:: python
+
+    def run_ps():
+        pass
+
+    def run_worker(name, rank, func, world_size):
+        os.environ['MASTER_ADDR'] = 'localhost'
+        os.environ['MASTER_PORT'] = '29500'
+        rpc.init_rpc(name, rank=rank, world_size=world_size)
```
rohan-varma: Should we provide some more details about `rpc.init_rpc`, or link to the docs?

intermediate_source/rpc_tutorial.rst
rohan-varma: Sorry, just noticed the details are in the above docstring

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+        hidden = (
+            torch.randn(nlayers, nindices, nhid),
+            torch.randn(nlayers, nindices, nhid)
+        )
+
+        model = rnn.RNNModel('ps', ntoken, ninp, nhid, nlayers)
+
+        # setup distributed optimizer
+        opt = DistributedOptimizer(
+            optim.SGD,
```
rohan-varma: we have support for other optimizers such as `Adam` right?

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+
+
+
+
+Distributed RNN using Distributed Autograd and Distributed Optimizer
+--------------------------------------------------------------------
+
+In this section, we use an RNN model to show how to build distributed model
+parallel training using the RPC API. The example RNN model is very small and
+easily fit into a single GPU, but developer can apply the similar techniques to
```
rohan-varma: Is the example designed to be run on GPU or CPU? Should we include the ability to run in on GPU by detecting if CUDA is available, and then calling `.to(device)` on the appropriate input tensors and model?

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+of a remote value, and ``local_value`` is only available on the owner of the
+``RRef``.
+
+.. code:: python
+
+    def _call_method(method, rref, *args, **kwargs):
+        return method(rref.local_value(), *args, **kwargs)
+
+
+    def _remote_method(method, rref, *args, async_call=False, **kwargs):
```
rohan-varma: Should we just do `rpc_sync` to keep it simple for tutorial purposes? If we want the possibility to do async in this tutorial maybe we should clarify that a future which can be awaited on will be returned?

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+.. code:: python
+
+    class RNNModel(nn.Module):
+        def __init__(self, ps, ntoken, ninp, nhid, nlayers, dropout=0.5):
+            super(RNNModel, self).__init__()
+
+            # setup embedding table remotely
+            self.emb_table_rref = rpc.remote(ps, EmbeddingTable, args=(ntoken, ninp, dropout))
+            # setup LSTM locally
+            self.rnn = nn.LSTM(ninp, nhid, nlayers, dropout=dropout)
```
rohan-varma: There is a `class RNN` defined above, should we remove that or use it here?

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+        # train for 10 iterations
+        for epoch in range(10):
+            # create distributed autograd context
+            with dist_autograd.context():
+                inp = torch.LongTensor(batch, nindices) % ntoken
+                hidden[0].detach_()
+                hidden[1].detach_()
+                output, hidden = model(inp, hidden)
+                # run distributed backward pass
+                dist_autograd.backward([output.sum()])
```
rohan-varma: Is this intended to be a full training example? Wouldn't we want a loss function in that case?

intermediate_source/rpc_tutorial.rst
rohan-varma: Generally, a pytorch training loop looks like

```
for epoch in n_epochs:
     for batch in get_next_batch(dataset):
         x, y = batch
         out = model(x)
         loss = loss_fun(out, y)
         loss.backward()
         opt.step()
  ```

For the model parallel case, it seems that the dist_autograd context should be created for each batch (so inside the inner for loop)? Would it be incorrect to only create one context per epoch in this case? If so, should we clarify?

File: intermediate_source/rpc_tutorial.rst (hunk 11 above)
Comment (agolynski) [inline intermediate_source/rpc_tutorial.rst]: nit: if you are leaving async_call, maybe do this order of args instead?
_remote_method(method, rref, async_call=False, *args, **kwargs)

File: intermediate_source/rpc_tutorial.rst [hunk 14]
```diff
@@ -0,0 +1,529 @@
...
+                    print('Episode {}\tLast reward: {:.2f}\tAverage reward: {:.2f}'.format(
+                          i_episode, last_reward, agent.running_reward))
+
+                if agent.running_reward > agent.reward_threshold:
+                    print("Solved! Running reward is now {}!".format(agent.running_reward))
+                    break
+        else:
+            # other ranks are the observer
+            rpc.init_rpc(OBSERVER_NAME.format(rank), rank=rank, world_size=world_size)
+            # observers passively waiting for instructions from agents
```
agolynski: nit: from the agent
(as we have a single one)

intermediate_source/rpc_tutorial.rst
mrshenli: Good point! I agree, let me add imports here. Ideally, I hope we could write notebooks instead of rst for tutorials, but I recall that I run into sphinx/multiprocessing/ipython error (see #478). Let me try if that bug is fixed. If yes, I will use a notebook.

intermediate_source/rpc_tutorial.rst
mrshenli: Yep, I really want to add one for that (https://github.com/pytorch/pytorch/issues/31743), but unfortunately, this won't be in 1.4. Let me add a N.B. here.

intermediate_source/rpc_tutorial.rst
mrshenli: `init_rpc` is in `run_worker` which is the entry point for this example. I see how this can be confusing, as I used different ways to spawn RPC processes. let me change this to let two examples using similar entry functions.

intermediate_source/rpc_tutorial.rst
mrshenli: Yes, let me add some for the `CartPole` RL example at least. For RNN as I am using random data as input, any suggestions on what we could show here?

intermediate_source/rpc_tutorial.rst
mrshenli: good point, I was doing one batch per epoch. Let me switch to your suggestions above to stay close to general flow.

intermediate_source/rpc_tutorial.rst
mrshenli: Yes, any local optimizers should work I think. Let me mention that in the docstring

intermediate_source/rpc_tutorial.rst
mrshenli: Oh yes, good catch, let me remove RNN class to make it shorter

intermediate_source/rpc_tutorial.rst
mrshenli: good point, will add

intermediate_source/rpc_tutorial.rst
mrshenli: the agent does use async call on multiple observers. Let me only use async here and explicitly call async and explain futures later when it is used.

mrshenli: I am remove `async_call` arg here and explicitly call `rpc_async` where it is used and explain APIs on Futures there as well

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+            self.decoder.bias.data.zero_()
+            self.decoder.weight.data.uniform_(-0.1, 0.1)
+
+        def forward(self, output):
+            return self.decoder(self.drop(output))
+
+
+With the above three sub-modules, we can now piece them together using RPC to
+create an RNN model. In the code below ``ps`` represents a parameter server,
+which hosts paremeters of the embedding table and the decoder. The constructor
```
agolynski: nit/typo: parameters

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,529 @@
...
+uses the `remote https://pytorch.org/docs/master/rpc.html#torch.distributed.rpc.remote`__
```
agolynski: seems like this link doesn't render correctly for some reason

mrshenli: Yep, I need a `<>` to wrap the URL. Good catch! Let me update.

mrshenli:
Hey @jlin27 @SethHWeidman, may I get some early reviews on this tutorial page? Thanks!

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,555 @@
...
+
+1) In reinforcement learning, it might be relatively expensive to acquire
+   training data from environments while the model itself can be quite small. In
+   this case, it might be useful to spawn multiple observers running in parallel
+   and share a single agent. In this case, the agent takes care of the training
+   locally, but the application would still need libraries to send and receive
+   data between observers and the trainer
+2) Your model might be too large to fit in GPUs on a single machine, and hence
+   would need a library to help split a model onto multiple machines. Or you
+   might be implementing a parameter server training framework, where model
```
rohan-varma: nit: Should we add a link to the parameter server paper for those not familiar? https://www.cs.cmu.edu/~muli/file/parameter_server_osdi14.pdf

mrshenli: yes, will add, thanks!

rohan-varma:
General question: There is a model parallel tutorial that is essentially about sharding the model on the same node across many GPUs: https://deploy-preview-807--pytorch-tutorials-preview.netlify.com/intermediate/model_parallel_tutorial.html. Should we update the title or change it accordingly now that we have the ability to do model parallel across multiple nodes?

intermediate_source/rpc_tutorial.rst
```diff
@@ -0,0 +1,623 @@
...
+contain the recorded action probs and rewards.
+
+
+.. code:: python
+
+    class Agent:
+        ...
+        def run_episode(self, n_steps=0):
+            for ob_rref in self.ob_rrefs:
+                # make async RPC to kick off an episode on all observers
```
rohan-varma: nit: need a  `futs = []`

rohan-varma:
From running the code, I don't think the `args` are defined? We can add something like this:

```
parser = argparse.ArgumentParser(
        description="RPC Reinforcement Learning Example",
        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
    )

parser.add_argument('--world_size', default=2, help='Number of workers')
parser.add_argument('--log_interval', default=1, help='Log every log_interval episodes')
parser.add_argument('--gamma', default=0.1, help='how much to value future rewards')
parser.add_argument('--seed', default=1, help='random seed for reproducibility')
args = parser.parse_args()
```

Also, some variables aren't defined:
```
AGENT_NAME = "agent"
OBSERVER_NAME="obs"
TOTAL_EPISODE_STEP = 100
```

mrshenli:
Thanks @rohan-varma!

> There is a model parallel tutorial that is essentially about sharding the model on the same node across many GPUs

Yes, let me edit the title and also mention the difference in this tutorial.

> From running the code for the RL example, I don't think the args are defined? We can add something like this:

Thanks, those are in the examples, not in the tutorial, let me add them here too. Thank you!

PR #812, opened 2020-01-09
reinforcement_q_learning: resolving torch.uint8 is now deprecated warning

vaibhawvipul:
- torch.uint8 is now deprecated, moving to torch.bool


soumith:
thanks!

PR #832, opened 2020-01-24
Tutorial for A3C in Deep Reinforcement Learning

salayatana66:
Adds a tutorial for Deep Reinforcement Learning. 

The tutorial solves the Cartpole problem using A3C; 
this is achieved using multiprocessing in PyTorch.


salayatana66:
If I do 
`python reinforcement_a3c_tutorial.py` on the branch I am submitting the PR from I get no error, the scripts runs fine; in circleci it complains about an 

"EOF while scanning triple-quoted string literal"

at line 33.

Any advice about what might be causing this?

PR #842, opened 2020-02-05
ReplayMemory to use deque instead of list

PatrikHolopainen:
Changed the ReplayMemory class to use a deque for self.memory instead of a list, as deque has the same functions but also a builti-in maxlen keyword argument, that pretty much does the entire "cyclic-buffer" thing.


holly1238:
Hi @PatrikHolopainen! Can you sign the CLA agreement so that I can merge the pull request? Thanks!

PR #843, opened 2020-02-06
Correct link

jlin27:
Correct link to Quantized Transfer Learning for Computer Vision Tutorial

### DIFF
--- a/beginner_source/transfer_learning_tutorial.py
+++ b/beginner_source/transfer_learning_tutorial.py
@@ -340,6 +340,6 @@ plt.show()
 # -----------------
 #
 # If you would like to learn more about the applications of transfer learning,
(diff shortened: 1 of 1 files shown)


jlin27:
Appears the  CI tests broke because deeplearning.net is down 
```
Feb 07 17:30:53 requests.exceptions.ConnectionError: HTTPConnectionPool(host='deeplearning.net', port=80): Max retries exceeded with url: /data/mnist/mnist.pkl.gz (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f1fac23ad68>: Failed to establish a new connection: [Errno 110] Connection timed out',))
```
Brian suggests that when they come back live, we should consider putting it into our binaries. Looks like this isn't the first time it's happened: https://github.com/Lasagne/Lasagne/issues/46
The issue links to another place that hosts the MNIST package, but I'd rather to grab the original first if that's an option (if/when it comes back up).

PR #849, opened 2020-02-13
add code comments and test code readability++

wAuner:
When I was working through the blitz cifar 10 tutorial it took me quite some time to understand the evaluation logic of the per-class accuracy.
I've added some code comments and rewrote this small part to make it much more obvious what is going on. Generally I've tried to replace the `range`s with fixed numbers for batchsize or classes with speaking names.
In detail:
* now using dicts with classnames as keys to collect classifications, instead of lists which map to class by index
* implicitly removed an unneccessary squeeze operation [diffline271](https://github.com/pytorch/tutorials/compare/master...wAuner:cifartut_improve?expand=1#diff-8b1014a85ac6d7d7450e9c7c4b5a6dfdL271 )
* removed the unexplained reference to the batchsize (hardcoded int)
* make the counting of the correct predictions per class more readable

### DIFF
--- a/beginner_source/blitz/cifar10_tutorial.py
+++ b/beginner_source/blitz/cifar10_tutorial.py
@@ -242,10 +242,13 @@ print('Predicted: ', ' '.join('%5s' % classes[predicted[j]]
 
 correct = 0
 total = 0
+# since we're not training, we don't need to calculate the gradients for our outputs
 with torch.no_grad():
     for data in testloader:
         images, labels = data
+        # calculate outputs by running images through the network 
         outputs = net(images)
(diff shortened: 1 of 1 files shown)


ranman (approved):

holly1238:
Hi @wAuner! Can you sign the CLA agreement so that I can merge the pull request? Thanks!

wAuner:
@holly1238 if it's not some 100 page document written in legal Klingon, then yes 😄

wAuner:
@holly1238 ok I've signed that cla. Maybe the check needs to be triggered again

PR #864, opened 2020-02-24
Remove spawn configuration in AWS tutorial

mrshenli:
The removed lines cause DDP to hang because somehow it triggered
the top layer code to run twice when launching the script from cmd.

Let's see if removing it would pass the CI tests


mrshenli:
Hey @teng-li, is there any specific reason for adding this?

PR #866, opened 2020-02-26
[TorchScript] Fix Intro to TorchScript tutorial

SplitInfinity:
**Summary**
There is a section of the Intro to TorchScript tutorial that is supposed
to illustrate that tracing cannot capture control flow but scripting can.
However, the generated IR for both cases is the same because the tutorial
only prints out the IR for the outer module, not the inner module that
contains the control flow. This commit fixes this issue by printing the IR
for both the inner and outer module to clearly show that tracing does not
capture control flow, but scripting does.

**Testing**
Ran the tutorial file.

Tracing Output:
```
def forward(self,
    argument_1: Tensor) -> None:
  return None

def forward(self,
    input: Tensor,
    h: Tensor) -> Tuple[Tensor, Tensor]:
  _0 = self.dg
  _1 = (self.linear).forward(input, )
  _2 = (_0).forward(_1, )
  _3 = torch.tanh(torch.add(_1, h, alpha=1))
  return (_3, _3)
```

Scripting Output:
```
def forward(self,
    x: Tensor) -> Tensor:
  _0 = bool(torch.gt(torch.sum(x, dtype=None), 0))
  if _0:
    _1 = x
  else:
    _1 = torch.neg(x)
  return _1

def forward(self,
    x: Tensor,
    h: Tensor) -> Tuple[Tensor, Tensor]:
  _0 = (self.dg).forward((self.linear).forward(x, ), )
  new_h = torch.tanh(torch.add(_0, h, alpha=1))
  return (new_h, new_h)
```

### DIFF
--- a/beginner_source/Intro_to_TorchScript_tutorial.py
+++ b/beginner_source/Intro_to_TorchScript_tutorial.py
@@ -274,6 +274,8 @@ class MyCell(torch.nn.Module):
 
 my_cell = MyCell(MyDecisionGate())
 traced_cell = torch.jit.trace(my_cell, (x, h))
+
+print(traced_cell.dg.code)
 print(traced_cell.code)
 
 
@@ -293,8 +295,10 @@ print(traced_cell.code)
 scripted_gate = torch.jit.script(MyDecisionGate())
 
 my_cell = MyCell(scripted_gate)
-traced_cell = torch.jit.script(my_cell)
-print(traced_cell.code)
+scripted_cell = torch.jit.script(my_cell)
+
+print(scripted_gate.code)
+print(scripted_cell.code)
 
 
 ######################################################################


SplitInfinity:
cc: @jamesr66a @suo 

For some reason GitHub won't let me add reviewers

beginner_source/Intro_to_TorchScript_tutorial.py
```diff
@@ -274,6 +274,8 @@ def forward(self, x, h):
 
 my_cell = MyCell(MyDecisionGate())
 traced_cell = torch.jit.trace(my_cell, (x, h))
+
+print(traced_cell.dg.code)
```
indigoviolet, line 278: Suggestion: Add some text here and similarly below, to say which `forward` is for the decision gate and which is for the cell

indigoviolet (approved):
lgtm, one comment inline. Thanks for fixing this!

PR #878, opened 2020-03-08
Replace mention of one-hot encodings with embedding explanation

mariomeissner:
This tutorial mentions that we will encode the words as one-hot encoded vectors, but this does not happen anywhere in the code since we are using embedding layers. I removed the paragraph talking about one-hot encodings and replaced it with a short explanation of embeddings.

Let me know if I should rephrase it or leave out the embedding explanation altogether.


zhangguanheng66 (commented):
It uses the regular vocabulary to convert tokens into indexes and feed the indexes into `Embedding` layer.

mariomeissner:
Yes, indexes, but not in a one-hot-encoded representation.

PR #884, opened 2020-03-12
Update RPC tutorial with new calls to dist autograd and optmizer

rohan-varma:
After recent changes, dist_autograd and dist_optimizer now require context_id to be explicitly passed in. Update that in the tutorial and adjust a few minor comments


mrshenli:
@nairbv shall we land this PR for v1.5?

PR #896, opened 2020-03-18
remove experimental from rpc tutorial

rohan-varma:
Removes experimental tag as we removed this from main RPC docs.


rohan-varma:
@jlin27 Could you please take a look and help land this when you get a chance? Thanks!

PR #921, opened 2020-04-01
C++ autograd tutorial

yf225:
Accompanying example code: https://github.com/pytorch/examples/pull/745.


advanced_source/cpp_autograd.rst
```diff
@@ -0,0 +1,442 @@
...
+  auto x = torch::ones({2, 2}, torch::requires_grad());
+  std::cout << x << std::endl;
+
+Out:
+
+.. code-block:: shell
+
+  1 1
+  1 1
+  [ CPUFloatType{2,2} ]
```
albanD, line 31: The cpp Tensor print does not show requires_grad and the grad_fn like we do in python?

advanced_source/cpp_autograd.rst
```diff
@@ -0,0 +1,442 @@
...
+Out:
+
+.. code-block:: shell
+
+    102.4000
+   1024.0000
+      0.1024
+  [ CPUFloatType{3} ]
+
+You can also stop autograd from tracking history on Tensors with ``.requires_grad() == true``
```
albanD, line 174: I don't think this reads as well as the python version.
I think you can simply state "on Tensors that require gradients".

advanced_source/cpp_autograd.rst
```diff
@@ -0,0 +1,442 @@
...
+  
+  // Calculate loss
+  auto target = torch::randn({3, 3});
+  auto loss = torch::nn::MSELoss()(output, target);
+  
+  // Use norm of gradients as penalty
+  auto grad_output = torch::ones_like(output);
+  auto gradient = torch::autograd::grad({output}, {input}, /*grad_outputs=*/{grad_output},
+                  /*retain_graph=*/true, /*create_graph=*/true,
+                  /*allow_unused=*/true)[0];
```
albanD, line 241: Why do you set allow_unused here?
Also retain_graph is already set to true if not specified and create_graph=true.

advanced_source/cpp_autograd.rst
```diff
@@ -0,0 +1,442 @@
...
+  gradient = gradient.view({-1, 1});
```
albanD, line 242: What is the goal of this view? Especially if you take the norm on dim 1 below.

advanced_source/cpp_autograd.rst
```diff
@@ -0,0 +1,442 @@
...
+and ``torch::autograd::grad``
+(`link <https://pytorch.org/cppdocs/api/function_namespacetorch_1_1autograd_1a1e03c42b14b40c306f9eb947ef842d9c.html>`_)
+for more information on how to use them.
+
+Using custom autograd function in C++
+-------------------------------------
+
+(Adapted from `this tutorial <https://pytorch.org/docs/stable/notes/extending.html#extending-torch-autograd>`_)
+
+Adding operations to ``torch::autograd`` requires implementing a new ``torch::autograd::Function``
```
albanD, line 270: "operations" here might not be the best naming (even though it is in the other tutorial).
Maybe "Adding a new elementary operation to"

advanced_source/cpp_autograd.rst
```diff
@@ -0,0 +1,442 @@
...
+  
+  using namespace torch::autograd;
+  
+  // Inherit from Function
+  class LinearFunction : public Function<LinearFunction> {
+   public:
+    // Note that both forward and backward are static functions
+  
+    // bias is an optional argument
+    static Variable forward(AutogradContext *ctx, Variable input, Variable weight, Variable bias = Variable()) {
```
albanD, line 291: nit: The default value for bias should be a Tensor not a Variable right?

advanced_source/cpp_autograd.rst
```diff
@@ -0,0 +1,442 @@
...
+      ctx->save_for_backward({input, weight, bias});
+      auto output = input.mm(weight.t());
+      if (bias.defined()) {
+        output += bias.unsqueeze(0).expand_as(output);
+      }
+      return output;
+    }
+  
+    static variable_list backward(AutogradContext *ctx, variable_list grad_outputs) {
+      auto saved = ctx->get_saved_variables();
```
albanD, line 301: Why is this not `get_saved_tensors()` ?

advanced_source/cpp_autograd.rst
```diff
@@ -0,0 +1,442 @@
...
+    static Variable forward(AutogradContext *ctx, Variable variable, double constant) {
+      // ctx is a context object that can be used to stash information
+      // for backward computation
+      ctx->saved_data["constant"] = constant;
+      return variable * constant;
+    }
+  
+    static variable_list backward(AutogradContext *ctx, variable_list grad_outputs) {
+      // We return as many input gradients as there were arguments.
+      // Gradients of non-Tensor arguments to forward must be `Variable()`.
```
albanD, line 362: Here again, shouldn't these be Tensors?

advanced_source/cpp_autograd.rst
```diff
@@ -0,0 +1,442 @@
...
+| ``torch.Tensor.is_leaf``       | ``torch::Tensor::is_leaf`` (`link <https://pytorch.org/cppdocs/api/classat_1_1_tensor.html#_CPPv4NK2at6Tensor7is_leafEv>`_)                                            |
++--------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| ``torch.Tensor._version``      | ``torch::Tensor::_version`` (`link <https://pytorch.org/cppdocs/api/classat_1_1_tensor.html#_CPPv4NK2at6Tensor8_versionEv>`_)                                          |
++--------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| ``torch.Tensor._base``         | ``torch::Tensor::_base`` (`link <https://pytorch.org/cppdocs/api/classat_1_1_tensor.html#_CPPv4NK2at6Tensor5_baseEv>`_)                                                |
```
albanD, line 428: Do you really want to mention the internal methods here?
If so, you're missing _is_view().

advanced_source/cpp_autograd.rst
```diff
@@ -0,0 +1,442 @@
...
+
+After translation, most of your Python autograd code should just work in C++.
+If that's not the case, please file a bug report at `GitHub issues <https://github.com/pytorch/pytorch/issues>`_
+and we will fix it as soon as possible.
+
+Conclusion
+----------
+
+You should now have a good overview of PyTorch's C++ autograd API.
+You can find the code examples displayed in this note (yf225 TODO update link) `here
```
albanD, line 439: nit: TODO

advanced_source/cpp_autograd.rst
yf225: No it doesn't :( We might want to fix it though to be consistent with Python tensor print. I filed an issue here: https://github.com/pytorch/pytorch/issues/36004.

advanced_source/cpp_autograd.rst
yf225: Yeah I think this should be `get_saved_tensors()`, and my guess is that this API was created before the Variable/Tensor merge was completed, and the author wanted to make a distinction between tensors and variables. Now that we don't have this distinction anymore, I think we should change it to match the Python API: https://github.com/pytorch/pytorch/issues/36099.

yf225: I renamed all `Variable`s to `Tensor` now, to minimize the inconsistency with Python to only this `get_saved_variables()` function :)

albanD: Yes I guess it is just a remain of the cpp Tensor/Variable merge.

yf225:
Note to self: merge https://github.com/pytorch/examples/pull/745 first and update link in this PR

PR #926, opened 2020-04-02
Blitz / Autograd Tutorial : Add word to distinguish math formulas.

sbrodehl:
In some browsers (Arch Linux + Chromium) math formulas are shown very small, so it was very hard to see the comma between the two adjacent formulas and thus, reading them.

For better reading comprehension I added the word _where_ at the relevant position.

### DIFF
--- a/beginner_source/blitz/autograd_tutorial.py
+++ b/beginner_source/blitz/autograd_tutorial.py
@@ -101,7 +101,7 @@ print(x.grad)
 ###############################################################
 # You should have got a matrix of ``4.5``. Let’s call the ``out``
 # *Tensor* “:math:`o`”.
-# We have that :math:`o = \frac{1}{4}\sum_i z_i`,
+# We have that :math:`o = \frac{1}{4}\sum_i z_i`, where
 # :math:`z_i = 3(x_i+2)^2` and :math:`z_i\bigr\rvert_{x_i=1} = 27`.
 # Therefore,
 # :math:`\frac{\partial o}{\partial x_i} = \frac{3}{2}(x_i+2)`, hence


sbrodehl:
Why is netlify unable to build this PR?

sbrodehl:
The underlying problem seems to be related to
- [pytorch/pytorch_sphinx_theme/PR/35](https://github.com/pytorch/pytorch_sphinx_theme/pull/35)
- which itself is related to [pytorch/pytorch#15202](https://github.com/pytorch/pytorch/issues/15202)

I suspect that some CSS on the page is maybe affecting the measurements that MathJax is making in order to determine the font sizes.

I use `Chromium Version 80.0.3987.163 (Official Build) Arch Linux (64-bit)`.

sbrodehl:
It looks like this still is an issue.
autograd_tutorial html_optional-reading-vector-calculus-using-autograd

Multiple issues and pull request tried to solve this, but no improvement can be seen.

Since the generated API docs look fine, e.g. [torch.nn.Conv2d](https://pytorch.org/docs/1.7.1/generated/torch.nn.Conv2d.html#torch.nn.Conv2d), maybe MathJax is not a good choice, but KaTex is?

bigfootjon:
Hi @sbrodehl, this was mistakenly closed due to a mixup with the default branch name. Unfortunately, I can't reopen the PR because the old master branch was deleted. Please create a new PR and reference this PR.

sbrodehl:
Thanks for the quick reply!
I have moved forward and replaced MathJax with KaTeX in a new PR #1362 , it is now the same as in the [PyTorch Docs](https://pytorch.org/docs/master/torch.html).

PR #934, opened 2020-04-08
Zeroing out gradients recipe

CamiWilliams:
Intro-level recipe for zeroing out gradients in PyTorch. Showcases CIFAR10 to train on.


CamiWilliams:
Updated py according to script.

PR #936, opened 2020-04-08
Saving and loading models for inference recipe

CamiWilliams:
Intro-level recipe for saving and loading models for inference in PyTorch. This is 2 of 6 for saving/loading models.


CamiWilliams:
Updated according to script

PR #938, opened 2020-04-08
Saving and loading a general checkpoint recipe

CamiWilliams:
Intro-level recipe for saving and loading a general checkpoint for inference or training in PyTorch. This is 3 of 6 for saving/loading models.


CamiWilliams:
Updated according to script

PR #943, opened 2020-04-09
Warmstarting a model using parameters from a different model recipe

CamiWilliams:
Intro-level recipe for warmstarting models in PyTorch. This is 5 of 6 for saving/loading models.


CamiWilliams:
Updated with script

PR #946, opened 2020-04-11
Update links in RPC tutorial

mrshenli:


mrshenli:
hey @jlin27 @rohan-varma, I noticed that the url has changed in master compared to stable. 

master: master/rpc/rpc.html
stable: stable/rpc.html

Does this mean we should update all links in tutorial?

rohan-varma:
I think that the difference is `master` will always show the latest docs (i.e. the ones that are landed to master), stable will show the ones corresponding to the current release. Given that we want to release Dist RPC framework as stable, I think we should do this update. cc @jlin27

abitrolly:
Looks like this got stalled because of branch conflict.

mrshenli:
My bad, will create a new one to fix links and also add links to DDP + RPC and autograd profiler.

PR #963, opened 2020-04-22
make tensor contiguous

vincentqb:
Make tensor contiguous, see pytorch/audio#571.


brianjo:
Thanks for the fix @vincentqb. I added to my PR so I'll close this as it's now in master.

PR #973, opened 2020-04-28
Update DDP Tutorial to remove Single-Process Multi-Device Use Case

mrshenli:
As we do not recommend using Single-Process Multi-Device mode with
DDP due to the reduced throughput and as this mode is broken in
v1.5, added this PR to update DDP tutorial to using Multi-Process
Single-Device mode.

This PR also includes the following minor fixes:

1. highlight that this tutorial requires at least 8 GPUs to run
2. added a pointer to RPC
3. added a pointer to torchelastic
4. as DDP now broadcast initial param values from rank 0 to others
   it no longer require setting RNG seed.

### DIFF
--- a/intermediate_source/ddp_tutorial.rst
+++ b/intermediate_source/ddp_tutorial.rst
@@ -2,22 +2,31 @@ Getting Started with Distributed Data Parallel
 =================================================
 **Author**: `Shen Li <https://mrshenli.github.io/>`_
 
-`DistributedDataParallel <https://pytorch.org/docs/stable/_modules/torch/nn/parallel/distributed.html>`__
-(DDP) implements data parallelism at the module level. It uses communication
-collectives in the `torch.distributed <https://pytorch.org/tutorials/intermediate/dist_tuto.html>`__
-package to synchronize gradients, parameters, and buffers. Parallelism is
-available both within a process and across processes. Within a process, DDP
-replicates the input module to devices specified in ``device_ids``, scatters
-inputs along the batch dimension accordingly, and gathers outputs to the
-``output_device``, which is similar to
-`DataParallel <https://pytorch.org/tutorials/beginner/blitz/data_parallel_tutorial.html>`__.
-Across processes, DDP inserts necessary parameter synchronizations in forward
-passes and gradient synchronizations in backward passes. It is up to users to
-map processes to available resources, as long as processes do not share GPU
-devices. The recommended (usually fastest) approach is to create a process for
-every module replica, i.e., no module replication within a process. The code in
-this tutorial runs on an 8-GPU server, but it can be easily generalized to
-other environments.
+`DistributedDataParallel <https://pytorch.org/docs/stable/nn.html#torch.nn.parallel.DistributedDataParallel>`__
+(DDP) implements data parallelism at the module level which can run across
+multiple machines. Applications using DDP should spawn multiple processes and
+create a single DDP instance per process. DDP uses collective communications in the
+`torch.distributed <https://pytorch.org/tutorials/intermediate/dist_tuto.html>`__
+package to synchronize gradients and buffers. More specifically, DDP registers
+an autograd hook for each parameter given by ``model.parameters()`` and the
+hook will fire when the corresponding gradient is computed in the backward
+pass. Then DDP uses that signal to trigger gradient synchronization across
+processes. Please refer to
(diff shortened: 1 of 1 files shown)


mrshenli:
The corresponding example PR is https://github.com/pytorch/examples/pull/763

mrshenli:
Hey @kiukchung this currently links to https://github.com/pytorch/elastic. Let me know if you prefer https://pytorch.org/elastic/0.2.0rc1/index.html or a different link. Thanks!

mrshenli:
cc @jlin27 @gchanan @zhaojuanmao @rohan-varma

intermediate_source/ddp_tutorial.rst
```diff
@@ -2,46 +2,53 @@ Getting Started with Distributed Data Parallel
...
-passes and gradient synchronizations in backward passes. It is up to users to
-map processes to available resources, as long as processes do not share GPU
-devices. The recommended (usually fastest) approach is to create a process for
-every module replica, i.e., no module replication within a process. The code in
-this tutorial runs on an 8-GPU server, but it can be easily generalized to
-other environments.
+`DistributedDataParallel <https://pytorch.org/docs/stable/nn.html#torch.nn.parallel.DistributedDataParallel>`__
+(DDP) implements data parallelism at the module level which can run across
+multiple machines. Applications using DDP should spawn multiple processes and
+create a DDP instance in each process. DDP uses collective communications in the
```
rohan-varma, line 8: nit: "create a single DDP instance per process" might be clearer?

intermediate_source/ddp_tutorial.rst
```diff
@@ -2,46 +2,53 @@ Getting Started with Distributed Data Parallel
...
+`torch.distributed <https://pytorch.org/tutorials/intermediate/dist_tuto.html>`__
+package to synchronize gradients and buffers. More specifically, DDP inserts
+an autograd hook for each model parameter which will fire when the
```
rohan-varma, line 11: Is the autograd hook indeed per-parameter as given by `model.parameters()`? If user takes a model and then manually adds a `nn.Parameter()` to it, will DDP sync gradients from this parameter as well? If so, might be worth mentioning this (maybe in the docs is better though)

intermediate_source/ddp_tutorial.rst
```diff
@@ -127,23 +125,27 @@ Otherwise, global gradient synchronizes will not make sense.
...
-In DDP, constructor, forward method, and differentiation of the outputs are
-distributed synchronization points. Different processes are expected to reach
-synchronization points in the same order and enter each synchronization point
-at roughly the same time. Otherwise, fast processes might arrive early and
-timeout on waiting for stragglers. Hence, users are responsible for balancing
-workloads distributions across processes. Sometimes, skewed processing speeds
-are inevitable due to, e.g., network delays, resource contentions,
-unpredictable workload spikes. To avoid timeouts in these situations, make
-sure that you pass a sufficiently large ``timeout`` value when calling
+In DDP, the constructor and the backward pass are distributed synchronization
```
rohan-varma, line 139: Noticed we removed forward method, was it inaccurate that it was a point of synchronization or did something change?

kiukchung: > kiuk

@mrshenli that would be awesome, the correct link is https://pytorch.org/elastic (it redirects to the lastest release)

intermediate_source/ddp_tutorial.rst
mrshenli: If user manually adds a `nn.Parameter()` the `Model.__setattr__` should detect that and call `register_parameter` for it, so this manually added parameter would also be included ini `model.parameters()`. Let me modify this to "each model parameter in `model.parameters()`".

intermediate_source/ddp_tutorial.rst
mrshenli: > Noticed we removed forward method, was it inaccurate that it was a point of synchronization or did something change?

There is sync in forward to broadcast buffers from rank 0 to all other ranks. I removed this  as 1) there is no intra-node sync in MPSD mode 2) buffer is not used in this tutorial. But give it a second thought, I think it might be better to keep this to stay consistent with the previous version which is correct for general cases. Thanks for spotting this!

mrshenli: links are updated, thx!

PR #976, opened 2020-05-03
Use python3 super syntax

ikamensh:
minor improvement - only python3 is supported by Torch 1.5, so we can use newer, shorter syntax.

https://www.python.org/dev/peps/pep-3135/
https://stackoverflow.com/questions/576169/understanding-python-super-with-init-methods


holly1238:
Hi @ikamensh! Can you sign the CLA agreement so that I can merge the pull request? Thanks!

ikamensh:
done

PR #979, opened 2020-05-05
Update torchvision_tutorial.rst

EMCP:
this is to reflect that, some users are seeing an error and you must put the integer into a string.. in order for the backbone neural network to be accepted during the tutorial.

See : https://github.com/pytorch/vision/issues/2021

I left the sourcecode as-is.. so tutorial members can get a little stuck in the event they need to modify this variable to a string.. and now the hint is in the comments


holly1238:
Closing this PR. This issue was fixed in https://github.com/pytorch/tutorials/pull/1085 (duplicate PR).

PR #989, opened 2020-05-11
[WIP][CheatSheet] update out of place ops, update act fn, optimzers, schedulers, samplers

ptrblck:
Added some update to out-of-place tensor operations as well as some updates to
* activation functions
* optimizers
* schedulers
* samplers

CC @jspisak 

Let me know, if these changes look OK and I can add them to the `PyTorch Cheat.md`.
Tagging this PR as WIP for now.

### DIFF
--- a/beginner_source/ptcheat.rst
+++ b/beginner_source/ptcheat.rst
@@ -80,8 +80,8 @@ Distributed Training
 
 .. code-block:: python
 
-    import torch.distributed as dist          # distributed communication
-    from multiprocessing import Process       # memory sharing processes
+    import torch.distributed as dist             # distributed communication
+    from torch.multiprocessing import Process    # memory sharing processes
 
 See `distributed <https://pytorch.org/docs/stable/distributed.html>`__
 and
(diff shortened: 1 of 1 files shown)


ptrblck:
Could it be failing, because I would need to add the same changes to this file: https://github.com/pytorch/tutorials/blob/master/beginner_source/PyTorch%20Cheat.md ?

ptrblck:
If the changes look OK, I'll add the same diffs to the other mentioned file later tonight.

jspisak:
yes, we should actually be updating the .md file which is then renders the .rts. Thanks!

jlin27:
Let me look into the Netlify builds. I cleared the cache and rebuild which usually clears it up. 

Thanks @ptrblck for these updates!

ptrblck:
@jlin27 Thanks for looking into it!
I'm not sure why there are two files, but I saw that both seem to have commits.
I've just updated one file and can update the other one in a follow-up PR, if necessary.

PR #1006, opened 2020-05-29
Fixing a bug in Captum Recipe

mandrulevicius:
In the Pytorch Recipe, MODEL INTERPRETABILITY USING CAPTUM there is a following bug:

In the following line of code, the comma after the brackets is interpreted as an additional array dimension, causing an assert error when passed to the viz.visualize_image_attr_multiple() function.

vis_signs = ["all", "all"],

Removing the comma solves the issue.


holly1238:
Closing this PR. This issue was fixed and merged in https://github.com/pytorch/tutorials/pull/1271/.

PR #1014, opened 2020-06-05
Apply PEP 8 formatting to `transformer_tutorial.py` imports

jgerityneurala:
This PR brings all imports to the top of the surrounding code, according to [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports).

The change that brought me here was

```python
class TransformerModel(nn.Module):
    def __init__(self, ntoken, ninp, nhead, nhid, nlayers, dropout=0.5):
        super(TransformerModel, self).__init__()
        from torch.nn import TransformerEncoder, TransformerEncoderLayer
        # ...
```

An import inside of `__init__()` like this is extremely uncommon, and it does not look like it's necessary to have it here in this case. If there is a reason the change shouldn't be made, there should explanatory text of why it is an exception to _"imports are always at the top of the file"_ as PEP 8 has it. I know this project isn't explicitly PEP 8, but this doesn't vary too much across style conventions IME.


holly1238:
Hi @jgerityneurala! Can you sign the CLA agreement so that I can merge the pull request? Thanks!

jgerityneurala:
> Hi @jgerityneurala! Can you sign the CLA agreement so that I can merge the pull request? Thanks!

Whoops, I didn't realize there was one when I submitted this PR. Just signed it, thanks.

jgerityneurala:
CI failure is due to exceeding the timeout, shouldn't be caused by this PR's content

PR #1021, opened 2020-06-11
Add TorchScript fork/join tutorial

jamesr66a:


advanced_source/torch_script_parallelism.rst
```diff
@@ -0,0 +1,276 @@
...
+in TorchScript. This parallelism has the following properties:
+
+* dynamic - The number of parallel tasks created and their workload can depend on the control flow of the program.
+* inter-op - The parallelism is concerned with running TorchScript program fragments in parallel. This is distinct from *intra-op parallelism*, which is concerned with splitting up individual operators and running subsets of the operator's work in parallel.
+Basic Syntax
+------------
+
+The two important APIs for dynamic parallelism are:
+
+* ``torch.jit.fork(fn : Callable, *args, **kwargs) -> torch.jit.Future[T]``
```
eellison, line 14: if we want, syntax for a function that returns `T` is `Callable[..., T]`

advanced_source/torch_script_parallelism.rst
```diff
@@ -0,0 +1,276 @@
...
+        return torch.neg(x)
+
+    @torch.jit.script
+    def example(x):
+        # Call `foo` normally
+        x_normal = foo(x)
+
+        # Call `foo` using parallelism:
+        # First, we "fork" off a task. This task will run `foo` with argument `x`
+        future = torch.jit.fork(foo, x)
```
eellison, line 33: maybe makes sense to put the fork call before the normal call to emphasize it executes asynchronously

advanced_source/torch_script_parallelism.rst
```diff
@@ -0,0 +1,276 @@
...
+
+    import torch
+    from typing import List
+
+    def foo(x):
+        return torch.neg(x)
+
+    @torch.jit.script
+    def example(x):
+        futures = torch.jit.annotate(List[torch.jit.Future[torch.Tensor]], [])
```
eellison, line 66: I would use python 3 syntax here, as we've deprecated python 2

advanced_source/torch_script_parallelism.rst
```diff
@@ -0,0 +1,276 @@
...
+        for _ in range(100):
+            futures.append(torch.jit.fork(foo, x))
+
+        results = []
+        for future in futures:
+            results.append(torch.jit.wait(future))
+
+        return torch.sum(torch.stack(results))
+
+    print(example(torch.ones([])))
```
eellison, line 76: maybe put a non-empty tensor input

advanced_source/torch_script_parallelism.rst
```diff
@@ -0,0 +1,276 @@
...
+
+.. note::
+
+    When we initialized an empty list of Futures, we needed to add an explicit
+    type annotation to the value via ``torch.jit.annotate``. This is because
```
eellison, line 81: similarly, maybe remove torch.jit.annotate here. We should probably have an example using list comprehensions, bc  it's easier without the type annotation. `futures = [torch.jit.fork(foo, x) for _ in range(100)]`.  

I think this whole section could be a lot shorter, something like. `Empty containers default to assuming Tensor values, so we annotate the list constructor as being of type `List[torch.jit.Future[Tensor]'`. And could maybe in inline in the example.

advanced_source/torch_script_parallelism.rst
```diff
@@ -0,0 +1,276 @@
...
+
+        def forward(self, x : torch.Tensor) -> torch.Tensor:
+            results = []
+            for model in self.models:
+                results.append(model(x))
+            return torch.stack(results).sum(dim=0)
+
+    # For a head-to-head comparison to what we're going to do with fork/wait, let's
+    # instantiate the model and compile it with TorchScript
+    ens = torch.jit.script(LSTMEnsemble(n_models=4))
```
eellison, line 146: what is `ens` supposed to be ?

advanced_source/torch_script_parallelism.rst
```diff
@@ -0,0 +1,276 @@
...
+                futures.append(torch.jit.fork(model, x))
+
+            # Collect the results from the launched tasks
+            results = []
+            for future in futures:
+                results.append(torch.jit.wait(future))
+
+            return torch.stack(results).sum(dim=0)
+
+Or, if you value brevity, we can use list comprehensions:
```
eellison, line 246: nice

advanced_source/torch_script_parallelism.rst
```diff
@@ -0,0 +1,276 @@
...
+You may have noticed that there is a further parallelization opportunity in our
+code: we can also run the models contained in ``LSTMEnsemble`` in parallel with
+each other. The way to do that is simple enough, this is how we should change
+the ``forward`` method of ``LSTMEnsemble``:
+
+.. code-block:: python
+
+        def forward(self, x : torch.Tensor) -> torch.Tensor:
+            # Launch tasks for each model
+            futures = torch.jit.annotate(List[torch.jit.Future[torch.Tensor]], [])
```
eellison, line 235: python 3 annotation

advanced_source/torch_script_parallelism.rst
```diff
@@ -0,0 +1,276 @@
...
+            for model in self.models:
+                futures.append(torch.jit.fork(model, x))
+
+            # Collect the results from the launched tasks
+            results = []
```
eellison, line 240: maybe worth putting the annotation here anyway, but could leave it.

advanced_source/torch_script_parallelism.rst
```diff
@@ -0,0 +1,276 @@
...
+while it continues to execute ``cell_b``. This causes the execution of both the
+cells to be overlapped with each other.
+
+Running the script again with this simple modification yields a runtime of
+``1.71`` seconds for an improvement of ``17%``!
+
+Aside: Visualizing Parallelism
+------------------------------
+
+We're not done optimizing our model but it's worth introducing the tooling we
```
eellison, line 201: I would move this to after the section wherer you improve parallelism further.

advanced_source/torch_script_parallelism.rst
jamesr66a: "ensemble"

advanced_source/torch_script_parallelism.rst
jamesr66a: I wanted to put this here because it allows us to see a delta of what happens when we add ``fork`` and ``wait``, rather than just seeing one example

advanced_source/torch_script_parallelism.rst
jamesr66a: this is passing in a scalar `1.0` tensor

PR #1031, opened 2020-06-16
[mobile] Mobile Perf Recipe

IvanKobzarev:


index.rst
```diff
@@ -288,6 +288,13 @@ Welcome to PyTorch Tutorials
    :link: advanced/static_quantization_tutorial.html
    :tags: Image/Video,Quantization,Model-Optimization
 
+.. customcarditem::
```
jlin27, line 291: If this will be on the recipes page, we'll need to move this over to recipes_index.rst instead of index.rst which is for the longer end-to-end tutorials: https://github.com/pytorch/tutorials/blob/master/recipes_source/recipes_index.rst

index.rst
```diff
@@ -288,6 +288,13 @@ Welcome to PyTorch Tutorials
...
+   :header: PyTorch Mobile Performance Recipes
+   :card_description: List of recipes for performance optimizations for using PyTorch on Mobile.
+   :image: _static/img/thumbnails/cropped/experimental-Dynamic-Quantization-on-an-LSTM-Word-Language-Model.png
+   :link: recipes/recipes/mobile_perf.html
```
jlin27, line 295: I think the CI is failing because we need add the ellipses because the recipes are nested one level deeper -  `.. /recipes/recipes/mobile_perf.html`

recipes_source/mobile_perf.rst
```diff
@@ -0,0 +1,191 @@
...
+::
+
+  from torch.utils.mobile_optimizer import optimize_for_mobile
+  traced_model = torch.jit.load("input_model_path")
+  optimized_model = optimize_for_mobile(traced_model)
+  torch.jit.save(optimized_model, "output_model_path")
+
+2. Fuse operators using ``torch.quantization.fuse_modules``
+Do not be confused that fuse_modules is in the quantization package.
+It works for all types of torch script modules.
```
fbbradheintz, line 36: Two things:
1) PMM is going to come back and say that we should write it TorchScript.
2) The code below *does not* pass a TorchScript module to `fuse_modules()` - that MobileNet v2 from TorchVision is not a subclass of `ScriptModule`.

Passing in the original torchvision module itself or a version of it processed by `torch.jit.script()` works, with the latter giving ~2% file size improvement.

recipes_source/mobile_perf.rst
```diff
@@ -0,0 +1,191 @@
...
+              torch.quantization.fuse_modules(m, ['0', '1'], inplace=True)
+          if name == 'InvertedResidual':
+              for idx in range(len(m.conv)):
+                  if type(m.conv[idx]) == torch.nn.Conv2d:
+                      torch.quantization.fuse_modules(m.conv, [str(idx), str(idx + 1)], inplace=True)
+
+  m = torchvision.models.mobilenet_v2(pretrained=True)
+  m.eval()
+  fuse_model(m)
+  torch.jit.trace(m, torch.rand(1, 3, 224, 224)).save("mobilenetV2-bnfused.pt")
```
fbbradheintz, line 67: Should we be guiding people to `torch.jit.trace()`? `torch.jit.script()` preserves control flow, `trace()` does not. `trace()` is still there for cases where `script()` hits an unsupported op.

recipes_source/mobile_perf.rst
```diff
@@ -0,0 +1,191 @@
...
+  import torchvision
+
+  supported_qengines = torch.backends.quantized.supported_engines
+  print(supported_qengines)
+  model = torchvision.models.quantization.__dict__['mobilenet_v2'](pretrained=True, quantize=True)
+  torch.backends.quantized.engine='qnnpack'
+  model.eval()
+  script_model = torch.jit.script(model)
+  x = torch.rand(1, 3 , 224, 224)
+  y = script_model(x)
```
fbbradheintz, line 96: `y` is never used, which means we could do without `x` as well.

recipes_source/mobile_perf.rst
```diff
@@ -0,0 +1,191 @@
...
+Both of those resources can be crucial for the performance of mobile applications.
+
+::
+
+  import torch
+  import torchvision
+
+  supported_qengines = torch.backends.quantized.supported_engines
+  print(supported_qengines)
+  model = torchvision.models.quantization.__dict__['mobilenet_v2'](pretrained=True, quantize=True)
```
fbbradheintz, line 91: Two things:

1) Is this the quantization workflow we want to show? This looks like it's just pulling down a pre-quantized version of the model, in which case this recipe is only useful if you want to quantize the torchvision models. In the general case, I'd think people would want to be able to quantize *their own* trained models for mobile deployment.
2) On MacOS, this line throws a warning:

```
/Users/bradheintz/anaconda2/envs/pyto16pre/lib/python3.8/site-packages/torch/nn/quantized/modules/utils.py:8: UserWarning: 0quantize_tensor_per_tensor_affine current rounding mode is not set to round-to-nearest-ties-to-even (FE_TONEAREST). This will cause accuracy issues in quantized models. (Triggered internally at  ../aten/src/ATen/native/quantized/affine_quantizer.cpp:25.)
  qweight = torch.quantize_per_tensor(
/Users/bradheintz/anaconda2/envs/pyto16pre/lib/python3.8/site-packages/torch/quantization/observer.py:134: UserWarning: must run observer before calling calculate_qparams.                                    Returning default scale and zero point
  warnings.warn(
```

(And yes, it really looks like that in my terminal.)

recipes_source/mobile_perf.rst
```diff
@@ -0,0 +1,191 @@
...
+  torch.backends.quantized.engine='qnnpack'
+  model.eval()
+  script_model = torch.jit.script(model)
```
fbbradheintz, line 94: I don't actually know the answer to this: Is it preferable to do quantization before or after TorchScript conversion, or does it matter at all?

recipes_source/mobile_perf.rst
raghuramank100: We should show a workflow where we start with a floating point model and then do the quantization. The steps are:
# Start with a fully trained floating point model
# The model code is modified to enable eager mode quantization, for more details
# please see the quantization tutorials at: https://pytorch.org/tutorials/advanced/static_quantization_tutorial.html

model = torchvision.models.quantization.__dict__['resnet18'](pretrained=True, quantize=False)
torch.backends.quantized.engine='qnnpack'
# We convert the float model with the appropriate_Qconfig
model.eval()
model.qconfig = torch.quantization.get_default_qconfig('qnnpack')
torch.quantization.prepare(model)
# Run model with representative data for calibration
# model(calibration_data)
torch.quantization.convert(model)
script_model = torch.jit.script(model)

# Export to mobile
script_model._save_for_lite_interpreter("model.bc")

recipes_source/mobile_perf.rst
raghuramank100: Currently, pytorch only supports doing quantization prior to scripting. We are working on adding support for quantization after scripting, but it is not part of release 1.6 yet.

recipes_source/mobile_perf.rst
```diff
@@ -0,0 +1,223 @@
...
+Quantization of the model not only moves computation to int8,
+but also reduces the size of your model on a disk.
+That size reduction helps to reduce disk read operations during the first load of the model and decreases the amount of RAM.
+Both of those resources can be crucial for the performance of mobile applications.
+This code does quantization, using stub for model calibration function, you can find more about it `here <https://pytorch.org/tutorials/advanced/static_quantization_tutorial.html#post-training-static-quantization>`__.
+
+::
+
+  model.qconfig = torch.quantization.get_default_qconfig('qnnpack')
+  torch.quantization.prepare(model)
```
fbbradheintz, line 101: `prepare()` has `inplace=False` by default, and the same goes for `convert()`, so this whole method is a no-op except for setting `model.qconfig`.

We either need to do:
```
model = torch.quantization.prepare(model)
# calibration
return torch.quantization.convert(model)
```

or:
```
torch.quantization.prepare(model, inplace=True)
# calibration
torch.quantization.convert(model, inplace=True)
```

PR #1039, opened 2020-06-24
Fix allreduce implmentation in tutorial

Lyken17:
As discussed in https://github.com/pytorch/pytorch/issues/31572, the allreduce implementation is not correct (confirmed by mutiple persons). This PR is proposed to fix the bug.


rohan-varma:
LGTM, I verified with a local repro that the allreduce now works as expected. @jlin27 Can we get this approved and landed? Thanks!

PR #1049, opened 2020-07-03
Add windows ci build

guyang3532:


.circleci/config.yml
```diff
@@ -538,3 +583,19 @@ workflows:
...
+      - pytorch_windows_build_worker_3:
+          name: win_test_worker_3
+      - pytorch_windows_build_worker_4:
+          name: win_test_worker_4
+      - pytorch_windows_build_worker_5:
+          name: win_test_worker_5
+      - pytorch_windows_build_worker_6:
+          name: win_test_worker_6
+      - pytorch_windows_build_worker_7:
+          name: win_test_worker_7
```
seemethere, line 601: This job seems extremely expensive to run basically the way the math works this would be `500 credits x 55 minutes * 8 workers`

We should at the very least gate these to run only on master and even, I think, gate these to only run when someone manually approves these to run.

cc @malfet, @yns88

seemethere, line 601: Here's what a manual approval would look like:

```suggestion
      - hold:
          type: approval
      - pytorch_windows_build_worker_0:
          name: win_test_worker_0
          requires:
            - hold
      - pytorch_windows_build_worker_1:
          name: win_test_worker_1
          requires:
            - hold
      - pytorch_windows_build_worker_2:
          name: win_test_worker_2
          requires:
            - hold
      - pytorch_windows_build_worker_3:
          name: win_test_worker_3
          requires:
            - hold
      - pytorch_windows_build_worker_4:
          name: win_test_worker_4
          requires:
            - hold
      - pytorch_windows_build_worker_5:
          name: win_test_worker_5
          requires:
            - hold
      - pytorch_windows_build_worker_6:
          name: win_test_worker_6
          requires:
            - hold
      - pytorch_windows_build_worker_7:
          name: win_test_worker_7
          requires:
            - hold
```

Makefile
```diff
@@ -45,18 +45,18 @@ download:
 	mkdir -p intermediate_source/data
 
 	# transfer learning tutorial data
-	wget -N https://download.pytorch.org/tutorial/hymenoptera_data.zip -P $(DATADIR)
+	wget -N http://download.pytorch.org/tutorial/hymenoptera_data.zip -P $(DATADIR)
```
seemethere, line 48: Why the switch from `https` -> `http`?

I'd much prefer it to use `https` over regular `http`?

.circleci/config.yml
yns88: Does every worker need a GPU machine to run? Can some of the work be shifted to CPU-only machines? Can the amount of coverage on the Windows build be reduced so it doesn't need 8 workers?

If not, then yes, we'll need to limit this to master-only (that might be sufficient, since this repository isn't nearly as active as pytorch/pytorch).

guyang3532: Considering the parallel jobs (8 jobs with most 55minutes for one) seems not much more efficient than one job (1 hour 36 minutes). I have changed the parallel jobs to a single one with manually approval.

guyang3532:
The CI failure seems not related to this change.

Makefile
guyang3532: > Why the switch from `https` -> `http`?
> 
> I'd much prefer it to use `https` over regular `http`?

fixed

seemethere:
I've re-kicked the failed jobs since they appear to be timeouts, will merge on green

brianjo:
The [CI](https://app.circleci.com/pipelines/github/pytorch/tutorials?branch=master) for this PR failed. I've kicked the build on the win thread that failed.

PR #1083, opened 2020-07-24
Update ptcheat.rst

edwardleardi:
scheduler.step() should not be called at the start of every epoch as of [PyTorch 1.1.0](https://pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate). Instead it should be called after the optimizer has updated the weights (after `optimizer.step()` is called)


holly1238:
Hi @edwardleardi! Can you sign the CLA agreement so that I can merge the pull request? Thanks!

PR #1092, opened 2020-07-29
fix bugs for data_loading_tutorial and dcgan_faces_tutorial

guyang3532:
bug in data_loading_tutorial:
      This tutorial fails on windows because code with multiple processes must be put in main() function on windows platform. And to not change the structure of the tutorial html page,  just modify the multiple processes to single process.

bug in dcgan_faces_tutorial:
       When pytorch updating from v1.5 to v1.6, one warning message "Providing a bool or integral fill value without setting the optional `dtype` or `out` arguments is currently unsupported." is changed to error. So change the tutorial correspondingly.


guyang3532:
The error seems not related.

guyang3532:
@seemethere

PR #1093, opened 2020-07-30
G to be in eval() mode for G's output on fixed_noise

snehansu-iisc:
For G's output on fixed_noise, netG should be in eval() mode for Batch_Norm layers to use running mean & running variance instead of batch_statistics.
netG's eval() mode makes sure that that G's output will also be valid on a single instance of fixed_noise e.g. fixed_noise = torch.randn(1, nz, 1, 1, device=device).
After G's output has been obtained, we reset netG to train() mode to allow G's training further.


holly1238:
closing this PR (old PR from an account that doesn't exist anymore)

Issue #1101, opened 2020-08-03
numeric_suite_tutorial and graph_mode_static_quantization_tutorial fail on CI test


guyang3532:
The fail job link:
https://app.circleci.com/pipelines/github/pytorch/tutorials/2260/workflows/28fe0633-2d0d-49e8-aa42-b524ca2bfaac/jobs/39040

Traceback:
C:\Users\circleci\project\prototype_source\numeric_suite_tutorial.py failed leaving traceback:
Traceback (most recent call last):
  File "C:\tools\miniconda3\envs\testenv\lib\site-packages\sphinx_gallery\gen_rst.py", line 394, in _memory_usage
    out = func()
  File "C:\tools\miniconda3\envs\testenv\lib\site-packages\sphinx_gallery\gen_rst.py", line 382, in __call__
    exec(self.code, self.globals)
  File "C:\Users\circleci\project\prototype_source\numeric_suite_tutorial.py", line 86, in <module>
    print(key, compute_error(wt_compare_dict[key]['float'], wt_compare_dict[key]['quantized'].dequantize()))
  File "C:\Users\circleci\project\intermediate_source\memory_format_tutorial.py", line 260, in check_cl
    'Operator `{}` lost channels_last property'.format(name))
Exception: Operator `dequantize` lost channels_last property

C:\Users\circleci\project\prototype_source\graph_mode_static_quantization_tutorial.py failed leaving traceback:
Traceback (most recent call last):
  File "C:\tools\miniconda3\envs\testenv\lib\site-packages\sphinx_gallery\gen_rst.py", line 394, in _memory_usage
    out = func()
  File "C:\tools\miniconda3\envs\testenv\lib\site-packages\sphinx_gallery\gen_rst.py", line 382, in __call__
    exec(self.code, self.globals)
  File "C:\Users\circleci\project\prototype_source\graph_mode_static_quantization_tutorial.py", line 234, in <module>
    ts_model = torch.jit.script(float_model).eval() # ts_model = torch.jit.trace(float_model, input)
  File "C:\tools\miniconda3\envs\testenv\lib\site-packages\torch\jit\__init__.py", line 1516, in script
    return torch.jit._recursive.create_script_module(obj, torch.jit._recursive.infer_methods_to_compile)
  File "C:\tools\miniconda3\envs\testenv\lib\site-packages\torch\jit\_recursive.py", line 318, in create_script_module
    return create_script_module_impl(nn_module, concrete_type, stubs_fn)
[...]

PR #1105, opened 2020-08-05
Disable memory_format_tutorial on Windows CI

guyang3532:
To resolve https://github.com/pytorch/tutorials/issues/1101

memory_format_tutorial.py uses setattr() to modify the attributes of torch. This will cause CI test failure to other tutorials because they are all executed in one process in sphinx.
This bugs will not influence the linux because tutorials are running parallelly with 20 workers on CircleCI for linux and memory_format_tutorial and other related tutorials are running in different workers.
This pr fix it by disable memory_format_tutorial.py when running windows CI test.

test at #1106


guyang3532:
@brianjo @seemethere 
This will fix the windows CircleCI test failures.
Test passed at #1106

brianjo:
@guyang3532 @seemethere The win_test_worker is going on 3 hours. Could you take a look? I don't think you need to revert, but if you could tweak whatever this is doing with a new PR we can start this build fresh. (Or I can restart that thread.) Thanks  https://app.circleci.com/pipelines/github/pytorch/tutorials?branch=master

PR #1131, opened 2020-08-23
Colab ready static_quantization_tutorial

ParthPatel-ES:
- Now, It is working via Runtime->Run all in Colab.

- Bug Fix:

  Section 3: 
> Imagenet has no public access via torchvision.datasets.ImageNet() function. So, it has removed.  

> Python cell is inserted to copy and extract ImageNet dataset.

> Cell has inserted to put MobileNet V2 model in saved_model_dir.


ParthPatel-ES:
Need to update " https://github.com/ParthPatel-ES/tutorials/blob/gh-pages/_downloads/03ac9a8e1176f5e39736885e8c439a82/static_quantization_tutorial.ipynb " file from my File.

PR #1137, opened 2020-08-25
Python recipe for automatic mixed precision

mcarilli:
^

Still need to tune the model architecture to make mixed precision look really good :-)

After brief slack discussion, the walkthrough is under recipes rather than tutorials.  The walkthrough is single-topic, but I wouldn't call it "bite-size," so IMO the final placement (tutorial vs recipe) remains open to bikeshedding based on how it looks in the preview.


gottbrath:
thanks for submitting this Mike!

brianjo:
Article preview [here](https://deploy-preview-1137--pytorch-tutorials-preview.netlify.app/intermediate/mixed_precision_tutorial.html#troubleshooting).  Might consider removing the links in the page to subsections, in favor of the right nav that contains the same links.

mcarilli:
> Might consider removing the links in the page to subsections, in favor of the right nav that contains the same links.

Not sure what you mean.  Do you mean I should replace things like
[:ref:`Troubleshooting<troubleshooting>`](https://github.com/pytorch/tutorials/pull/1137/files#diff-15941aec48cd514f488513c4cb8a6516R65)
with e.g.
`torch.cuda.amp.autocast <https://pytorch.org/tutorials/intermediate/mixed_precision_tutorial.html#troubleshooting>`_
?
I figured intra-doc references would be _more_ robust if and where I could use them, but I can easily change all refs to use external URLs.

brianjo:
Thanks @mcarilli, I was suggesting removing that links completely as they already exist on the right. 

Screen Shot 2020-09-02 at 3 36 24 PM

PR #1148, opened 2020-09-03
Add CI test for windows

guyang3532:
This pr added windows tutorial ci test using cache for saving time and parallizing on 4 workers with each running about 1 hour.

The data directory is cached on CiecleCI with checksum of Makefile and .circleci-weekly as key. This cache will refresh on Sunday or the Makefile changed. This will save the time for downloading data.

### DIFF
--- a/.circleci/scripts/build_for_windows.sh
+++ b/.circleci/scripts/build_for_windows.sh
@@ -16,7 +16,7 @@ mingw32/bin/mingw-get.exe install mingw32-make
 mingw32/bin/mingw-get.exe install msys-findutils
 mv mingw32/bin/mingw32-make.exe mingw32/bin/make.exe
 curl -k https://eternallybored.org/misc/wget/1.20.3/64/wget.exe -o mingw32/bin/wget.exe
-export PATH="${SOURCE_DIR}/mingw32/bin:$PATH"
+export PATH="${SOURCE_DIR}/mingw32/bin:${SOURCE_DIR}/mingw32/msys/1.0/bin:$PATH"
 
 #install anaconda3
 export CONDA_HOME="${SOURCE_DIR}/conda"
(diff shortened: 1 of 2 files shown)


.circleci/scripts/build_for_windows.sh
```diff
@@ -34,16 +34,77 @@ conda activate testenv
...
+   for filename in $(find prototype_source/ -name '\*.py' -not -path '\*/data/\*'); do
+    if [ $(($count % $NUM_WORKERS)) != $WORKER_ID ]; then
+      echo "Removing runnable code from "$filename
+      python $DIR/remove_runnable_code.py $filename $filename
+    else
+      echo "Keeping "$filename
+      FILES_TO_RUN+=($(basename $filename .py))
+    fi
+    count=$((count+1))
+  done
```
malfet, line 102: Please use nested loops instead of repeating the same code
```suggestion
  for work_directory  in beginner_source intermediate_source advanced_source recipes_source prototype_source; do
    for filename in $(find $work_directory -name '\*.py' -not -path '\*/data/\*'); do
      if [ $(($count % $NUM_WORKERS)) != $WORKER_ID ]; then
        echo "Removing runnable code from "$filename
        python $DIR/remove_runnable_code.py $filename $filename
      else
        echo "Keeping "$filename
        FILES_TO_RUN+=($(basename $filename .py))
      fi
      count=$((count+1))
    done
  done
```

Makefile
```diff
@@ -46,18 +46,18 @@ download:
 	mkdir -p prototype_source/data
 
 	# transfer learning tutorial data
-	wget -N https://download.pytorch.org/tutorial/hymenoptera_data.zip -P $(DATADIR)
+	wget -q -N https://download.pytorch.org/tutorial/hymenoptera_data.zip -P $(DATADIR)
```
malfet, line 49: Can you please explain why those changes are necessary?

beginner_source/hyperparameter_tuning_tutorial.py
```diff
@@ -224,12 +224,12 @@ def train_cifar(config, checkpoint_dir=None, data_dir=None):
         train_subset,
         batch_size=int(config["batch_size"]),
         shuffle=True,
-        num_workers=8)
+        num_workers=0)
```
malfet, line 227: Can you please explain why those are necessary?

beginner_source/hyperparameter_tuning_tutorial.py
```diff
@@ -296,7 +296,7 @@ def test_accuracy(net, device="cpu"):
     trainset, testset = load_data()
 
     testloader = torch.utils.data.DataLoader(
-        testset, batch_size=4, shuffle=False, num_workers=2)
+        testset, batch_size=4, shuffle=False, num_workers=0)
```
malfet, line 299: Same as above

beginner_source/hyperparameter_tuning_tutorial.py
```diff
@@ -224,12 +224,12 @@ def train_cifar(config, checkpoint_dir=None, data_dir=None):
...
         batch_size=int(config["batch_size"]),
         shuffle=True,
-        num_workers=8)
+        num_workers=0)
     valloader = torch.utils.data.DataLoader(
         val_subset,
         batch_size=int(config["batch_size"]),
         shuffle=True,
-        num_workers=8)
+        num_workers=0)
```
malfet, line 232: Same as above

malfet (changes requested):

.circleci/scripts/build_for_windows.sh
guyang3532: done

Makefile
guyang3532: This is downloading quietly to avoid the log being too long.

beginner_source/hyperparameter_tuning_tutorial.py
guyang3532: Just to work around the error happened when spawning threads on windows. This would not impact the performance much and we will investigate it later.

Makefile
malfet: Please correct me if I'm wrong, but this change is unrelated to the goal of this PR. So please, do not make in this change, but feel free to file a separate PR/issue if you want to modify this behavior.

beginner_source/hyperparameter_tuning_tutorial.py
malfet: But this changes the behavior specific to this tutorial. If this tutorial can not be validated on Windows yet, please exclude it from the testing and then work in a separate PR to fix the issue on Windows

malfet (changes requested):
Please do not make modification to the tutorial code in PR that supposed to just add CI testing on a new platform. If modifying the tutorial is prerequisite for enabling CI, please resolve tutorial problems in separate PR first.

malfet (approved):
Thank you for working on that @guyang3532 !

PR #1161, opened 2020-09-18
Add Performance Tuning guide recipe

szmigacz:
Initial version of Performance Tuning Guide.

The guide is an extended version based on materials prepared for [ECCV 2020 Accelerating Computer Vision with Mixed Precision Tutorial](https://nvlabs.github.io/eccv2020-mixed-precision-tutorial/) ([slides](https://nvlabs.github.io/eccv2020-mixed-precision-tutorial/files/szymon_migacz-pytorch-performance-tuning-guide.pdf), [recording](https://youtu.be/9mS1fIYj1So)).

### DIFF
--- a/recipes_source/recipes/tuning_guide.py
+++ b/recipes_source/recipes/tuning_guide.py
@@ -329,20 +329,26 @@ torch.backends.cudnn.benchmark = True
 # perform the required gradient all-reduce.
 
 ###############################################################################
-# Match the order of layers in constructors with order during the execution if training with DistributedDataParallel
-# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Match the order of layers in constructors and during the execution if using DistributedDataParallel(find_unused_parameters=True)
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 # `torch.nn.parallel.DistributedDataParallel <https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html#torch.nn.parallel.DistributedDataParallel>`_
-# uses the order of layers and parameters from constructors to build buckets for
-# `DistributedDataParallel` gradient all-reduce. `DistributedDataParallel`
-# overlaps all-reduce with the backward pass. All-reduce for a particular bucket
-# is asynchronously triggered only when all gradients for parameters in a given
-# bucket are available.
+# with ``find_unused_parameters=True`` uses the order of layers and parameters
+# from constructors to build buckets for `DistributedDataParallel` gradient
+# all-reduce. `DistributedDataParallel` overlaps all-reduce with the backward
+# pass. All-reduce for a particular bucket is asynchronously triggered only when
+# all gradients for parameters in a given bucket are available.
 #
 # To maximize the amount of overlap, the order in constructors should match the
 # order during the execution. If the order doesn't match, then all-reduce for
 # the entire bucket waits for the gradient which is the last to arrive, this may
 # reduce the overlap between backward pass and all-reduce, all-reduce may end up
 # being exposed, which slows down the training.
+#
+# ``DistributedDataParallel`` with ``find_unused_parameters=False`` (which is
+# the default setting) relies on automatic bucket formation based on order of
+# operations encountered during the backward pass. With
+# ``find_unused_parameters=False`` it's not necessary to reorder layers or
+# parameters to achieve optimal performance.
 
 ###############################################################################
 # Load-balance workload in a distributed setting
--- a/recipes_source/recipes/tuning_guide.py
+++ b/recipes_source/recipes/tuning_guide.py
@@ -23,8 +23,8 @@ Presented techniques often can be implemented by changing only a few lines of co
 # ``num_workers`` should be tuned depending on the workload, CPU, accelerator, and location of training data.
 #
 # ``DataLoader`` accepts ``pin_memory`` argument, which defaults to ``False``.
-# In almost all settings it's better to set ``pin_memory=True``,
-# this instructs ``DataLoader`` to use pinned memory and enables faster and asynchronous memory copy from the host to the GPU.
+# When using an accelerator it's better to set ``pin_memory=True``,
+# this instructs ``DataLoader`` to use pinned memory and enables faster and asynchronous memory copy from the host to the accelerator.
 
 ##########################################################
 # Disable debugging APIs
@@ -128,7 +128,7 @@ for param in model.parameters():
 # Use mixed precision and AMP
 # ----------------
 # Mixed precision leverages `Tensor Cores <https://www.nvidia.com/en-us/data-center/tensor-cores/>`_ and offers up to 3x overall speedup on Volta and newer GPU architectures.
-# To use Tensor Cores AMP should be enabled and matrix/tensor dimensions should satisfy requirements for calling Tensor Cores kernels.
+# To use Tensor Cores AMP should be enabled and matrix/tensor dimensions should satisfy requirements for calling kernels that use Tensor Cores.
 #
 # To use Tensor Cores:
 #
--- a/recipes_source/recipes/tuning_guide.py
+++ b/recipes_source/recipes/tuning_guide.py
@@ -8,14 +8,9 @@ accelerate training and inference of deep learning models in PyTorch. Presented
 techniques often can be implemented by changing only a few lines of code and can
 be applied to a wide range of deep learning models across all domains.
 
+General optimizations
+---------------------
 """
-
-# dummy text to correctly generate first header
-# TODO: how to fix this?
-
-###############################################################################
-# General optimizations
-# ---------------------
 
 ###############################################################################
 # Enable async data loading and augmentation
--- a/recipes_source/recipes/tuning_guide.py
+++ b/recipes_source/recipes/tuning_guide.py
@@ -3,8 +3,10 @@ Performance Tuning Guide
 *************************
 **Author**: `Szymon Migacz <https://github.com/szmigacz>`_
 
-Performance Tuning Guide is a set of optimizations and best practices which can accelerate training and inference of deep learning models in PyTorch.
-Presented techniques often can be implemented by changing only a few lines of code and can be applied to a wide range of deep learning models across all domains.
+Performance Tuning Guide is a set of optimizations and best practices which can
+accelerate training and inference of deep learning models in PyTorch. Presented
+techniques often can be implemented by changing only a few lines of code and can
+be applied to a wide range of deep learning models across all domains.
 
 """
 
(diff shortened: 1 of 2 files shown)


szmigacz:
@soumith @ngimel @wconstab ^^^

wconstab:
LGTM.  I can't actually 'approve' since you didn't make me a 'reviewer'.

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+# which means that the data loading is synchronous and done in the main process.
+# As a result the main training process has to wait for the data to be available
+# to continue the execution.
+#
+# Setting ``num_workers > 0`` enables asynchronous data loading and overlap
+# between the training and data loading. ``num_workers`` should be tuned
+# depending on the workload, CPU, accelerator, and location of training data.
+#
+# ``DataLoader`` accepts ``pin_memory`` argument, which defaults to ``False``.
+# When using an accelerator it's better to set ``pin_memory=True``, this
```
shangw-nvidia, line 30: I think you can safely specify `accelerator` to `GPU` here. TPU via pytorch-xla works quite differently: the training script executes in a GCP VM on one machine, and the TPUs are attached in another machine (VM sends the input data and the IR to TPUs' machine via Ethernet). I don't think setting `pin_memor=True` for TPU do much (cuz this would refers to the VM), and pytorch-xla has its own dataloader to pipeline data uploading and graph execution.

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+#
+# This is also applicable to 1d and 3d convolutions as long as ``BatchNorm`` (or
+# other normalization layer) normalizes on the same dimension as convolution's
+# bias.
+
+###############################################################################
+# Use parameter.grad = None instead of model.zero_grad()
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Instead of calling:
+model.zero_grad()
```
shangw-nvidia, line 90: Also, `optimizer.zero_grad()`

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+# ``nn.Conv2d(..., bias=False, ....)``. Bias is not needed because in the first
+# step ``BatchNorm`` subtracts the mean, which effectively cancels out the
+# effect of bias.
+#
+# This is also applicable to 1d and 3d convolutions as long as ``BatchNorm`` (or
+# other normalization layer) normalizes on the same dimension as convolution's
+# bias.
+
+###############################################################################
+# Use parameter.grad = None instead of model.zero_grad()
```
mcarilli, line 87: model.zero_grad() or optimizer.zero_grad()

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Instead of calling:
+model.zero_grad()
+
+###############################################################################
+# to zero out gradients, use the following method instead:
+
+for param in model.parameters():
+    param.grad = None
+
```
mcarilli, line 97: Setting grads to None recently got first class citizenship as a kwarg to zero_grad.
https://pytorch.org/docs/master/optim.html#torch.optim.Optimizer.zero_grad
You can also say something like "Alternatively, in recent versions of Pytorch, call `model` or `optimizer.zero_grad(set_to_none=True)`."

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+#
+# Support for `channels_last` is new and experimental, but it's expected to work
+# for standard computer vision models (e.g. ResNet-50, SSD). To convert models
+# to `channels_last` format follow
+# `Channels Last Memory Format Tutorial <https://pytorch.org/tutorials/intermediate/memory_format_tutorial.html>`_.
+# The tutorial includes a section on
+# `converting existing models <https://pytorch.org/tutorials/intermediate/memory_format_tutorial.html#converting-existing-models>`_.
+
+###############################################################################
+# Parallelize kernel execution
```
ngimel, line 147: Maybe we should remove this section completely for now, and reincarnate it after 1.7 with for_each operations appears? I'd hate to see people modifying their codes to do it now when in just a few weeks it will be so much better.

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+# large layer outputs that have small re-computation cost. The example target
+# layers are activation functions (e.g. ``ReLU``, ``Sigmoid``, ``Tanh``),
+# up/down sampling and matrix-vector operations with small accumulation depth.
+#
+# PyTorch supports a native
+# `torch.utils.checkpoint <https://pytorch.org/docs/stable/checkpoint.html>`_
+# API to automatically perform checkpointing and recomputation.
+
+###############################################################################
+# Pre-allocate memory in case of variable input length
```
ngimel, line 195: This is a GPU-specific optimization because CPU is unlikely to run out of memory, and does not rely on caching allocator.

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+# Avoid unnecessary synchronizations, to let the CPU run ahead of the
+# accelerator as much as possible to make sure that the accelerator work queue
+# contains many operations.
+#
+# When possible, avoid operations which require synchronizations, for example:
+#
+# * ``print(CUDA_tensor)``
+# * ``CUDA_tensor.item()``
+# * memory copies: ``tensor.cuda()``,  ``CUDA_tensor.cpu()`` and equivalent
+#   ``tensor.to(device)`` calls
```
ngimel, line 257: makes sense to maybe add .nonzero() and result of masking operations in here (`(x!=0).all()` etc)

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+###############################################################################
+# Use efficient data-parallel backend
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# PyTorch has two ways to implement data-parallel training:
+#
+# * `torch.nn.DataParallel <https://pytorch.org/docs/stable/generated/torch.nn.DataParallel.html#torch.nn.DataParallel>`_
+# * `torch.nn.parallel.DistributedDataParallel <https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html#torch.nn.parallel.DistributedDataParallel>`_
+#
+# ``DistributedDataParallel`` offers much better performance and scaling to
+# multiple-GPUs. In the case of ``DataParallel``, a single CPU core and a single
```
ngimel, line 319: "single CPU core" is not strictly true, better to remove it.

ngimel (commented):
This looks good, I left a few comments. 
I'm not sure about how optimizations are divided into GPU-specific and not, some of supposedly general optimizations looked gpu-specific to me (e.g. pin memory in dataloader?). This division probably needs to be rethought.

recipes_source/recipes/tuning_guide.py
szmigacz: I'll add optimizer.zero_grad()

recipes_source/recipes/tuning_guide.py
szmigacz: True, I forgot that it's already in master, I'll add this

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+model.zero_grad()
+
+###############################################################################
+# to zero out gradients, use the following method instead:
+
+for param in model.parameters():
+    param.grad = None
+
+###############################################################################
+# The second code snippet doesn't call memset on each individual parameter, also
```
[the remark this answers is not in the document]
szmigacz: ok

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+# the subsequent backward pass uses assignment instead of addition assignment to
```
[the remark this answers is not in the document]
szmigacz: I wanted to say that with the optimization we do grad = value (1 mem op, just a write), but w/o we do grad += value (2 mem ops). I think "+=" is usually called 'addition assignment', why do you want to change to 'addition'?

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+# store gradients, this reduces number of memory operations.
+
```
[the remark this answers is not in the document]
szmigacz: ok

recipes_source/recipes/tuning_guide.py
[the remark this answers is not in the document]
szmigacz: well, 'avoid launching small kernels' is a general topic and there are many ways to achieve that (e.g. JIT fusion or rewriting python loops to use advanced indexing etc.), suggested optimization is another strategy to reduce number of small kernels. Maybe section should be renamed to "Stack tensors to parallelize kernel execution"?

recipes_source/recipes/tuning_guide.py
szmigacz: I'll move it to GPU-specific section

recipes_source/recipes/tuning_guide.py
szmigacz: does DataParallel use multiple processes? if not then effectively it would be a single core due to GIL, am I missing something? i.e. data loading can of course still use multiple cores, but that's slightly orthogonal.

recipes_source/recipes/tuning_guide.py
szmigacz: I'll add

recipes_source/recipes/tuning_guide.py
szmigacz: I'll add

recipes_source/recipes/tuning_guide.py
szmigacz: makes sense, I'll switch it back to `GPU`.

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+# between the training and data loading. ``num_workers`` should be tuned
+# depending on the workload, CPU, accelerator, and location of training data.
+#
+# ``DataLoader`` accepts ``pin_memory`` argument, which defaults to ``False``.
+# When using an accelerator it's better to set ``pin_memory=True``, this
+# instructs ``DataLoader`` to use pinned memory and enables faster and
+# asynchronous memory copy from the host to the accelerator.
+
+###############################################################################
+# Disable debugging APIs
```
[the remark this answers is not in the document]
szmigacz: Makes sense, I'll move it to the end of the section.

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+#
+# If a ``nn.Conv2d`` layer is directly followed by a ``nn.BatchNorm2d`` layer,
+# then the bias in the convolution is not needed, instead use
+# ``nn.Conv2d(..., bias=False, ....)``. Bias is not needed because in the first
+# step ``BatchNorm`` subtracts the mean, which effectively cancels out the
+# effect of bias.
+#
+# This is also applicable to 1d and 3d convolutions as long as ``BatchNorm`` (or
+# other normalization layer) normalizes on the same dimension as convolution's
+# bias.
```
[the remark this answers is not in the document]
szmigacz: ok

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+###############################################################################
+# The second code snippet doesn't call memset on each individual parameter, also
+# the subsequent backward pass uses assignment instead of addition assignment to
+# store gradients, this reduces number of memory operations.
+
+###############################################################################
+# Fuse pointwise operations
+# ~~~~~~~~~~~~~~~~~~~~~~~~~
+# Pointwise operations (elementwise addition, multiplication, math functions -
+# ``sin()``, ``cos()``, ``sigmoid()`` etc.) should be fused into a single kernel
```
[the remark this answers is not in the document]
szmigacz: ok

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+# to amortize memory access time and kernel launch time.
+#
+# `PyTorch JIT <https://pytorch.org/docs/stable/jit.html>`_ can fuse kernels
+# automatically, although there could be additional fusion opportunities not yet
+# implemented in the compiler, and not all device types are supported equally.
+#
+# Unfused pointwise operations are memory-bound, for each unfused operation
```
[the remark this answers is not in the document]
szmigacz: ok

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+# Fused operator launches only one kernel for multiple fused pointwise ops and
+# loads/stores data only once to the memory. This makes JIT very useful for
+# activation functions, optimizers, custom RNN cells etc.
+#
+# In the simplest case fusion can be enabled by applying
+# `torch.jit.script <https://pytorch.org/docs/stable/generated/torch.jit.script.html#torch.jit.script>`_
+# decorator to the function definition.
+# Refer to
+# `TorchScript documentation <https://pytorch.org/docs/stable/jit.html>`_
+# for more advanced use cases.
```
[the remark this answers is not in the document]
szmigacz: ok, I'll add example with gelu activation

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,376 @@
...
+###############################################################################
+# Enable channels_last memory format for computer vision models (experimental)
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# PyTorch 1.5 introduced support for `channels_last` memory format for
+# convolutional networks. This format is meant to be used in conjunction with
+# `AMP <https://pytorch.org/docs/stable/amp.html>`_ to further accelerate
+# convolutional neural networks with
+# `Tensor Cores <https://www.nvidia.com/en-us/data-center/tensor-cores/>`_.
+#
+# Support for `channels_last` is new and experimental, but it's expected to work
```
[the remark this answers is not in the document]
szmigacz: ok

recipes_source/recipes/tuning_guide.py
szmigacz: I agree. If there is going to be a better way then let's remove this section for now.

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,366 @@
...
+# `DistributedDataParallel` gradient all-reduce. `DistributedDataParallel`
+# overlaps all-reduce with the backward pass. All-reduce for a particular bucket
+# is asynchronously triggered only when all gradients for parameters in a given
+# bucket are available.
+#
+# To maximize the amount of overlap, the order in constructors should match the
+# order during the execution. If the order doesn't match, then all-reduce for
+# the entire bucket waits for the gradient which is the last to arrive, this may
+# reduce the overlap between backward pass and all-reduce, all-reduce may end up
+# being exposed, which slows down the training.
```
mcarilli, line 347: This entire section can be removed.  DDP now rebuckets based on a live backward pass.  There's no need to scare readers into a potentially difficult refactor.

szmigacz: @mcarilli Is it also true with find_unused_parameters=True? I think if it's True then DDP still follows order from constructors? So either we remove this section or clarify that it's only needed if find_unused is True (if I'm right here).

recipes_source/recipes/tuning_guide.py
ngimel: DataParallel uses threads, but GIL is released for every op as soon as it gets to c++. So, Distributed is still way better, and that's what we should say. but describing the exact details of what happens in DataParallel is beyond the scope of this guide.

recipes_source/recipes/tuning_guide.py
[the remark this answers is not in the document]
ngimel: I think just "addition" would be clearer? But up to you.

recipes_source/recipes/tuning_guide.py
mcarilli: I believe the dynamic rebucketing is orthogonal to find_unused_parameters=True.  I'll ping @mrshenli and @pritamdamania87 (@pritamdamania ?) on pytorch slack for confirmation.

recipes_source/recipes/tuning_guide.py
mcarilli: Are we planning to publicize `_foreach*` in 1.7?  I thought we were keeping it underscored for now so we can tweak the interface if we want.  (Personally, I think the leanest public interface would be: extend existing ops to accept lists of tensors/scalars, rather than exposing a new set of torch.foreach* functions)

recipes_source/recipes/tuning_guide.py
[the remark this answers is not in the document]
szmigacz: Probably you're right, 'addition' will be less confusing. I'll change

recipes_source/recipes/tuning_guide.py
szmigacz: Ok, makes sense. I'll remove that sentence, I'll just say that DDP offers better perf and better scaling and refer to the section from cuda best practices doc.

recipes_source/recipes/tuning_guide.py
pritamdamania87: As per my understanding rebucketing is done only if `find_unused_parameters=False`. So I think we should update this section to first mention about rebucketing and then mention that if `find_unused_parameters=True`, this option is disabled and we need to worry about the order in that case. We should retain this section for the `find_unused_parameters=True` case.

mcarilli: Heh, i was just checking myself.  Yeah that looks right.  https://github.com/pytorch/pytorch/blob/92f8f75c59ce72b2862ad7f32a78a624b62e451d/torch/csrc/distributed/c10d/reducer.cpp#L491-L500

szmigacz: I changed that section and said that this optimization is only for find_unused = True, here is a direct link to the preview: https://deploy-preview-1161--pytorch-tutorials-preview.netlify.app/recipes/recipes/tuning_guide.html#match-the-order-of-layers-in-constructors-and-during-the-execution-if-using-distributeddataparallel-find-unused-parameters-true 
@mcarilli does it look ok?

mcarilli: Almost perfect, notes inline.

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,370 @@
...
+# ``no_sync()`` should applied to first ``N-1`` iterations of gradient
+# accumulation, the last iteration should follow the default execution and
+# perform the required gradient all-reduce.
+
+###############################################################################
+# Match the order of layers in constructors and during the execution if using DistributedDataParallel(find_unused_parameters=True)
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# `torch.nn.parallel.DistributedDataParallel <https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html#torch.nn.parallel.DistributedDataParallel>`_
+# with ``find_unused_parameters=True`` uses the order of layers and parameters
+# from constructors to build buckets for ``DistributedDataParallel`` gradient
```
mcarilli, line 336: I think `uses the order of layers and parameters from constructors` is a little ambiguous, I would say `model constructors`, not just `constructors`.

recipes_source/recipes/tuning_guide.py
```diff
@@ -0,0 +1,370 @@
...
+# all-reduce. ``DistributedDataParallel`` overlaps all-reduce with the backward
+# pass. All-reduce for a particular bucket is asynchronously triggered only when
+# all gradients for parameters in a given bucket are available.
+#
+# To maximize the amount of overlap, the order in constructors should match the
```
mcarilli, line 341: s/match/roughly match/
Sometimes it's hard to get a perfect match between constructor order and execution order, and backward pass op order is not in general strictly the reverse of forward pass op order.

szmigacz: Thanks, changed in: https://github.com/szmigacz/tutorials/commit/3c4e84dabaa75d644768bc0d8092130102f18b99

recipes_source/recipes/tuning_guide.py
szmigacz: Thanks, changed in: https://github.com/szmigacz/tutorials/commit/3c4e84dabaa75d644768bc0d8092130102f18b99

szmigacz:
@brianjo - would you be able to help with merging this PR?

PR #1164, opened 2020-09-21
Update tutorials to use TensorPipeRpcBackendOptions.

pritamdamania87:


rohan-varma (approved):
Thanks! I'm assuming we would need @lw's PR for inferring the right type of backend for this to work properly?

lw:
Thanks for doing this! That PR is in https://github.com/pytorch/pytorch/pull/45065, and it's ready for review.

pritamdamania87:
> Thanks! I'm assuming we would need @lw's PR for inferring the right type of backend for this to work properly?

I think this would work without the PR since we've updated the tutorials to use the right options when the default is TensorPipe.

mrshenli:
Can we hold on to this PR for several days. There is a chance we might need to revert default backend to Process Group. @lw is discussing that with @jspisak

pritamdamania87:
Landing this since we decided to go with TensorPipe as the default backend.

PR #1169, opened 2020-09-29
Fix MultiScaleRoIAlign featmap_names to string

yonigottesman:
When using a backbone that returns a tensor an OrderedDict with key '0' is created (string not int).
So when creating a MultiScaleRoIAlign the featmap_names should be a string '0'.


holly1238:
Closing this PR. This was fixed and merged in https://github.com/pytorch/tutorials/pull/1085.

PR #1176, opened 2020-10-07
Benchmark Recipe

heitorschueroff:
This PR adds a recipe for PyTorch benchmark module.

TODO

- [x] Fix output formatting

cc @robieta, @brianjo, @ngimel

### DIFF
--- a/recipes_source/recipes/benchmark.py
+++ b/recipes_source/recipes/benchmark.py
@@ -1,0 +1,892 @@
+"""
+PyTorch Benchmark
+====================================
+This recipe provides a quick-start guide to using PyTorch
+``benchmark`` module to measure and compare code performance.
+
+Introduction
+------------
+Benchmarking is an important step in writing code. It helps
+us validate that our code meets performance expectations,
+compare different approaches to solving the same problem and
+prevent performance regressions.
+
+There are many options when it comes to benchmarking PyTorch code
+including the Python builtin ``timeit`` module. However, benchmarking
+PyTorch code has many caveats that can be easily overlooked such as
+managing the number of threads and synchronizing CUDA devices. Moreover,
+generating Tensor inputs for benchmarking can be quite tedious.
+
+This recipe demonstrates how to use PyTorch ``benchmark`` module to avoid
+common mistakes while making it easier to compare performance of
+different code, generate input for benchmarking and more.
+
+Setup
+-----
+Before we begin, install ``torch`` if it isn’t already available.
+
+::
+
+   pip install torch
+
+"""
+
+
+######################################################################
+# Steps
+# -----
+#
+# 1. Defining functions to benchmark
+# 2. Benchmarking with ``timeit.Timer``
+# 3. Benchmarking with ``torch.utils.benchmark.Timer``
+# 4. Benchmarking with `Blocked Autorange`
+# 5. Comparing benchmark results
+# 6. Saving/Loading benchmark results
+# 7. Generating inputs with `Fuzzed Parameters`
+# 8. Collecting instruction counts with `Callgrind`
+#
+# 1. Defining functions to benchmark
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# As of the time of this writing, `torch.dot <https://pytorch.org/docs/stable/generated/torch.dot.html?highlight=dot#torch.dot>`__
+# does not support batched mode, so we will compare two approaches to
+# implementing it using existing ``torch`` operators: one approach uses a
+# combination of ``mul`` and ``sum`` while the other reduces the problem to ``bmm``.
+#
+
+import torch
+
+
+def batched_dot_mul_sum(a, b):
+    '''Computes batched dot by multiplying and summing'''
+    return a.mul(b).sum(-1)
+
+
+def batched_dot_bmm(a, b):
+    '''Computes batched dot by reducing to bmm'''
+    a = a.reshape(-1, 1, a.shape[-1])
+    b = b.reshape(-1, b.shape[-1], 1)
+    return torch.bmm(a, b).flatten(-3)
+
+
+# Input for benchmarking
+x = torch.randn(10000, 64)
+
+# Ensure that both functions compute the same output
+assert batched_dot_mul_sum(x, x).allclose(batched_dot_bmm(x, x))
+
+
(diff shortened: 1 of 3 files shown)


heitorschueroff:
@brianjo Is there a better way to display output from code besides putting it in a code block? Something like what you'd get from a python notebook?

recipes_source/recipes/benchmark.py
```diff
@@ -0,0 +1,828 @@
...
+######################################################################
+# 6. Generating bechmark inputs with Fuzzed Parameters
+#
+# We'll take a brief detour and use fuzzed inputs to discuss transpose
+# and contiguous before returning to shift.
+#
+
+from torch.utils.benchmark import Fuzzer, FuzzedParameter, FuzzedTensor, ParameterAlias
+
+example_fuzzer = Fuzzer(
```
ngimel, line 397: a description of what this fuzzer does (generates 2d tensors with sizes distributed according to specified distributions) is useful. You are talking to people who see Fuzzer and FuzzedParameter for the first time.

recipes_source/recipes/benchmark.py
```diff
@@ -0,0 +1,828 @@
...
+# Times are in nanoseconds (ns).
+
+######################################################################
+# The fuzzed benchmarks reveal several noteworthy features:
+#
+# If a Tensor is already contiguous we do not need to construct a new
+# Tensor and the operation is extremely cheap. O(100 ns)
+#
+# For N x 1 Tensors transpose requires that we create a new Tensor,
+# but we can reuse the same buffer.
```
ngimel, line 448: `buffer` is an unclear and overloaded word, "we don't have to do any physical data movement"

recipes_source/recipes/benchmark.py
```diff
@@ -0,0 +1,828 @@
...
+#
+# For N x N tensors, either contiguous or transposed contiguous will
+# be expensive depending on the underlying data layout.
+
+######################################################################
+# 6.1. Canned fuzzers: back to our x+1 kernel
+#
+# When benchmarking an op, there are a lot of things to consider:
```
ngimel, line 456: yeah, something along these lines should be in the beginning of section 6, and it's ok to repeat it here, now introducing fuzzers for unary and binary ops.

recipes_source/recipes/benchmark.py
```diff
@@ -0,0 +1,828 @@
...
+#
+
+from torch.utils.benchmark.op_fuzzers import unary
+
+def skip_if(x):
+    # We know that around this point our custom kernel is slower than vanilla add, so
+    # there's not much useful signal from testing larger Tensors.
+    return x.numel() > 64 * 1024
+
+def to_description(x, x_params):
```
ngimel, line 470: comment what this function does

recipes_source/recipes/benchmark.py
```diff
@@ -0,0 +1,828 @@
...
+# [5]    [325, 128]                                        [::2, :]
+# [6]    [118, 166]                                        
+# [7]    [6309]                                            
+# [8]    [26, 512]                                         
+# [9]    [27, 225]           (1, 0)                        
+# [10]   [380, 58]                                         
+
+######################################################################
+# There's an interesting pattern in our benchmark results above. The
+# dicontiguous input tensors [1], [2], [5] and [9] take much longer
```
ngimel, line 543: "discontiguous"

recipes_source/recipes/benchmark.py
```diff
@@ -0,0 +1,828 @@
...
+# to run on than expected. The reason is that we are cloning the
```
ngimel, line 544: you don't say which implementation takes much longer to run, and also how do you know what to expect?

ngimel (commented):
This tutorial feels like 2 in 1 - basic use of benchmarking tools, and how to optimize your custom op implementation. Many people to whom benchmarking tools are addressed don't write a single line of c++ in their daily lives, and yet would still benefit from knowing how to benchmark their code correctly. So it might make sense to either split it in 2, or structure it in such a way so that there's a first part for everyone, and second part for those who are still able to follow. 
I didn't get to the end, but let me know if it makes sense.

recipes_source/recipes/README.txt
```diff
@@ -55,8 +55,16 @@ PyTorch Recipes
 
 14. mobile_perf.py
          PyTorch Mobile Performance Recipes
-         https://pytorch.org/tutorials/recipes/mobile_perf.html
+         https://pytorch.org/tutorials/recipes/recipes/mobile_perf.html
```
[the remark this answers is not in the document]
heitorschueroff: The actual link has recipes twice in it.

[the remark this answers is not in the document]
heitorschueroff: When I was testing the links had recipes twice in it, but now I'm trying it and it was right before. I'll fix it.

recipes_source/recipes/benchmark.py
```diff
@@ -0,0 +1,828 @@
...
+# 2. Writing the custom C++ extension
+# 3. Simple benchmark using timeit.Timer
+# 4. Better benchmark using torch.benchmark.Timer
+# 4.1. Timer.blocked_autorage
+# 4.2. Why runtime awareness matters
+# 5. Comparing different benchmarks
+# 6. Generating bechmark inputs with Fuzzed Parameters
+# 6.1. Canned fuzzers: back to our x+1 kernel
+# 7. Taking more precise measurements with instruction counting
+#
```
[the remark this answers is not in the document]
heitorschueroff: I agree with you! We will update it so the first few steps are targeted to general users and we will keep it python only. Later when looking into callgrind stats we might add a custom C++ op.

recipes_source/recipes/benchmark.py
```diff
@@ -0,0 +1,765 @@
...
+    cpp_sources=batched_dot_src,
+    extra_cflags=['-O3'],
+    extra_include_paths=[
+        # `load_inline` needs to know where to find Pybind11 headers.
+        os.path.join(os.getenv('CONDA_PREFIX'), 'include')
+    ],
+    functions=['batched_dot_mul_sum_v0', 'batched_dot_mul_sum_v1']
+)
+
+module_import_str = f"""\
```
ngimel, line 612: a comment on what's going on here?

ngimel (approved):
python portion looks good now! C++ section could use some updating once latest improvements that avoid python altogether are landed.

Issue #1108, opened 2020-08-05
Optimize Tutorial Build with Docker Images


brianjo:
Tutorials currently runs the entire build from scratch multiple times for every PR merge. 
Persisting a docker image after the tutorial build will allow us to do incremental builds on tutorials. 
@seemethere and I met to discuss updating the build with this in mind and we think it's possible to persist the output of the build and do these incrementally.
We'll try to get this going over the next couple of weeks.

malfet:
I wonder if we can re-use sccache for tutorial builds?

brianjo:
Hey all, @v-jizhang is going to take a look at this. He's most recently worked on the Buck Circle CI builds. Thanks!

v-jizhang:
@brianjo I'll take a look, thank you!

v-jizhang:
I think we can utilize one of the CircleCI’s caching mechanisms [DLC, Docker Layer Caching.](https://circleci.com/docs/2.0/docker-layer-caching/) What we need to do is to change the current steps into docker building steps(in file Dockerfile) based on the existing docker image (tag 291). By turning on DLC in config.yml, CircleCI will cache the layers(steps) automatically for us.

v-jizhang:
DLC has a 50G disk space limit, not enough for us. We have to save docker images somewhere else.
We can build 20 images and each image includes a list of tutorial files (the same as FILES_TO_RUN in build.sh) that built into it.
When a PR comes in, if

- There are new files or non-tutorial file changes, build everything from beginning and save the new images, 20 in total. This should not happen very often.
- Only tutorial file changes, rebuild these images that include the changed files and save the new images. For all other images, run step 5 (Copy generated files to S3, tag with commit ID) in build.sh only.

holly1238:
This issue was addressed by https://github.com/pytorch/tutorials/pull/1184

PR #1184, opened 2020-10-12
Added cache for incremental build.

v-jizhang:
Added cache for incremental build.


v-jizhang:
Fix for #1108

seemethere (commented):
This looks mostly okay, but wouldn't this only save the cache for the current PR / job?

Like, from my overview, it seems to me that the first runs would take exactly the same amount of time as it currently takes. Only subsequent runs would actually benefit from this cache.

v-jizhang:
> This looks mostly okay, but wouldn't this only save the cache for the current PR / job?
> 
> Like, from my overview, it seems to me that the first runs would take exactly the same amount of time as it currently takes. Only subsequent runs would actually benefit from this cache.

Ii caches for master branch too. And yes, the first time run does not have cache to use and subsequent run benefit for the cache.

brianjo:
I'm thinking about merging this if we pass tests. Could reviewers chime in on it?

ranman (approved):

PR #1185, opened 2020-10-12
Update tensors and autograd

suraj813:
Refactor content of tensors and autograd in 60 Minute Blitz


suraj813:
Add "What is PyTorch" context to blitz homepage

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,166 @@
...
-print(x)
- 
-###############################################################
-# Construct a randomly initialized matrix:
+data = [[1, 2],[3, 4]]
+np_array = np.array(data)
 
-x = torch.rand(5, 3)
-print(x)
+tnsr_from_data = torch.tensor(data)
```
albanD, line 32: I think here you want to mention that the types are inferred either from the data or from the numpy array (will echo later in the error you try to show).

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,166 @@
...
+######################################################################
+# Tensor Attributes
+# ~~~~~~~~~~~~~~~~~
+#
+# Tensors have attributes that describe their contents and functions. They
+# also have attributes that autograd uses to keep track of them in the
+# computational graph (more on this in the next section).
+#
+# **Docs Issues** - https://pytorch.org/docs/stable/tensor_attributes.html
+# is not comprehensive (missing data, grad, grad_fn, shape). Contains
```
albanD, line 83: `data` and `grad_fn` are explicitly not mentioned as they should not be used. They are just left for backward compatibility. And will hopefully go away soon(ish).
`grad` and `shape` do appear to be missing.

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,166 @@
...
+# ``memory_format`` which is not an attribute
```
albanD, line 84: Does that mean that we have no way to know the memory format of a given Tensor?

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,166 @@
...
-print(x[:, 1])
+tnsr = torch.rand(3,4)
 
-###############################################################
-# Resizing: If you want to resize/reshape tensor, you can use ``torch.view``:
-x = torch.randn(4, 4)
-y = x.view(16)
-z = x.view(-1, 8)  # the size -1 is inferred from other dimensions
-print(x.size(), y.size(), z.size())
+print(f"Data stored in tensor:\n{tnsr.data}\n")
```
albanD, line 89: nope nope nope nope nope!

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,166 @@
...
 
-###############################################################
-# **Read later:**
-#
-#
-#   100+ Tensor operations, including transposing, indexing, slicing,
-#   mathematical operations, linear algebra, random numbers, etc.,
-#   are described
-#   `here <https://pytorch.org/docs/torch>`_.
+# Autograd-related attributes; more on this later
```
albanD, line 96: I think these can wait until the autograd part no?

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,166 @@
...
+# Settings.
 #
-# Converting a Torch Tensor to a NumPy Array
-# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-a = torch.ones(5)
-print(a)
+# We move our tensor to the GPU if available
+if torch.cuda.is_available:
+  tnsr.to('cuda')
```
albanD, line 125: This doesn't actually work as the `.to()` op is not inplace. You need to save the result

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,166 @@
...
+print(f"Does tensor require gradients: {tnsr.requires_grad}")
+print(f"Accumulated gradients of tensor: {tnsr.grad}")
+print(f"Function that computed this tensor's gradient: {tnsr.grad_fn}")
+
+
+
+######################################################################
+# --------------
 #
-# NumPy Bridge
```
albanD, line 138: Did the NumPy Bridge section go away? I do think it is a very important one.

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,166 @@
...
+  tnsr.to('cuda')
 
-###############################################################
-#
 
-b = a.numpy()
-print(b)
+######################################################################
+# Try out some of the operations from the list. You may come across multiple
+# syntaxes for the same operation. Don’t get confused by the aliases!
```
albanD, line 130: I'm not sure what you refer to as aliases here?

albanD: I think it's fine to mention that some ops can have different names but I would still keep all the discussion around numpy arrays as most users these days come from numpy and heavily use numpy around pytorch!

beginner_source/blitz/autograd_tutorial.py
```diff
@@ -1,198 +1,264 @@
...
+---------------------------------
 
-Central to all neural networks in PyTorch is the ``autograd`` package.
-Let’s first briefly visit this, and we will then go to training our
-first neural network.
+Autograd is PyTorch’s automatic differentiation engine that powers
+neural network training. In this section, you will get a conceptual
+understanding of how autograd works under the hood.
 
+Skip to advanced readings:
```
albanD, line 10: Why isn't this at the end as a "going further" section?

albanD (commented):
I only had time to look at the tensor section, will do the autograd later.

beginner_source/blitz/tensor_tutorial.py
suraj813: I think so, yes. Not sure if I'm doing this wrong
```
In [52]: a = a.to(memory_format=torch.channels_last)                                                                                                                                                     

In [53]: a.memory_format                                                                                                                                                                                 
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-53-a1dabca59cd9> in <module>
----> 1 a.memory_format

AttributeError: 'Tensor' object has no attribute 'memory_format'

```

suraj813: Looks like it, unless I'm mistaken 

```
In [52]: a = a.to(memory_format=torch.channels_last)                                                                                                                                                     

In [53]: a.memory_format                                                                                                                                                                                 
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-53-a1dabca59cd9> in <module>
----> 1 a.memory_format

AttributeError: 'Tensor' object has no attribute 'memory_format'
```

beginner_source/blitz/tensor_tutorial.py
suraj813: Yes, removing from here

beginner_source/blitz/tensor_tutorial.py
suraj813: Added it back

beginner_source/blitz/autograd_tutorial.py
suraj813: Fixed

beginner_source/blitz/autograd_tutorial.py
```diff
@@ -1,198 +1,276 @@
...
+# Differentiation in Autograd
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-``Tensor`` and ``Function`` are interconnected and build up an acyclic
-graph, that encodes a complete history of computation. Each tensor has
-a ``.grad_fn`` attribute that references a ``Function`` that has created
-the ``Tensor`` (except for Tensors created by the user - their
-``grad_fn is None``).
+# The ``requires_grad`` flag lets autograd know
+# if we need gradients w.r.t. these tensors. If it is ``True``, autograd
```
ranman, line 69: This doesn't appear to render correctly as text in the preview: https://deploy-preview-1185--pytorch-tutorials-preview.netlify.app/beginner/blitz/autograd_tutorial.html#differentiation-in-autograd

beginner_source/blitz/autograd_tutorial.py
```diff
@@ -1,198 +1,276 @@
...
+#
+# Conceptually, autograd keeps a record of data (tensors) & all executed
+# operations (along with the resulting new tensors) in a directed acyclic
+# graph (DAG) consisting of
+# `Function <https://pytorch.org/docs/stable/autograd.html#torch.autograd.Function>`__
+# objects. In this DAG, leaves are the input tensors, roots are the output
+# tensors. By tracing this graph from roots to leaves, you can
+# automatically compute the gradients using the chain rule.
+#
+# In a forward pass, autograd does two things simultaneously: \* run the
```
ranman, line 195: are the \* here supposed to render as bullet points? they don't appear to in the preview.

beginner_source/blitz/autograd_tutorial.py
```diff
@@ -1,198 +1,276 @@
...
+# requested operation to compute a resulting tensor, and \* maintain the
+# operation’s *gradient function* in the DAG. This is stored in the
+# resulting tensor’s .\ ``grad_fn`` attribute.
+#
+# The backward pass kicks off when ``.backward()`` is called on the DAG
+# root. Autograd then \* computes the gradients from each ``.grad_fn``, \*
+# accumulates them in the respective tensor’s ``.grad`` attribute, and \*
+# using the chain rule, propagates all the way to the leaf tensors.
+#
```
ranman, line 204: Are there any pictures or graphics that would work to show this?

beginner_source/blitz/autograd_tutorial.py
```diff
@@ -1,198 +1,276 @@
...
+x = torch.rand(5, 5)
+y = torch.rand(5, 5)
+z = torch.rand((5, 5), requires_grad=True)
+
+a = x + y
+print("Does `a` require gradients?")
+print(a.requires_grad==True)
+b = x + z
+print("Does `b` require gradients?")
+print(b.requires_grad==True)
```
ranman, line 238: would it be helpful to show the output here?

Also would it be sufficient to simply say `print(a.requires_grad)` ?

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,189 @@
...
 
-Getting Started
----------------
+Tensors are a specialized data structure that are very similar to arrays
+and matrices. In PyTorch, we use tensors to encode the inputs and
+outputs of a model, as well as the model’s parameters.
 
-Tensors
-^^^^^^^
+Tensors are similar to NumPy’s ndarrays, except that tensors can run on
```
ranman, line 9: Might be worth adding "on GPUs or other specialized hardware"

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,189 @@
...
 
-x = torch.rand(5, 3)
-print(x)
+data = [[1, 2],[3, 4]]
+np_array = np.array(data)
 
-###############################################################
-# Construct a matrix filled zeros and of dtype long:
+tnsr_from_data = torch.tensor(data)
+tnsr_from_np = torch.from_numpy(np_array)
```
ranman, line 33: Is it worth calling out that you can also go *back* to numpy? I'm not sure it is but just throwing it out there for consideration. Maybe just link to the "bridge back to numpy" section below.

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,189 @@
...
-# There are multiple syntaxes for operations. In the following
-# example, we will take a look at the addition operation.
-#
-# Addition: syntax 1
-y = torch.rand(5, 3)
-print(x + y)
 
-###############################################################
-# Addition: syntax 2
+shape = (2,3,)
```
ranman, line 62: I think you should add a very brief explanation of shape here. It will be obvious to most but for some it a one sentence explanation will provide a lot of value.

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,189 @@
...
-# .. note::
-#     Any operation that mutates a tensor in-place is post-fixed with an ``_``.
-#     For example: ``x.copy_(y)``, ``x.t_()``, will change ``x``.
+
+######################################################################
+# Tensor Attributes
+# ~~~~~~~~~~~~~~~~~
 #
-# You can use standard NumPy-like indexing with all bells and whistles!
+# Tensor attributes describe their shape data type and where they live.
```
ranman, line 82: need commas after shape and data type - "where they live" makes sense to me but I worry it might not translate well... maybe "on which device the tensor is stored" or something similar.

ranman (changes requested):

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,189 @@
...
-x = torch.empty(5, 3)
-print(x)
- 
-###############################################################
-# Construct a randomly initialized matrix:
+######################################################################
+# Tensor Initialization
+# ~~~~~~~~~~~~~~~~~~~~~
+#
+# Tensors can be initialized in various ways. Take a look at the following examples
```
soumith, line 24: each introduction of a concept followed by a code snippet. one at a time.
Cleanly separating introducing concepts is way more important than minimizing code.

--------------

Tensors can be initialized in various ways. Take a look at the following examples

Directly from data

```
data = [[1, 2],[3, 4]]
x_data = torch.tensor(data)
print(x_data)
```

From a NumPy array:

```
x_np = np.array(data)
x_data = torch.from_numpy(x_np)
print(x_data)
```

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,189 @@
...
+# Tensors can be created directly by passing a Python list or sequence using the ``torch.tensor()`` constructor. The data type is automatically inferred from the data.
 
-x = torch.rand(5, 3)
-print(x)
+data = [[1, 2],[3, 4]]
+np_array = np.array(data)
 
-###############################################################
-# Construct a matrix filled zeros and of dtype long:
+tnsr_from_data = torch.tensor(data)
```
soumith, line 32: the use of the variable name `tnsr` all over the tutorial is just weird. Either call it `tensor` or call it `x`

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,189 @@
...
 
-x = torch.tensor([5.5, 3])
-print(x)
+new_ones_tnsr = torch.ones_like(tnsr_from_data) # 2 x 2 matrix of ones
 
-###############################################################
-# or create a tensor based on an existing tensor. These methods
-# will reuse properties of the input tensor, e.g. dtype, unless
-# new values are provided by user
+try:
```
soumith, line 42: if you are trying to make a point that typecasting will throw a runtime error, please do so in a Note.

Actually showcasing that behavior is not needed, and the try/catch with a verbose RuntimeError logic is more intimidating than introducing.

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,189 @@
...
-y = torch.rand(5, 3)
-print(x + y)
 
-###############################################################
-# Addition: syntax 2
+shape = (2,3,)
+rand_tnsr = torch.rand(shape)
+ones_tnsr = torch.ones(shape)
+zeros_tnsr = torch.zeros(shape)
+print(f"Random Tensor:\n{rand_tnsr} \n\n \
```
soumith, line 66: three separate print statments, one in each line are cleaner and better.

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,189 @@
...
 
 ###############################################################
+# **Standard numpy-like indexing and slicing:**
+
+tnsr = torch.ones(4, 4)
+tnsr[:,1] = 0
+print(tnsr)
+
+######################################################################
+# **Both of these are joining ops, but they are subtly different.**
```
soumith, line 127: Introduce one joining op in words, and in code, provide a reference to the other. The tutorial needs to introduce just-enough to a new user.

-------

You can join two tensors with `torch.cat`, which concatenates them in a given dimension.

```
t1 = torch.cat([tnsr, tnsr], dim=1)
print(t1)
```

Also see `torch.stack`, another subtly different tensor joining operation.

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,189 @@
...
+
+t1 = torch.cat([tnsr, tnsr], dim=1)
+t2 = torch.stack([tnsr, tnsr], dim=1)
+print(t1)
+print()
+print(t2)
+
+######################################################################
+# **Multiply op - multiple syntaxes**
```
soumith, line 136: Again here, it's important to first introduce one syntax, and then tell the user that there's another syntax.
It's also really important to separate out the code for each into each line. `tnsr * tnsr == tnsr.mul(tnsr)` tries to be too clever, while intimidating a new user who has their guard up while doing the first tutorial of a new package.

------

Multiplying Tensors

 this computes an element-wise product of two tensors
print(tensor * tensor)

this also computes an element-wise product of two tensors, and is an alternative syntax
print(tnsr.mul(tnsr))

this computes a matrix product, often also called matrix multiplication
tnsr @ tnsr.T

this can also be written as 
tnsr.matmul(tnsr.T)

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,189 @@
...
+print(tnsr * tnsr == tnsr.mul(tnsr))
+
+# both ops compute matrix product
+print(tnsr @ tnsr.T == tnsr.matmul(tnsr.T))
+
+######################################################################
+# .. note::
+#     Operations that have a '_' suffix are in-place. For example:
+#     ``x.copy_(y)``, ``x.t_()``, will change ``x``. In-place operations
+#     don't work well with Autograd, and their use is discouraged.
```
soumith, line 148: In-place operations save some memory, but they don't always work to compute derivative formulas because of immediate loss-of-history. Hence, their use is discouraged.

beginner_source/blitz/tensor_tutorial.py
```diff
@@ -1,195 +1,189 @@
...
+
+print(tnsr)
+tnsr.t_()
```
soumith, line 151: it's much better to show something like `tnsr.add_(1)`, much more easy to recognize what happened

beginner_source/deep_learning_60min_blitz.rst
```diff
@@ -8,13 +8,18 @@ Deep Learning with PyTorch: A 60 Minute Blitz
      <iframe width="560" height="315" src="https://www.youtube.com/embed/u7x8RXwLKcA" frameborder="0" allow="accelerometer; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
    </div>
 
-Goal of this tutorial:
+What is PyTorch?
+~~~~~~~~~~~~~~~~~~~~~
+PyTorch is a Python-based scientific computing package for two broad purposes:
+
+-  A replacement for NumPy to use the power of GPUs.
```
soumith, line 15: the power of GPUs. -> the power of GPUs and other accelerators.

beginner_source/deep_learning_60min_blitz.rst
```diff
@@ -8,13 +8,18 @@ Deep Learning with PyTorch: A 60 Minute Blitz
...
+-  A deep learning research platform that provides maximum flexibility and speed
```
soumith, line 16: An automatic differentiation library that is often useful to implement neural networks

beginner_source/blitz/autograd_tutorial.py
```diff
@@ -1,198 +1,276 @@
...
-Autograd: Automatic Differentiation
-===================================
-
-Central to all neural networks in PyTorch is the ``autograd`` package.
-Let’s first briefly visit this, and we will then go to training our
-first neural network.
+A Gentle Introduction to Autograd
+---------------------------------
+
+Autograd is PyTorch’s automatic differentiation engine that powers
```
soumith, line 6: Fix: The `torch.autograd` package is PyTorch's automatic differentiation engine

Context for Suraj: this is the real autograd package: https://github.com/HIPS/autograd , we used the name, somewhat accidentally. So, it's important to introduce `torch.autograd` or as the `autograd` pacakge, not just Autograd. Please adjust references in this tutorial to Autograd as needed after this context.

beginner_source/blitz/autograd_tutorial.py
```diff
@@ -1,198 +1,276 @@
...
+This collects the gradients for each parameter in the model.
+"""
 
+import torch, torchvision
 
-The ``autograd`` package provides automatic differentiation for all operations
-on Tensors. It is a define-by-run framework, which means that your backprop is
-defined by how your code is run, and that every single iteration can be
-different.
+model = torchvision.models.resnet18(pretrained=True)
```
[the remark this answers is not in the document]
soumith: I agree with Randall here

beginner_source/blitz/autograd_tutorial.py
```diff
@@ -1,198 +1,276 @@
...
+
+######################################################################
+# ``a`` and ``b`` can be viewed as parameters of an NN, with ``Q``
+# analogous to the error. In training we want gradients of the error
+# w.r.t. parameters, i.e.
+#
+# .. math::
+#
+#
+#    \frac{\partial Q}{\partial a} = 9a^2
```
soumith, line 90: first list the original equation, for clarity, i.e. `Q = 3a^3 - b^2`

beginner_source/blitz/autograd_tutorial.py
```diff
@@ -1,198 +1,276 @@
...
+
+from torch import nn, optim
+
+model = torchvision.models.resnet18(pretrained=True)
+
+for param in model.parameters():
+    param.requires_grad = False
+# Replace the last fully-connected layer
+# Parameters of nn.Module instances have requires_grad=True by default
+model.fc = nn.Linear(512, 100)
```
soumith, line 256: after this line, it's worth stating what happened

```
Now, all parameters in the model, except the parameters of model.fc do not compute gradients. All these parameters that don't compute gradients are usually called frozen parameters.

The only parameters that compute gradients are the weights and bias of model.fc. model.fc being the last layer of this network is often called as the classifier.
```

soumith (changes requested):
see comments inline

beginner_source/blitz/autograd_tutorial.py
suraj813: Added an image of the DAG. Would you say the description is accurate and informative? cc: @soumith

PR #1249, opened 2020-11-18
Profiler tutorial

suraj813:


suraj813:
@ilia-cher pending your clarification regarding the "overhead" from `record_function` I'll update the tutorial. For now I have commented `record_function` and added a placeholder disclaimer

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,205 @@
+"""
+Profiling your PyTorch Module
+------------
+
+PyTorch includes a profiler API that is useful to identify the time and
+memory costs of various PyTorch operations in your code. It records the
+events of functions executed in C++, and exposes them to Python.
```
ilia-cher, line 7: it's not only C++, and we also have C++ version of profiler api, so let's remove this sentence.

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,205 @@
...
+Profiler can be easily integrated in your code, and the results are
+printed in a pretty table or as a JSON trace file.
```
ilia-cher, line 9: nit: remove 'pretty'

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,205 @@
...
+# - a linear transformation on the input, and
+# - use the transformation result to get indices on a mask tensor.
+#
+# We wrap the code for each sub-task in separate labelled context managers using
+# ``profiler.record_function("label")``. In the profiler output, the
+# aggregate performance metrics of all operations in the sub-task will
+# show up under its corresponding label.
+#
+#
+# Note that this incurs some overhead, and is best used only while
```
ilia-cher, line 47: i wouldn't single out `record_function` as a source of overhead, it's more complex (and depends on both RecordFunction and profiler)

among things that are known though to increase profiler overhead is, e.g. `with_stack=True` option

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,205 @@
...
+# investigating the trace. Remove it if you are benchmarking runtimes.
+#
+
+class CustomLinear(torch.nn.Linear):
+    def __init__(self, in_features: int, out_features: int, bias: bool = True) -> None:
+        super().__init__(in_features, out_features, bias)
+
+    def forward(self, input, mask):
+        # with profiler.record_function("LINEAR PASS"):
```
ilia-cher, line 56: let's uncomment this and below, a few user-defined `record_function`(s) should be generally ok

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,205 @@
...
+with profiler.profile(with_stack=True, profile_memory=True) as prof:
+    out, idx = model(input, mask)
+
+
+######################################################################
+# Pretty print profiler results
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Finally, we print the profiler results. ``profiler.key_averages``
+# aggregates the results either by input shapes or stack trace events.
```
ilia-cher, line 92: or both shapes and stack; it also groups by operator name (always)

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,205 @@
...
+# Grouping by input shapes is useful to identify which tensor shapes
+# contribute to runtime.
```
ilia-cher, line 94: "are utilized by the model" is prob. a bit more accurate

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,205 @@
...
+# Here, we use ``group_by_stack_n=5`` which aggregates runtimes by the
+# operation and its traceback (truncated to the most recent 5 events), and
+# display the events in the order they are registered. The table can also
+# be sorted by passing a ``sort_by`` argument (refer to the
+# `docs <https://pytorch.org/docs/stable/autograd.html#profiler>`__ for
+# valid sorting keys).
+#
+
+print(prof.key_averages(group_by_stack_n=5).table())
+
```
ilia-cher, line 105: we should also include the output of the profiler

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,205 @@
...
+# operation consumes 953.67 Mb. ``mask`` is initialized with a
+# ``torch.double`` datatype. Can we reduce the memory footprint by casting
+# it to ``torch.float``?
+#
+
+model = CustomLinear(500, 10).cuda()
+input = torch.rand(128, 500).cuda()
+mask = torch.rand((500, 500, 500), dtype=torch.float).cuda()
+
+model(input, mask) #warm up
```
ilia-cher, line 122: nit: double space before #, single space after

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,205 @@
...
+#
+
+
+######################################################################
+# Caveat
+# ^^^^^^^^^^^^^^
+#
+# You might notice introducing ``torch.nonzero`` in our module has shown a
+# new memory footprint of 2.59 Gb that wasn’t in the stacktrace when we
+# used ``np.argwhere``. This is because Profiler tracks only operations in
```
ilia-cher, line 182: let me check what's going on here;

also, more precisely, profiler tracks PyTorch tensor memory consumption, and does not track numpy tensors, but we plan to have a better memory tracking

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,205 @@
...
+#
+# Thread local
+# ^^^^^^^^^^^^^^
+#
+# Profiler runs in the same thread as the operation but it will also
+# profile child operators that might run in another thread.
+# Concurrently-running profilers will be scoped to their own thread to
+# prevent mixing of results.
+#
+# Module-level, not low-level system profiling
```
ilia-cher, line 200: module -> model

tbh, we're actively working on system metrics, so maybe let's not focus too much on that or mention that extra events from the system are being added

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,205 @@
...
+# ^^^^^^^^^^^^^^
+#
+# PyTorch profiler only reports runtime of PyTorch functions, and not at
+# the lower-level system operations.
+#
```
ilia-cher, line 205: we need some conclusion, e.g. "we considered an example and showed how to use profiler to improve time and memory consumption of the model. For more examples, please check profiler recipe/profiler [links]"

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,205 @@
...
+# That worked, the memory footprint has halved - but while the time
+# consumed has reduced a bit (previously 283ms), it’s still too high at
+# 194ms. Turns out copying a matrix from CUDA to CPU is pretty expensive!
+# The ``aten::copy_`` operator in ``forward (11)`` copies ``mask`` to CPU
+# so that it can use the NumPy ``argwhere`` function. We could eliminate
+# the need to copy into a NumPy array if we use a ``torch`` function
+# ``nonzero()`` here instead.
+#
+
+class CustomLinearLite(torch.nn.Linear):
```
ilia-cher, line 142: let's maybe, to avoid confusion, call it `MyModule` or smth similar, so it's clear that this is a custom user model code
further, we could also explicitly create a linear module inside the ctor and not inherit from Linear

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,205 @@
...
+    def __init__(self, in_features: int, out_features: int, bias: bool = True) -> None:
+        super().__init__(in_features, out_features, bias)
+
+    def forward(self, input, mask):
+        # with profiler.record_function("LINEAR PASS"):
+            out = super().forward(input)
+
+        # with profiler.record_function("MASK INDICES"):
+            threshold = out.sum(axis=1).mean().item()
+            hi_idx = np.argwhere(mask.cpu().numpy() > threshold) # fix: use torch.nonzero()
```
ilia-cher, line 61: should we move it back to gpu?

ilia-cher (commented):
Thanks! Left some comments.
Also cc. @ngimel

beginner_source/profiler_tutorial.py
suraj813: Can we add a general note about overheads then, instead of singling out `record_function`?

beginner_source/profiler_tutorial.py
suraj813: The output doesn't show up in the netlify preview, but will build and show when deployed

beginner_source/profiler_tutorial.py
suraj813: Gotcha, I'll update the note to reflect this.

beginner_source/profiler_tutorial.py
suraj813: Okay, I'll remove this for now as we are specifying earlier that currently only torch.Tensors are being tracked

beginner_source/profiler_tutorial.py
suraj813: i think it's not necessary because moving to cpu alone is pretty expensive and showcases inefficiencies of using numpy where torch functions could have worked

beginner_source/profiler_tutorial.py
suraj813: agreed

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,203 @@
...
+# Profile the forward pass
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# We initialize random input and mask tensors, and the ``CustomLinear`` module.
+#
+# Before we run the profiler, we warm-up CUDA to ensure accurate
+# performance benchmarking. We wrap the forward pass of our module in the
+# ``profiler.profile`` context manager. The ``with_stack=True`` parameter appends the
+# file and line number of the operation in the trace.
+#
```
ilia-cher, line 76: pls add a warning that `with_stack=True` incurs an additional overhead

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,203 @@
...
+        super(MyModule, self).__init__()
+        self.linear = nn.Linear(in_features, out_features, bias)
+
+    def forward(self, input, mask):
+        with profiler.record_function("LINEAR PASS"):
+            out = self.linear(input)
+
+        with profiler.record_function("MASK INDICES"):
+            threshold = out.sum(axis=1).mean().item()
+            hi_idx = np.argwhere(mask.cpu().numpy() > threshold)
```
ilia-cher, line 61: I think we probably want to also move hi_idx back to gpu, if the tensor was on gpu

ilia-cher:
Looks good, thanks!
Let me take a look at what happens with memory ('caveat' section) and then it'll be good to go

beginner_source/profiler_tutorial.py
ilia-cher: and return pytorch tensor not a numpy array

beginner_source/profiler_tutorial.py
```diff
@@ -0,0 +1,203 @@
...
+#
+# Here, we use ``group_by_stack_n=5`` which aggregates runtimes by the
+# operation and its traceback (truncated to the most recent 5 events), and
+# display the events in the order they are registered. The table can also
+# be sorted by passing a ``sort_by`` argument (refer to the
+# `docs <https://pytorch.org/docs/stable/autograd.html#profiler>`__ for
+# valid sorting keys).
+#
+
+print(prof.key_averages(group_by_stack_n=5).table())
```
ilia-cher, line 107: when outputting stats here and in other places - let's use `table(sort_by=<metric>, row_limit=X)`

where X - is up to you, just to comfortably see the interesting parts, -1 means 'all rows'
and metric - for memory it is "cpu_memory_usage" and "cuda_memory_usage"

ilia-cher (approved):
thanks for adding the tutorial!

PR #1223, opened 2020-11-07
Fix name error in audio tutorial

vincentqb:
Follow up to #1204

- [x] Rename tutorial as mentioned in #1222
- [x] There is an error in the multiline string.
- [ ] ~~Make notebook run outside of google colab. #1252~~
- [x] Fix tqdm import

cc @brianjo

PR #1252, opened 2020-11-19
non-colab recording in speech tutorial

vincentqb:
Follow up to #1223 to add non-colab recording.


intermediate_source/speech_command_recognition_with_torchaudio_tutorial.py
```diff
@@ -525,6 +525,32 @@ def record(seconds=1):
     return torchaudio.load(filename)
 
 
+def record_noncolab(seconds=1):
+
+    import sounddevice
```
vincentqb, line 530: sounddevice needs to be installed in the environment for this to work

brianjo:
(Fixed a couple of issues in master, so I merged them into this PR.)

PR #1274, opened 2020-12-09
Add fx graph mode ptq static tutorial

jerryzh168:


prototype_source/fx_graph_mode_ptq_static_tutorial.py
```diff
@@ -0,0 +1,446 @@
...
+===========================================================
+
+**Author**: `Jerry Zhang <https://github.com/jerryzh168>`_
+
+This tutorial introduces the steps to do post training static quantization in graph mode based on
+`torch.fx <https://github.com/pytorch/pytorch/blob/master/torch/fx/__init__.py>`_. 
+The advantage of FX graph mode quantization is that as long as the model can be symbolically traced, 
+we can perform quantization fully automatically on the model. 
+
+We have a separate tutorial for FX Graph Mode Post Training Dynamic Quantization(todo).
```
jlin27, line 12: Will this todo be replaced by a link before this PR is merged?

prototype_source/fx_graph_mode_ptq_static_tutorial.py
```diff
@@ -0,0 +1,446 @@
...
+
+We have a separate tutorial for FX Graph Mode Post Training Dynamic Quantization(todo).	
+
+tldr; The FX Graph Mode API looks like the following:
+
+.. code:: python
+
+    import torch
+    from torch.quantization import get_default_qconfig
+    # Note that this is temporary, we'll expose these functions to torch.quantization after official releasee
```
jlin27, line 20: The wording "after official release" is ambiguous here.  Suggestion to rephrase - "Note that this is temporary and we will expose these functions to torch.quantization after this feature is no longer in prototype and moved to stable"

prototype_source/fx_graph_mode_ptq_static_tutorial.py
```diff
@@ -0,0 +1,446 @@
...
+    model.eval()
+    with torch.no_grad():
+        for image, target in data_loader:
+            model(image)
+calibrate(prepared_model, data_loader_test)  # run calibration on sample data
+
+#####################################################################
+# 7. Convert the Model to a Quantized Model
+# -----------------------------------------
+# ``convert_fx`` takes a calibrated model and prodcuces a quantized model.
```
jlin27, line 335: Spelling: prodcuces -> produces

jlin27 (changes requested):

prototype_source/fx_graph_mode_ptq_static_tutorial.py
jerryzh168: this will be added later, I'll have another PR to add tutorial for dynamic quantization

prototype_source/fx_graph_mode_ptq_static_tutorial.py
jerryzh168: oh, this will change before the prototype release. I'll have a PR later to fix this

brianjo: I'm guessing this isn't implemented yet. If you would like this to merge without building the file, you can rename the tutorial, removing _tutorial from the name. The tutorial will be available and you can even open it in colab, but the page will not be rendered with output during the build.

jerryzh168: this is implemented, it should work in master. is it possible for this to be build with master?

brianjo: We built tutorials with stable, so if it's not there it won't build. Should it work in 1.7.1 now?

jerryzh168: I see, it probably won't work in 1.7.1, then do people publish tutorials after the stable release?
I think we can try to merge without building the file then. I'll remove _tutorial from the file.

brianjo: Yes then we can publish the recipe to the regular location instead of prototype. :)

jerryzh168: oh, prototype means the feature introduced by the tutorial is a prototype feature. v.s. beta/stable. The tutorial will be marked as prototype in 1.8, I'm not sure if we can publish it in regular location or not.

Coderx7:
Hi, sorry to bring this up here but what's the difference between this and [this](https://pytorch.org/tutorials/prototype/graph_mode_static_quantization_tutorial.html#prototype-graph-mode-post-training-static-quantization-in-pytorch) 
they are both doing graph mode post-training quantization and nearly everything seems identical except the API!

jerryzh168:
> Hi, sorry to bring this up here but what's the difference between this and [this](https://pytorch.org/tutorials/prototype/graph_mode_static_quantization_tutorial.html#prototype-graph-mode-post-training-static-quantization-in-pytorch)
> they are both doing graph mode post-training quantization and nearly everything seems identical except the API!

yeah, we changed the implementation and also the API for graph mode quantization, the [old one] (https://pytorch.org/tutorials/prototype/graph_mode_static_quantization_tutorial.html#prototype-graph-mode-post-training-static-quantization-in-pytorch) is based on TorchScript, and the new one is based on `torch.fx`, going forward we'll be using FX Graph Mode Quantization. the old one (based on TorchScript) will be deprecated.

PR #1278, opened 2020-12-11
Update transformer_tutorial.py

robz:
`cur_loss` and ms/batch should be calculated with `batch` (the # of batches processed so far), not `log_interval` (the interval of printing out learning statistics)

### DIFF
--- a/beginner_source/transformer_tutorial.py
+++ b/beginner_source/transformer_tutorial.py
@@ -274,13 +274,13 @@ def train():
         total_loss += loss.item()
         log_interval = 200
         if batch % log_interval == 0 and batch > 0:
-            cur_loss = total_loss / log_interval
+            cur_loss = total_loss / batch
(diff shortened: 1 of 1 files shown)


zhangguanheng66:
we used `log_interval` here to count the batch intervals. By dividing the `log_interval` value, we make sure that the 'loss' happens within the same interval.

robz:
> we used `log_interval` here to count the batch intervals. By dividing the `log_interval` value, we make sure that the 'loss' happens within the same interval.

@zhangguanheng66 I'm not following you. log_interval is a fixed value, but total_loss increases every iteration. So total_loss/log_interval would increase at every iteration. Likewise, for “ms/batch” to be the time taken per batch, it should be the time elapsed divided by the # of batches processed so far, not some fixed value.

robz:
Oh I see, total_loss and start_time are reset every time the data is logged. Abandoning.

PR #1295, opened 2020-12-25
Correct neural_networks_tutorial.py

Lucien-cs:


brianjo:
I think this one is OK without the comma.

PR #1306, opened 2021-01-11
Registering all the parameters to the optimizer

fgiobergia:
The tutorial states the following:
> Notice **although we register all the parameters in the optimizer**, the only parameters that are computing gradients (and hence updated in gradient descent) are the weights and bias of the classifier.

However, at the moment, only the classifier (`model.fc`)'s parameters are registered in the optimizer. With this (minor) fix, the code should be in agreement with the text.


holly1238:
Closing this PR. The latest version of the tutorial is already updated with this fix: https://pytorch.org/tutorials/beginner/blitz/autograd_tutorial.html#sphx-glr-beginner-blitz-autograd-tutorial-py

PR #1314, opened 2021-01-21
Add tutorial for extending dispatcher for new backend.

ailzhang:


advanced_source/dispatcher.rst
```diff
@@ -32,13 +32,16 @@ Defining schema and backend implementations
 
 The general principle behind the dispatcher is that it divides the
 implementation of an operator into multiple kernels, each of which implements
-functionality for a specific *dispatch key*; for example, CPU, CUDA or Autograd.
+functionality for a specific *dispatch key*; for example, CPU, CUDA, AutogradCPU
+or AutogradCUDA.  Starting from 1.8 release the dispatcher allows separate registration
+to Autograd backend dispatch keys, and introduced the notion of alias dispatch keys
```
ezyang, line 37: nit: comma splice

advanced_source/dispatcher.rst
```diff
@@ -179,6 +182,17 @@ functions:
...
+Here in the example we register the kernel to ``Autograd``, which is an alias
+dispatch key that propagates the kernel to all Autograd backend keys, e.g. ``AutogradCPU``,
+``AutogradCUDA``, ``AutogradXLA``, ``AutogradOther`` etc.  The propagation happens
+in dispatch table precomputation when the process starts up.  Optionally you
+can register optimized autograd kernel for a backend (e.g. CPU) to its corresponding
+Autograd dispatch key (e.g. AutogradCPU) as well.  When registration to ``AutogradCPU``
+and ``Autograd`` both exist, the direct registration to ``AutogradCPU`` takes precedence.
+To better understand the precompute logic and precedence among all dispatch keys, you can use
+``PythonDispatcher`` provided in `tools/python_dispatcher.py <https://github.com/pytorch/pytorch/blob/master/tools/python_dispacher.py`_.
+For a certain operator, it shows how the computed dispatch table looks like after your registrations.
```
ezyang, line 194: There is a sort of pedagogical question here, which is whether or not to pretend Autograd is a normal dispatch key in the initial explanation, and then only later explain alias keys as a refinement to the model, or attempt to talk about alias keys immediately from the getgo. Personally, I think it makes more sense to not mention alias keys until later (or maybe even not at all in this document), because the average user probably only needs cookbook style understanding of dispatch keys (if I want to do X, then I want key Y) and wouldn't benefit all that much from understanding the underlying model (in part because they don't actually have the ability to affect it, e.g., define new alias keys).

advanced_source/extend_dispatcher.rst
```diff
@@ -0,0 +1,295 @@
...
+to enable the optimization.  You are welcome to send PRs to ``pytorch/pytorch`` to
+upstream those CPU/CUDA kernels directly.  Adding a new backend is only applicable for cases that
+the way you write kernels are fundamentally different, for example:
+
+* supporting a new hardware like Google TPU and customized chips, which often requires using
+  hardware-specific API to write kernels.
+* supporting operators with a different Tensor layout/representation like sparse and quantized, which
+  enforces your kernels to be written in a way that's more efficient given the layout/representation
+  limitation.
+
```
ezyang, line 31: I'd probably want a pointer to how to setup build system for C++ extension to PyTorch somewhere. Fortunately we have preexisting infrastructure so you don't have to redux it here.

advanced_source/extend_dispatcher.rst
```diff
@@ -0,0 +1,295 @@
...
+
+Get a dispatch key for your backend
+-----------------------------------
+
+Pytorch operators are implemented in C++ and made available in Python frontend through a thin wrapper.
```
ezyang, line 36: "thin" wrapper? :P

advanced_source/dispatcher.rst
```diff
@@ -32,13 +32,16 @@ Defining schema and backend implementations
...
+or AutogradCUDA.  Starting from 1.8 release the dispatcher allows separate registration
+to Autograd backend dispatch keys, and introduced the notion of alias dispatch keys
+for `dispatch table precomputation <https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/README.md#dispatch>`_.
 The dispatcher determines what the highest priority dispatch key is at the time
 you call an operator (this is done by looking at both the tensor arguments as
 well as some thread local state), and transfers control to the kernel for that
 dispatch key.  The end effect is that when you call an operator, we first
-execute the Autograd kernel, and then we redispatch to the CPU or CUDA kernel
-depending on the device types of the passed in tensors.
+check the device type of the passed in tensors and execute the corresponding Autograd
```
bhosmer, line 43: I think this last sentence raises more questions than it answers for the careful reader:
- it says we redispatch only to CPU or CUDA - easily fixed
- it says we always execute an Autograd kernel - accurate for now, but confusing unless you're pretty familiar with PT inside baseball (why would we call this kernel during inference?). 
- it doesn't mention any of the other dispatch keys "above" autograd, which is going to be confusing if you *are* familiar with PT inside baseball

Rather than overstuff this first part of the tutorial, I'd suggest just cutting the last sentence. That leaves them with the "determines the highest priority dispatch key" mental picture, which is accurate and clear. But if we definitely want to mention Autograd here, I don't think we can get away without some clarification per above.

advanced_source/dispatcher.rst
bhosmer: s/introduced/introduces/

advanced_source/dispatcher.rst
```diff
@@ -32,13 +32,16 @@ Defining schema and backend implementations
 
 The general principle behind the dispatcher is that it divides the
 implementation of an operator into multiple kernels, each of which implements
-functionality for a specific *dispatch key*; for example, CPU, CUDA or Autograd.
+functionality for a specific *dispatch key*; for example, CPU, CUDA, AutogradCPU
+or AutogradCUDA.  Starting from 1.8 release the dispatcher allows separate registration
+to Autograd backend dispatch keys, and introduced the notion of alias dispatch keys
+for `dispatch table precomputation <https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/README.md#dispatch>`_.
 The dispatcher determines what the highest priority dispatch key is at the time
```
bhosmer, line 39: I'd start a new paragraph with "The dispatcher determines..."

advanced_source/dispatcher.rst
```diff
@@ -114,8 +117,8 @@ At this point, we have an operator with both CPU and CUDA implementations.  How
 can we add autograd support to it?  As you might guess, we will register an
 autograd kernel (similar to what's described in the `custom autograd function <cpp_autograd>`_ tutorial)!
 However, there is a twist: unlike the CPU and CUDA kernels, the autograd kernel
-needs to *redispatch*: it needs to call back into the dispatcher to get to
-the final CPU and CUDA implementations.
+usually has *redispatch*: it does proper backward node setup in autograd kernel,
+then calls back into the dispatcher to get to the CPU/CUDA inference implementations.
```
bhosmer, line 121: The only thing about saying this here is that it doesn't actually match up to anything we see in the code sample below (because it's hidden away by `torch::autograd::Function`). I think it might be better to leave it as before (but following your lead, I'd change "get to the final CPU or CUDA implementations" to something like "get to the inference kernels, e.g. CPU or CUDA implementations" or similar).

advanced_source/dispatcher.rst
bhosmer: I'd agree that we don't need to point the full firehose at them here, but I think we miss an opportunity by not mentioning it at all - first because we're already opening the door (rightly) with "corresponding Autograd key" above, and second because if they wind up using PythonDispatcher they're going to be better equipped to understand what's going on if we've primed them at least a little here. 

What about a sidebar that's a bit more abbreviated, something like

```
.. note::
    
    In this example we register the kernel to ``Autograd``, which installs it as the
    autograd kernel for all backends. You can also register optimized kernels for specific
    backends by using the corresponding backend-specific dispatch key - for example, 
    ``AutogradCPU`` or ``AutogradCUDA``. To explore these and other dispatch key 
    options in more detail, check out the ``PythonDispatcher`` tool provided in 
    `tools/python_dispatcher.py <https://github.com/pytorch/pytorch/blob/master/tools/python_dispacher.py`_.

```

Note: there's a typo in the link to python_dispatcher.py but I don't know how to fix it 😬

bhosmer (commented):
Hey, posting a first round of comments on the edits in `dispatcher.rst`, more on the rest shortly 😁

advanced_source/extend_dispatcher.rst
```diff
@@ -0,0 +1,295 @@
...
+In general PyTorch tensor extensions live below PyTorch’s Autograd layer, which means PyTorch handles
+backward graph setup before dispatching to device kernels.  Because gradient formulas are often purely mathematical
+and thus are general for all backends, PyTorch often registers a kernel to alias dispatch key Autograd,
+which means it can be used by all backends.
+
+In some cases, PyTorch backward kernel implementations are also device specific so that they can squeeze out
+max performance out of each backend. For those operators you’ll see op_backward showing up in
+``RegistrationDeclarations.h`` as *required registration* as well.
+
+In other words, for most operators you don't have to worry about their derivative formulas,
```
bhosmer, line 141: I'd move this paragraph and the code sample that follows up above the "In some cases", and change the "In other words, for most operators..." to "For these operators...". That way you're describing the simplest case first, from start to finish.

Then you can follow with "In some cases..." and I think I'd add a similarly simple code sample for a backwards implementation and registration there. (The registration is important to clarify which dispatch key they should be using to register `<op>_backward`)

advanced_source/extend_dispatcher.rst
```diff
@@ -0,0 +1,295 @@
...
+With this trick you have full control over both training and inference behavior for ``my_add`` operator in your backend.
+Here's `an example <https://github.com/pytorch/xla/blob/master/torch_xla/csrc/aten_autograd_ops.h>`_ in the ``pytorch/xla`` repository.
+
+
+Custom operator support
+-----------------------
+
+Your new backend should work seamlessly with
+`customized operators extended in python <https://pytorch.org/docs/stable/notes/extending.html>`_
+without writing any new kernels as long as the customized operator is composable of existing
```
bhosmer, line 200: nit: s/composable/composed/

advanced_source/extend_dispatcher.rst
```diff
@@ -0,0 +1,295 @@
...
+PyTorch operators (which are already supported by your backend).
+
+For `custom operators extended in C++ <cpp_autograd>`_ they often come with a
+`backend specific C++ kernel implementation e.g. nms kernel in torchvsion <https://github.com/pytorch/vision/blob/master/torchvision/csrc/ops/cuda/nms_kernel.cu>`_
+as well as `a customized Python API e.g. torch.ops.torchvision.nms <https://github.com/pytorch/vision/blob/master/torchvision/csrc/ops/nms.cpp#L18>`_.
+To support these operators, backend extenders will need to write a C++ kernel for your backend and properly
+register them to the corresponding namespace in the dispatcher similar to supporting PyTorch native operators.
```
bhosmer, line 207: nit: s/them/it/

advanced_source/extend_dispatcher.rst
```diff
@@ -0,0 +1,295 @@
...
+Alternatively you could also add a customized API in your extension e.g ``torch_xla.core.functions.nms`` for
+these adhoc requests.
+
+JIT support
+-----------
+
+As we mentioned in `Registering a Dispatched Operator in C++ <dispatcher>`_, kernels registered through `m.impl()` API
+supports being called in both unboxed and boxed ways. In other words your customized backend can also work with our
```
bhosmer, line 215: nit: s/supports/support/

advanced_source/extend_dispatcher.rst
```diff
@@ -0,0 +1,295 @@
...
+JIT tracing/scripting frontend like native CPU/CUDA backend.  You could potentially also write specialized optimization
```
bhosmer, line 216: I'd change "like native CPU/CUDA backend" to something like "just like the in-tree backends like CPU or CUDA do", just for clarity

bhosmer (commented):
Another round of edits 😁  Probably worth another pass after these edits are in, and I'd also like to get @gchanan's eyeballs on it if we can before pushing the button. But this is gonna be a really excellent resource to have in the public facing docs, nice work!

advanced_source/dispatcher.rst
```diff
@@ -115,7 +125,7 @@ can we add autograd support to it?  As you might guess, we will register an
 autograd kernel (similar to what's described in the `custom autograd function <cpp_autograd>`_ tutorial)!
 However, there is a twist: unlike the CPU and CUDA kernels, the autograd kernel
 needs to *redispatch*: it needs to call back into the dispatcher to get to
-the final CPU and CUDA implementations.
+the inference kernels, e.g. CPU or CUDA implementations
```
bhosmer, line 128: Missing a period :)

advanced_source/dispatcher.rst
bhosmer: Aha, I was actually thinking the shorter note could replace the original paragraph above (it was a way of addressing Ed's concern of not digging into alias key details too early, but giving enough context that they'll know what's going on when they encounter them later). 

But I'd either stick with the original paragraph or use the new note - they're really two versions of the same content.

advanced_source/extend_dispatcher.rst
```diff
@@ -0,0 +1,350 @@
...
+
+.. code-block:: cpp
+
+
+  Tensor my_op2_backward(const Tensor& self, const Tensor& other) {
+    // call your backend-specific APIs to implement my_op2_backward so that
+    // it matches PyTorch's native behavior
+  }
+
+  // Note backward kerenl is still registered to PrivateUse1 instead of AutogradPrivateUse1.
```
bhosmer, line 199: typo: kerenl :P

bhosmer (approved):
This is looking good! One followup and a couple of nits (and with any luck @gchanan will have a chance to eyeball it before we land) but LGTM 😁 

[Edit: note that the nontrivial followup is hidden bc it refers to one of the resolved things - it's in between the two other review comments below]

advanced_source/extend_dispatcher.rst
```diff
@@ -0,0 +1,350 @@
...
+  )
+
+
+See `our C++ extension tutorial <https://pytorch.org/tutorials/advanced/cpp_extension.html#building-with-setuptools>`_
+for more details.
+
+Get a dispatch key for your backend
+-----------------------------------
+
+Pytorch operators are implemented in C++ and made available in Python frontend through pybind11.
```
ezyang, line 68: nit: PyTorch. Also, technically, regular PyTorch operator bindings don't actually use pybind11

advanced_source/extend_dispatcher.rst
```diff
@@ -0,0 +1,350 @@
...
+The PyTorch dispatcher divides the implementation of an operator into multiple kernels, each of which is
+associated with a specific dispatch key.  Supporting a new backend in PyTorch essentially means writing
+a kernel for each PyTorch operator in C++ and then registering them to a dispatch key representing your
+customized backend in the dispatcher.
+
+To make sure dispatcher is aware of your backend and dispatches operators to the kernels you provide, you first
+need to get a dispatch key to represent your backend in the system.  PyTorch provides 3 reserved dispatch keys
```
ezyang, line 75: nit: three

advanced_source/extend_dispatcher.rst
```diff
@@ -0,0 +1,350 @@
...
+  }
+
+Now let's zoom in and what operator requires a kernel from a customized backend and what's
+inside the kernels exactly.
+
+PyTorch currently has more than 1600 operators and it’s still growing.  It’s unrealistic
+for backend extensions to keep up with this speed.  Even for native backends like CPU
+or CUDA, it often requires a lot of work to write dedicated kernels for every new op.
+
+Fortunately, the PyTorch dispatcher allows registration to *alias dispatch keys* which can
```
ezyang, line 143: nit: as a backend implementor, I don't really care about alias dispatch keys. The second sentence is what I care about: some kernels in PyTorch don't need to be implemented, we've already decomposed them into several known operators, so you only need to implement the known operator.

advanced_source/extend_dispatcher.rst
```diff
@@ -0,0 +1,350 @@
...
+  PyTorch's codegen as much as possible.
+* Support a backend fallback kernel to automatic convert inputs to CPU and convert the
+  result back to the customized backend. This will allow "full" operator coverage even
+  though you don't have kernels written for every operator.
+
+
+Stay in touch
+-------------
+
+Please use `PyTorch forum <discuss.pytorch.org>`_ for questions and discussions. If you have
```
ezyang, line 345: I'd suggest dev-discuss these days!

advanced_source/extend_dispatcher.rst
```diff
@@ -0,0 +1,350 @@
...
+any feature requests or bug reports, please `file an issue on github <https://github.com/pytorch/pytorch/issues>`_.
+
+If you're interested in helping in any of the future work items above (e.g adding more ``Math``
+kernels for PyTorch operators in C++), please reach out to us through github or slack!
```
ezyang, line 349: nit: GitHub Slack

ezyang (approved):

advanced_source/extend_dispatcher.rst
```diff
@@ -14,53 +14,23 @@ to `register a dispatched operator in C++ <dispatcher>`_ and how to write a
...
+What's a new backend?
+---------------------
 
-Before adding a new backend, let's first look at your use cases.  If you have better ideas about
-how to make a few operators run faster on CPU/GPU, you don't need create a new backend
-to enable the optimization.  You are welcome to send PRs to ``pytorch/pytorch`` to
-upstream those CPU/CUDA kernels directly.  Adding a new backend is only applicable for cases where
-the way you write kernels are fundamentally different, for example:
+Adding a new backend to PyTorch requires a lot developement and maintainence from backend extenders.
+Before adding a new backend, let's first consider a few common use cases and recommended solutions for them:
```
bhosmer, line 21: Nit: "a lot of development"

advanced_source/extend_dispatcher.rst
```diff
@@ -14,53 +14,23 @@ to `register a dispatched operator in C++ <dispatcher>`_ and how to write a
...
-          'build_ext': Build,  # Build is a derived class of BuildExtension
-      }
-      # more configs...
-  )
-
-
-See `our C++ extension tutorial <https://pytorch.org/tutorials/advanced/cpp_extension.html#building-with-setuptools>`_
-for more details.
+In this tutorial we'll mainly focus on adding a new out-of-tree device below.  Adding out-of-tree support
+for a different tenosr layout might share many common steps with devices, but we haven't seen an example of
```
bhosmer, line 32: Nit: s/tenosr/tensor/

advanced_source/extend_dispatcher.rst
```diff
@@ -71,17 +41,37 @@ associated with a specific dispatch key.  Supporting a new backend in PyTorch es
...
+  TensorImpl(
+      Storage&& storage,
+      DispatchKeySet ks,
+      const caffe2::TypeMeta data_type);
+
+  // To create a TensorImpl on PrivateUse1 backend, pass in the following ks to TensorImpl creation.
+  DispatchKeySet ks = c10::DispatchKeySet{c10::DispatchKey::PrivateUse1, c10::DispatchKey::AutogradPrivateUse1};
+
+
+Note that ``TensorImpl`` class above assumes your Tensor is backed by a storage like CPU/CUDA, we also
```
bhosmer, line 66: Nit: should be two sentences

advanced_source/extend_dispatcher.rst
```diff
@@ -108,9 +98,9 @@ There're multiple fields associated with a single operator. Let's break it down
 
 * ``Tensor & abs_out(Tensor & out, const Tensor & self);`` is the C++ signature of the operator, your C++
   kernel should match this signature exactly.
-* ``aten::abs.out(Tensor self, *, Tensor(a!) out) -> Tensor(a!)`` is the unique JIT schema representing the operator.
-  Since C++ allows overloads but JIT schema is guaranteed to be unique, operator name in JIT schema might be different
-  from the C++ signature to differentiate between multiple overloads of the same operator.
+* ``aten::abs.out(Tensor self, *, Tensor(a!) out) -> Tensor(a!)`` is the unique schema representing the operator,
+  which also contains aliasing and mutation annotations compared to C++ signature.  This is the unique identifier
```
bhosmer, line 102: Nit: "compared to [the] C++ signature"

bhosmer (approved):
Just noticed a few minor nits, looks good!

PR #1338, opened 2021-02-04
Update spaCy model download to full name

malfet:
Short 'en' and 'de' names has been deprecated since spaCy 3.0, so use fullname instead as suggested by the warning:
```
[!] As of spaCy v3.0, shortcuts like 'en' are deprecated. Pleaseuse the full
pipeline package name 'en_core_web_sm' instead.
[!] As of spaCy v3.0, shortcuts like 'de' are deprecated. Pleaseuse the full
pipeline package name 'de_core_news_sm' instead.
```
Also, add `--retry 3` to all curl commands in Windows CI path


brianjo (approved):
Thanks Nikita!

PR #1347, opened 2021-02-05
Tutorial combining DDP with Pipeline Parallelism to Train Transformer models

pritamdamania87:
Summary: Tutorial which places a pipe on GPUs 0 and 1 and another Pipe
on GPUs 2 and 3. Both pipe replicas are replicated via DDP. One process
drives GPUs 0 and 1 and another drives GPUs 2 and 3.


advanced_source/ddp_pipeline_tutorial.py
```diff
@@ -0,0 +1,452 @@
...
+# In 'run_worker'
+    ntokens = len(TEXT.vocab.stoi) # the size of vocabulary
+    emsize = 4096 # embedding dimension
+    nhid = 4096 # the dimension of the feedforward network model in nn.TransformerEncoder
+    nlayers = 8 # the number of nn.TransformerEncoderLayer in nn.TransformerEncoder
+    nhead = 16 # the number of heads in the multiheadattention models
+    dropout = 0.2 # the dropout value
+
+    from torch.distributed import rpc
+    tmpfile = tempfile.NamedTemporaryFile()
```
rohan-varma, line 290: So do both processes initialize an independent instance of the RPC framework (i.e. both processes launched by run_worker have RPC initialized, but they're not interconnected?) Is it possible for only one process to init RPC?

advanced_source/ddp_pipeline_tutorial.py
```diff
@@ -0,0 +1,452 @@
...
+            )
+
+    # Initialize process group and wrap model in DDP.
+    from torch.nn.parallel import DistributedDataParallel
+    import torch.distributed as dist
+    os.environ['MASTER_ADDR'] = 'localhost'
+    os.environ['MASTER_PORT'] = '29500'
+    dist.init_process_group(
+                backend="nccl", rank=rank, world_size=2)
+    model = DistributedDataParallel(model)
```
rohan-varma, line 319: Do we not need to pass in the appropriate `device_ids` list here? ([0,1], and [2,3])?

advanced_source/ddp_pipeline_tutorial.py
```diff
@@ -0,0 +1,452 @@
...
+                torch.nn.Sequential(
+                    TransformerModelStage1(ntokens, emsize, nhead, nhid, int(nlayers/2), dropout).cuda(2 * rank),
+                    TransformerModelStage2(ntokens, emsize, nhead, nhid, int(nlayers/2), dropout).cuda(2 * rank + 1),
+                ),
+                chunks = 8,
+                checkpoint = "never"
+            )
+
+    # Initialize process group and wrap model in DDP.
+    from torch.nn.parallel import DistributedDataParallel
```
mrzzd, line 313: is there a technical reason we import here?

advanced_source/ddp_pipeline_tutorial.py
```diff
@@ -0,0 +1,452 @@
...
+
+
+######################################################################
+# ``get_batch()`` function generates the input and target sequence for
+# the transformer model. It subdivides the source data into chunks of
+# length ``bptt``. For the language modeling task, the model needs the
+# following words as ``Target``. For example, with a ``bptt`` value of 2,
+# we’d get the following two Variables for ``i`` = 0:
+#
+# .. image:: ../_static/img/transformer_input_target.png
```
mrzzd, line 243: how does this line help? github does not show the picture inline

mrzzd (commented):
Thanks Pritam! Looks great. 
Wonder if we it helps to reduce some of modeling complexity (or the comments) in this file, since the main point is about the pipelining, not other aspects. There is extensive description of say loss function or the input generation, while it could refer to other tutorials for those parts.

advanced_source/ddp_pipeline_tutorial.py
```diff
@@ -0,0 +1,452 @@
...
+######################################################################
+# Define the model
+# ----------------
+#
+
+
+######################################################################
+# In this tutorial, we will split a Transformer model across two GPUs and use
+# pipeline parallelism to train the model. In addition to this, we use
+# `Distributed Data Parallel <https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html>`__
```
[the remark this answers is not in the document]
pritamdamania87: Its nicer when viewing this in a browser where you have a clickable link to quickly go to the DDP docs. The `netlify` bot on the PR has a link to view these tutorials in a browser. I'd suggest viewing the tutorials using that. This is the link for reference: https://deploy-preview-1347--pytorch-tutorials-preview.netlify.app/advanced/ddp_pipeline_tutorial.html

advanced_source/ddp_pipeline_tutorial.py
pritamdamania87: You can view the tutorial in a browser: https://deploy-preview-1347--pytorch-tutorials-preview.netlify.app/advanced/ddp_pipeline_tutorial.html.

advanced_source/ddp_pipeline_tutorial.py
pritamdamania87: Both processes need to initialize RPC since both are driving independent pipes. The RPC framework across both don't have to be interconnected since these are two independent pipes. When we do support cross host pipelining, we need RPCs only across a single instance of the pipe.

In this particular example, I didn't create an RPC worker group of world_size 2 since it might be misleading that we're performing RPCs across two processes.

advanced_source/ddp_pipeline_tutorial.py
pritamdamania87: Just closer to the usage of the module.

advanced_source/ddp_pipeline_tutorial.py
pritamdamania87: No, the model is already placed on the appropriate devices. Also, isn't `device_ids` for the case where the module resides on a single device? From our docs:

```
        device_ids (list of int or torch.device): CUDA devices. This should
                   only be provided when the input module resides on a single
                   CUDA device. For single-device modules, the i'th
                   :attr:`module` replica is placed on ``device_ids[i]``. For
                   multi-device modules and CPU modules, ``device_ids`` must be
                   ``None`` or an empty list, and input data for the forward
                   pass must be placed on the correct device. (default: all
                   visible devices for single-device modules)
```

pritamdamania87 (commented):
> Wonder if we it helps to reduce some of modeling complexity (or the comments) in this file, since the main point is about the pipelining, not other aspects. There is extensive description of say loss function or the input generation, while it could refer to other tutorials for those parts.

Agree that there is a lot of repeated stuff in this tutorial from other tutorials. But I feel it is still useful to have those sections to keep the tutorial as standalone as possible.

advanced_source/ddp_pipeline_tutorial.py
```diff
@@ -0,0 +1,455 @@
...
+            best_model = model
+
+        scheduler.step()
+
+
+######################################################################
+# Evaluate the model with the test dataset
+# -------------------------------------
+#
+# Apply the best model to check the result with the test dataset.
```
mrzzd, line 440: Could we find best model across replicas? Say do an all-reduce and print only if the model has best loss.

advanced_source/ddp_pipeline_tutorial.py
```diff
@@ -0,0 +1,455 @@
...
+if torch.cuda.device_count() < 4:
+    print('Need at least four GPU devices for this tutorial')
+    sys.exit(0)
+
+class TransformerModelStage1(nn.Module):
+
+    def __init__(self, ntoken, ninp, nhead, nhid, nlayers, dropout=0.5):
+        super(TransformerModelStage1, self).__init__()
+        self.src_mask = None
+        self.pos_encoder = PositionalEncoding(ninp, dropout)
```
rohan-varma, line 66: In the rendered webpage, this is the first block of code and `PositionalEncoding` is defined in the 2nd block. Shall we define it above this so users can copy and paste individual blocks and the script still runs without errors?

advanced_source/ddp_pipeline_tutorial.py
```diff
@@ -0,0 +1,455 @@
...
+
+        output = self.transformer_encoder(src, self.src_mask)
+        output = self.decoder(output)
+        return output
+
+######################################################################
+# ``PositionalEncoding`` module injects some information about the
+# relative or absolute position of the tokens in the sequence. The
+# positional encodings have the same dimension as the embeddings so that
+# the two can be summed. Here, we use ``sine`` and ``cosine`` functions of
```
rohan-varma, line 129: Should we add a sentence mentioning to go to https://pytorch.org/tutorials/beginner/transformer_tutorial.html for more details?

advanced_source/ddp_pipeline_tutorial.py
```diff
@@ -0,0 +1,455 @@
...
+                checkpoint = "never"
+            )
+
+    # Initialize process group and wrap model in DDP.
+    from torch.nn.parallel import DistributedDataParallel
+    import torch.distributed as dist
+    os.environ['MASTER_ADDR'] = 'localhost'
+    os.environ['MASTER_PORT'] = '29500'
+    dist.init_process_group(
+                backend="nccl", rank=rank, world_size=2)
```
rohan-varma, line 321: nit: since this is in run_worker, pass in `world_size=world_size`?

advanced_source/ddp_pipeline_tutorial.py
```diff
@@ -0,0 +1,455 @@
...
+        world_size=1,
+        rpc_backend_options=rpc.TensorPipeRpcBackendOptions(
+            init_method="file://{}".format(tmpfile.name),
+        )
+    )
+
+    # Need to use 'checkpoint=never' since as of PyTorch 1.8, Pipe checkpointing
+    # doesn't work with DDP.
+    from torch.distributed.pipeline.sync import Pipe
+    model = Pipe(
```
rohan-varma, line 306: Not really related to the tutorial, but as a follow up, might be useful to see the performance win we get by using Pipeline parallel here. I'm assuming this would work if the user just did a reqular nn.Sequential not wrapped with Pipe and manually handled the split across multiple devices, but it would be a lot less performant.

advanced_source/ddp_pipeline_tutorial.py
rohan-varma: As I understand it, they should be equivalent right? With DDP they should start off with the same params and gradients are synced every iteration.

rohan-varma (approved):
LGTM, thanks for adding this tutorial!

advanced_source/ddp_pipeline_tutorial.py
mrzzd: I see, although each of them my come to different val_loss and best_val_loss, but I see at the end it should not matter which one you choose.

mrzzd (approved):
Looks good to me, thanks!

PR #1356, opened 2021-02-11
[FX] Added fuser tutorial

Chillee:
Not sure how to test it in notebook format.

Also, perhaps I'd like to bake in the output somehow? It would be somewhat embarassing if the fused version was slower due to noise :)


intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,229 @@
...
+*******************************************************
+**Author**: `Horace He <https://github.com/chillee>`_
+
+In this tutorial, we are going to use FX to do the following:
+
+1) Find patterns of conv/batch norm.
+2) Fold the batch norm statistics into the convolution weights.
+
+Despite this being a fairly trivial graph rewrite, this has been surprisingly
+difficult to do in PyTorch for quite some time.
```
jamesr66a, line 13: Yeah IDK about this sentence, probably remove it

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,229 @@
+# -*- coding: utf-8 -*-
+"""
+(beta) Building a Conv/BN fuser in FX
+*******************************************************
+**Author**: `Horace He <https://github.com/chillee>`_
+
+In this tutorial, we are going to use FX to do the following:
+
+1) Find patterns of conv/batch norm.
```
jamesr66a, line 9: Might want to emphasize that these are patterns over data dependencies

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,229 @@
...
+import torch
+import torch.nn as nn
+
+######################################################################
+# For this tutorial, we are going to create a model consisting of convolutions
+# and batch norms. Note that this model has some tricky components - some of
+# the modules are hidden within sequentials and one of the modules is wrapped
+# inside of another PyTorch module.
+
+class Wrapper(nn.Module):
```
jamesr66a, line 37: From linear scanning, it's not clear to me what this code example is. Is `M` the model we're testing on? What is `Wrapper` and why is it needed? The lead-in sentence isn't clear on whether we're matching on `Sequential` or otherwise

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,229 @@
...
+# One of the primary challenges with trying to automatically fuse convolution
+# and batch norm in PyTorch is that PyTorch does not provide an easy way of
+# accessing the computational graph. FX resolves that problem.
+
+traced_model = torch.fx.symbolic_trace(model)
+print(traced_model.graph)
+
+######################################################################
+# This gives us a graph representation of our model. Note that both the modules
+# hidden within the sequential as well as the wrapped modue have been inlined
```
jamesr66a, line 80: s/modue/module/

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,229 @@
...
+# -----------------------------------------
+# One of the primary challenges with trying to automatically fuse convolution
+# and batch norm in PyTorch is that PyTorch does not provide an easy way of
+# accessing the computational graph. FX resolves that problem.
+
+traced_model = torch.fx.symbolic_trace(model)
+print(traced_model.graph)
+
+######################################################################
+# This gives us a graph representation of our model. Note that both the modules
```
jamesr66a, line 79: Recommend capitalizing class names like `Module` and `Sequential`

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,229 @@
...
+# hidden within the sequential as well as the wrapped modue have been inlined
+# into the graph. More information can be found at the FX documentation
+# https://pytorch.org/docs/master/fx.html.
```
jamesr66a, line 82: The docs are pretty long, probably better to point out what specific information you're referring to, e.g. link to the symbolic tracing section and mention this is the default level of abstraction that's traced

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+####################################
+# Fusing Convolution with Batch Norm
+# ----------------------------------
+# Unlike some other fusions, fusion of convolution with batch norm does not
+# require any additional kernels. Instead, as batch norm during inference
```
jamesr66a, line 89: Define "kernel". Readers may not know what that is. I recommend saying "operator" instead

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,229 @@
+# -*- coding: utf-8 -*-
+"""
+(beta) Building a Conv/BN fuser in FX
+*******************************************************
+**Author**: `Horace He <https://github.com/chillee>`_
+
+In this tutorial, we are going to use FX to do the following:
+
+1) Find patterns of conv/batch norm.
+2) Fold the batch norm statistics into the convolution weights.
```
jamesr66a, line 10: Probably worth noting at the beginning that this is for inference mode only

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,229 @@
...
+    assert(not (conv.training or bn.training)), "Fusion only for eval!"
+    fused_conv = copy.deepcopy(conv)
+
+    fused_conv.weight, fused_conv.bias = \
+        fuse_conv_bn_weights(fused_conv.weight, fused_conv.bias,
+                             bn.running_mean, bn.running_var, bn.eps, bn.weight, bn.bias)
+
+    return fused_conv
+
+def fuse_conv_bn_weights(conv_w, conv_b, bn_rm, bn_rv, bn_eps, bn_w, bn_b):
```
jamesr66a, line 106: Worth adding documentations (e.g. docstrings) to both of these functions so the reader can get their bearings and know what each parameter is

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,229 @@
...
+    parent_name, name = _parent_name(node.target)
+    setattr(modules[parent_name], name, new_module)
+
+
+def fuse(model: torch.nn.Module) -> torch.nn.Module:
+    model = copy.deepcopy(model)
+    fx_model = fx.symbolic_trace(model) # We symbolically trace our model here
+    modules = dict(fx_model.named_modules())
+
+    for node in fx_model.graph.nodes:
```
jamesr66a, line 148: Go into more detail on this code example. This tutorial will be one of the introductions to FX. We should introduce the basic concepts of FX (GraphModule, Graph, Node, op, target), and explain how each of these are being used in this code listing.

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,229 @@
...
+            fused_conv = fuse_conv_bn_eval(conv, bn)
+            replace_node_module(node.args[0], modules, fused_conv)
+            # As we've folded the batch nor into the conv, we need to replace all uses
+            # of the batch norm with the conv.
+            node.replace_all_uses_with(node.args[0])
+            # Now that all uses of the batch norm have been replaced, we can
+            # safely remove the batch norm.
+            fx_model.graph.erase_node(node)
+    fx_model.graph.lint()
+    fx_model.recompile()
```
jamesr66a, line 165: Definitely document recompilation

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,229 @@
...
+
+fused_model = fuse(model)
+inp = torch.randn(5, 1, 1, 1)
+assert(abs(fused_model(inp).sum() - model(inp).sum()) < 1e-5)
+
+
+# Benchmarking our Fusion on ResNet18
+# ----------
+# We can test our fusion pass on a larger model like ResNet18 and see how much
+# this pass improves inference performance.
```
jamesr66a, line 191: NOt sure if it's intended, but this comment and the ones below are showing up as part of the code listing

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,229 @@
...
+######################################################################
+# Testing out our Fusion Pass
+# -----------------------------------------
+# We can now run this fusion pass on our initial toy model and verify that our
+# results are identical.
+
+
+fused_model = fuse(model)
+inp = torch.randn(5, 1, 1, 1)
+assert(abs(fused_model(inp).sum() - model(inp).sum()) < 1e-5)
```
jamesr66a, line 185: `torch.testing.assert_allclose()`

intermediate_source/fx_conv_bn_fuser.py
Chillee: It's not needed, it's just meant to making the matching a little bit harder. I'll rename `Wrapper` to `WrappedBatchnrom`

index.rst
```diff
@@ -215,6 +215,15 @@ Welcome to PyTorch Tutorials
    :link: advanced/super_resolution_with_onnxruntime.html
    :tags: Production
 
+.. Code Transformations with FX
+.. customcarditem::
+   :header: Building a Convolution/Batch Norm fuser in FX
+   :card_description: Build a simple FX interpreter to record the runtime of op, module, and function calls and report statistics
```
gchanan, line 221: are the card description and images correct?  They look like they belong to other tutorials?

index.rst
```diff
@@ -215,6 +215,15 @@ Welcome to PyTorch Tutorials
    :link: advanced/super_resolution_with_onnxruntime.html
    :tags: Production
 
+.. Code Transformations with FX
+.. customcarditem::
+   :header: Building a Convolution/Batch Norm fuser in FX
```
gchanan, line 220: I've seen this technique more commonly referred to as "folding" but both make sense (https://towardsdatascience.com/speed-up-inference-with-batch-normalization-folding-8a45a83a89d8, https://arxiv.org/abs/1611.09842 calls it "absorbing").

Might be nice to use different terminology in case we want to add a "first class" fusion tutorial later that e.g. directly calls into NNC.

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,231 @@
+# -*- coding: utf-8 -*-
+"""
+(beta) Building a Conv/BN fuser in FX
```
gchanan, line 3: nit: use the full name Convolution in the title (to match what's in the index).

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,231 @@
...
+*******************************************************
+**Author**: `Horace He <https://github.com/chillee>`_
+
+In this tutorial, we are going to use FX to do the following:
+
+1) Find patterns of conv/batch norm in the data dependencies.
+2) Fold the batch norm statistics into the convolution weights.
+
+Note that this optimization only works for models in inference mode.
```
gchanan, line 12: nit: a little more detail here is probably worth it, eg. something like "only works for models in inference model (i.e. `model.eval()`).

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,231 @@
+# -*- coding: utf-8 -*-
+"""
+(beta) Building a Conv/BN fuser in FX
+*******************************************************
+**Author**: `Horace He <https://github.com/chillee>`_
+
+In this tutorial, we are going to use FX to do the following:
+
+1) Find patterns of conv/batch norm in the data dependencies.
+2) Fold the batch norm statistics into the convolution weights.
```
gchanan, line 10: I think it's good to explicitly connect 1) and 2) to make it obvious, e.g. something like
2) For the patterns found in 1), fold the batch norm...

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,231 @@
...
+
+Note that this optimization only works for models in inference mode.
+
+We will be building the fuser that exists here:
+https://github.com/pytorch/pytorch/blob/master/torch/fx/experimental/fuser.py
```
gchanan, line 15: I usually recommend linking to an explicit version (could be the 1.8 branch) because this will go out of date pretty quickly

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,231 @@
+# -*- coding: utf-8 -*-
+"""
+(beta) Building a Conv/BN fuser in FX
+*******************************************************
+**Author**: `Horace He <https://github.com/chillee>`_
+
+In this tutorial, we are going to use FX to do the following:
```
gchanan, line 7: link to some FX documentation or provide a brief description?  This just jumps in the deep end.

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,231 @@
...
+import torch
+import torch.nn as nn
+
+######################################################################
+# For this tutorial, we are going to create a model consisting of convolutions
+# and batch norms. Note that this model has some tricky components - some of
+# the conv/batch norm patterns are hidden within Sequentials and one of the
+# BatchNorms is wrapped in another module.
+
+class WrappedBatchnorm(nn.Module):
```
gchanan, line 36: nit: use the same capitalization as in torch, WrappedBatchNorm.

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,231 @@
...
+model = M()
+
+model.eval()
+
+######################################################################
+# Fusing Convolution with Batch Norm
+# -----------------------------------------
+# One of the primary challenges with trying to automatically fuse convolution
+# and batch norm in PyTorch is that PyTorch does not provide an easy way of
+# accessing the computational graph. FX resolves that problem.
```
gchanan, line 72: I think a little more detail is worth specifying here because this is the meat of the program capture point.  Something like (please wordsmith):

"PyTorch does not provide an easy way of accessing the computational graph hidden within modules.  FX resolves this problem by recording the actual operations called, so we can track the data dependencies even though the calls to the operators may be inside the `forward` of modules, nested within a sequential, or wrapped in a user-defined module"

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,231 @@
...
+# This gives us a graph representation of our model. Note that both the modules
+# hidden within the sequential as well as the wrapped module have been inlined
+# into the graph. More information can be found at the FX documentation
+# https://pytorch.org/docs/master/fx.html.
+
+
+####################################
+# Fusing Convolution with Batch Norm
+# ----------------------------------
+# Unlike some other fusions, fusion of convolution with batch norm does not
```
gchanan, line 87: nice, good explanation.  This is why I suggested using the term "folding" above, to use different terminology for these two cases.

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,231 @@
...
+# require any new operators. Instead, as batch norm during inference
+# consists of a pointwise add and multiply, these operations can be "baked"
+# into the preceding convolution's weights. Read
+# https://nenadmarkus.com/p/fusing-batchnorm-and-conv/ for further details. The
+# code here is copied from
+# https://github.com/pytorch/pytorch/blob/master/torch/nn/utils/fusion.py for
```
gchanan, line 93: nit : as above, link to specific version.

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,231 @@
...
+# https://pytorch.org/docs/master/fx.html.
+
+
+####################################
+# Fusing Convolution with Batch Norm
+# ----------------------------------
+# Unlike some other fusions, fusion of convolution with batch norm does not
+# require any new operators. Instead, as batch norm during inference
+# consists of a pointwise add and multiply, these operations can be "baked"
+# into the preceding convolution's weights. Read
```
gchanan, line 90: probably worth making it explicit why you'd want to do this / why this is faster.

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,231 @@
...
+        model(inp)
+    begin = time.time()
+    for _ in range(iters):
+        model(inp)
+    return str(time.time()-begin)
+
+fused_rn18 = fuse(rn18)
+print("Unfused time: ", benchmark(rn18))
+print("Fused time: ", benchmark(fused_rn18))
+# As FX is a source to source transformation, our transformation can still
```
gchanan, line 214: do you ever display the FX output?  I feel like you could make this point much clearer if you displayed that and said something like "as our output was (Torchscriptable) PyTorch code, we can easily `jit.script` the output to try and increase our performance even more.  In this way, our FX model transformation composes with Torchscript with no issues."

index.rst
Chillee: Description is wrong, but not sure what to put for the image. @jamesr66a is there any reason you chose this image for the performance profiling for FX? https://github.com/pytorch/tutorials/pull/1319/files#diff-54a294a5d016e1a8e98bc95668ed84a99a9edd5c10394d9a2b1ee848006e98a7R223

index.rst
Chillee: I agree the terminology is confusing, but I think fusion is an acceptable (and more widely understood) term. If we add a fusion tutorial later I'd be glad to rename it to something to avoid name conflicts.

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,261 @@
+# -*- coding: utf-8 -*-
+"""
+(beta) Building a Convolution/Batch Norm fuser in FX
+*******************************************************
+**Author**: `Horace He <https://github.com/chillee>`_
+
+In this tutorial, we are going to use FX, a Py to do the following:
```
gchanan, line 7: a Py?

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,261 @@
...
+    begin = time.time()
+    for _ in range(iters):
+        model(inp)
+    return str(time.time()-begin)
+
+fused_rn18 = fuse(rn18)
+print("Unfused time: ", benchmark(rn18))
+print("Fused time: ", benchmark(fused_rn18))
+######################################################################
+# As we previously saw, the output of FX is (Torchscriptable) PyTorch code, we
```
gchanan, line 243: nit: say something like "the output of our FX transformation" -- it currently sounds like the output of every FX transformation is torchscriptable python.

Chillee: Hmm, that's kinda true, if you're outputting a `GraphModule`. But I get the confusion. Updated.

intermediate_source/fx_conv_bn_fuser.py
jamesr66a: This still looks unresolved lol

index.rst
jamesr66a: I put that because it's just a generic PyTorch logo: 
image

And wasn't sure if I should copy that into a new filename or if we should come up with some logo to use

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,261 @@
...
+
+######################################################################
+# Fusing Convolution with Batch Norm
+# -----------------------------------------
+# One of the primary challenges with trying to automatically fuse convolution
+# and batch norm in PyTorch is that PyTorch does not provide an easy way of
+# accessing the computational graph. FX resolves this problem by symbolically
+# tracing the actual operations called, so that we can track the computations
+# through the `forward` call, nested within Sequential modules, or wrapped in
+# an user-defined module.
```
jamesr66a, line 75: nit: a user-defined module

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,261 @@
...
+
+def _parent_name(target : str) -> Tuple[str, str]:
+    """
+    Splits a qualname into parent path and last atom.
+    For example, `foo.bar.baz` -> (`foo.bar`, `baz`)
+    """
+    *parent, name = target.rsplit('.', 1)
+    return parent[0] if parent else '', name
+
+def replace_node_module(node: fx.Node, modules: Dict[str, Any], new_module: torch.nn.Module):
```
jamesr66a, line 145: This could probably use a docstring describing preconditions (i.e. node must be a `call_module`) and what this function does

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,261 @@
...
+            bn = modules[node.target]
+            fused_conv = fuse_conv_bn_eval(conv, bn)
+            replace_node_module(node.args[0], modules, fused_conv)
+            # As we've folded the batch nor into the conv, we need to replace all uses
+            # of the batch norm with the conv.
+            node.replace_all_uses_with(node.args[0])
+            # Now that all uses of the batch norm have been replaced, we can
+            # safely remove the batch norm.
+            fx_model.graph.erase_node(node)
+    fx_model.graph.lint()
```
jamesr66a, line 189: Comment what this is doing (making sure our `Graph` is still valid) and why it's important to call it

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,261 @@
...
+# Testing out our Fusion Pass
+# -----------------------------------------
+# We can now run this fusion pass on our initial toy model and verify that our
+# results are identical. In addition, we can print out the code for our fused
+# model and verify that there are no more batch norms.
+
+
+fused_model = fuse(model)
+print(fused_model.code)
+inp = torch.randn(5, 1, 1, 1)
```
jamesr66a, line 213: Should we run this on a more realistic input shape?

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -0,0 +1,261 @@
...
+import time
+
+rn18 = models.resnet18()
+rn18.eval()
+
+inp = torch.randn(10, 3, 224, 224)
+output = rn18(inp)
+
+def benchmark(model, iters=20):
+    for _ in range(10):
```
jamesr66a, line 232: Comment that these are warmup iterations and why they are important

jamesr66a (commented):
Looks good to me!

intermediate_source/fx_conv_bn_fuser.py
Chillee: whups forgot to push.

intermediate_source/fx_conv_bn_fuser.py
Chillee: :thinking: Shouldn't it be `an` before `user`? Since `user` starts with a vowel?

gchanan: the rule is based on the sound, not the letter.

intermediate_source/fx_conv_bn_fuser.py
Chillee: I just wrote all the conv/batch norm modules to operate on a `[1,1,1]` shape. We're not measuring the performance of this module so I don't think it matters.

jamesr66a (approved):

PR #1360, opened 2021-02-15
Correction of LaTeX math in tutorial

wangkuiyi:
The current render of the document looks like the following

Screen Shot 2021-02-15 at 1 55 56 PM

It seems that rst/Sphinx doesn't support `\frac` in math.  We could verify the lack of support indirectly in some (online) rst/Sphinx previewers like [this one](https://livesphinx.herokuapp.com/), which cannot render `\frac` correctly either.

Hence, this PR uses `3/2` over `\frac{3}{2}`.

It also corrects `pi` by `\pi`.


malfet (approved):

beginner_source/examples_autograd/polynomial_custom_function.py
```diff
@@ -6,7 +6,7 @@
 A third order polynomial, trained to predict :math:`y=\sin(x)` from :math:`-\pi`
 to :math:`\pi` by minimizing squared Euclidean distance. Instead of writing the
 polynomial as :math:`y=a+bx+cx^2+dx^3`, we write the polynomial as
-:math:`y=a+b P_3(c+dx)` where :math:`P_3(x)=\frac{1}{2}\left(5x^3-3x\right)` is
+:math:`y=a+b P_3(c+dx)` where :math:`P_3(x)=\left(5x^3-3x\right)/2` is
```
malfet, line 9:
```suggestion
:math:`y=a+b P_3(c+dx)` where :math:`P_3(x)=\\frac{1}{2}\\left(5x^3-3x\\right)` is
```

malfet (changes requested):
RST does seem to support frac, but backslashes must be correctly escaped

PR #1364, opened 2021-02-18
Fix link to pybind11 docs

fritzo:


brianjo:
Thanks for the fix @fritzo!

PR #1367, opened 2021-02-23
Correct mistake in autograd_tutorial.py

taha-yassine:
For the two equations to be consistent, the LHS of the first one should be replaced by `\vec{v}`. In addition, I moved the scalar function definition `l=g\left(\vec{y}\right)` to the first sentence.

image


suraj813 (approved):
@taha-yassine Thanks for fixing this!

PR #1371, opened 2021-02-23
[Lite Interpreter] Add lite interpreter workflow in Android and iOS

cccclai:
Create a recipe to use lite interpreter workflow in Android and iOS

### DIFF
--- a/prototype_source/lite_interpreter.rst
+++ b/prototype_source/lite_interpreter.rst
@@ -1,0 +1,144 @@
+(Prototype) Introduce lite interpreter workflow in Android and iOS
+
+**Author**: `Chen Lai <https://github.com/cccclai>`_, `Martin Yuan <https://github.com/iseeyuan>`_
+
+Introduction
+------------
+
+This tutorial introduces the steps to use lite interpreter on iOS and Android. We'll be using the ImageSegmentation demo app as an example. Since lite interpreter is currently in the prototype stage, a custom pytorch binary from source is required.
+
+Android
+-------------------
+
+1. **Prepare model**: Prepare the lite interpreter version of model by run the script below to generate the scripted model deeplabv3_scripted.pt and deeplabv3_scripted.ptl
+
+.. code:: python
+
+    import torch
+
+    model = torch.hub.load('pytorch/vision:v0.7.0', 'deeplabv3_resnet50', pretrained=True)
+    model.eval()
+
+    scripted_module = torch.jit.script(model)
+    # Export full jit version model (not compatible lite interpreter), leave it here for comparison
+    scripted_module.save("deeplabv3_scripted.pt")
+    # Export lite interpreter version model (compatible with lite interpreter)
+    scripted_module._save_for_lite_interpreter("deeplabv3_scripted.ptl")
+
+2. **Build libtorch lite for android**: Build libtorch for android for all 4 android abis (``armeabi-v7a``, ``arm64-v8a``, ``x86``, ``x86_64``) ``BUILD_LITE_INTERPRETER=1 ./scripts/build_pytorch_android.sh``. For example, if it will be tested on Pixel 4 emulator with ``x86``, use cmd ``BUILD_LITE_INTERPRETER=1 ./scripts/build_pytorch_android.sh x86`` to specify abi to save built time. After the build finish, it will show the library path:
+
+
+.. code-block:: bash
+
+   BUILD SUCCESSFUL in 55s
+   134 actionable tasks: 22 executed, 112 up-to-date
+   + find /Users/chenlai/pytorch/android -type f -name '*aar'
+   + xargs ls -lah
+   -rw-r--r--  1 chenlai  staff    13M Feb 11 11:48 /Users/chenlai/pytorch/android/pytorch_android/build/outputs/aar/pytorch_android-release.aar
+   -rw-r--r--  1 chenlai  staff    36K Feb  9 16:45 /Users/chenlai/pytorch/android/pytorch_android_torchvision/build/outputs/aar/pytorch_android_torchvision-release.aar
+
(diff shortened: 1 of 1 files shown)


prototype_source/lite_interpreter.rst
```diff
@@ -0,0 +1,229 @@
...
+Introduction
+------------
+
+This tutorial introduces the steps to use lite interpreter on iOS and Android. We'll be using the ImageSegmentation demo app as an example. Since lite interpreter is currently in the prototype stage, a custom pytorch binary from source is required.
+
+
+To get ImageSegmenation demo app,
+
+**iOS**: https://github.com/pytorch/ios-demo-app/tree/master/ImageSegmentation
+
```
xta0, line 15: Missing the Android link?

cccclai: Oh good catch. I should remove this line. I move the demo link to the corresponding Android/iOS session.

prototype_source/lite_interpreter.rst
```diff
@@ -0,0 +1,229 @@
...
+5. **Test app**: Build and run the `ImageSegmentation` app in Android Studio
+
+iOS
+-------------------
+Get ImageSegmentation demo app in iOS: https://github.com/pytorch/ios-demo-app/tree/master/ImageSegmentation
+
+1. **Prepare model**: Same as Android.
+
+2. **Build libtorch lite for android**:
```
xta0, line 102: typo: for iOS

prototype_source/lite_interpreter.rst
```diff
@@ -0,0 +1,229 @@
...
+        }
+        return self;
+    }
+
+
+5.2 Update ``ViewController.swift``
+
+.. code-block:: swift
+
+    //  if let filePath = Bundle.main.path(forResource:
```
xta0, line 169: Why changing the swift code? Looks like nothing changed

xta0, line 163: I think if we save the model as `.pt` rather than `.ptl`, the swift code can still be reused.

prototype_source/lite_interpreter.rst
```diff
@@ -0,0 +1,229 @@
...
+
+.. code-block:: bash
+
+   SELECTED_OP_LIST=deeplabv3_scripted.yaml BUILD_PYTORCH_MOBILE=1 IOS_PLATFORM=SIMULATOR BUILD_LITE_INTERPRETER=1 ./scripts/build_ios.sh
+
+**Android**: Take the x86 build for example, the command should be:
+
+.. code-block:: bash
+
+   SELECTED_OP_LIST=deeplabv3_scripted.yaml BBUILD_LITE_INTERPRETER=1 ./scripts/build_pytorch_android.sh x86
```
xta0, line 209: typo : BBUILD

prototype_source/lite_interpreter.rst
cccclai: Yeah it will help reducing the code change in the app. But I'm a bit concerned about mixing these two different model types and thought it's harder to explain it to users if they use the wrong model type in jit by accident.

model.ptl -> lite interpreter is ok
model.pt -> lite interpreter will throw error

model.ptl -> jit is ok
model.pt -> jit is ok

prototype_source/lite_interpreter.rst
```diff
@@ -0,0 +1,223 @@
...
+
+In the snippet above, you first need to load the ScriptModule. Then, use export_opnames to return a list of operator names of the ScriptModule and its submodules. Lastly, save the result in a yaml file.
+
+3. To run the build script locally with the prepared yaml list of operators, pass in the yaml file generate from the last step into the environment variable SELECTED_OP_LIST. Also in the arguments, specify BUILD_PYTORCH_MOBILE=1 as well as the platform/architechture type.
+
+**iOS**: Take the simulator build for example, the command should be:
+
+.. code-block:: bash
+
+   SELECTED_OP_LIST=deeplabv3_scripted.yaml BUILD_PYTORCH_MOBILE=1 IOS_PLATFORM=SIMULATOR BUILD_LITE_INTERPRETER=1 ./scripts/build_ios.sh
```
xta0, line 203: Dont need `BUILD_PYTORCH_MOBILE=1` I believe

cccclai: Hmm for simulator as well? I've been using this cmd and it's also mentioned on the mobile page `https://pytorch.org/mobile/ios/#build-pytorch-ios-libraries-from-source`? If it's not needed, we probably also need to update the mobile page?

prototype_source/lite_interpreter.rst
```diff
@@ -0,0 +1,221 @@
...
+.. code-block:: bash
+
+   BUILD SUCCESSFUL in 55s
+   134 actionable tasks: 22 executed, 112 up-to-date
+   + find /Users/chenlai/pytorch/android -type f -name '*aar'
+   + xargs ls -lah
+   -rw-r--r--  1 chenlai  staff    13M Feb 11 11:48 /Users/chenlai/pytorch/android/pytorch_android/build/outputs/aar/pytorch_android-release.aar
+   -rw-r--r--  1 chenlai  staff    36K Feb  9 16:45 /Users/chenlai/pytorch/android/pytorch_android_torchvision/build/outputs/aar/pytorch_android_torchvision-release.aar
+
+3. **Use the PyTorch Android libraries built from source in the ImageSegmentation app**: Create a folder `libs` in the path, the path from repository root will be `ImageSegmentation/app/libs`. Copy `pytorch_android-release` to the path ``ImageSegmentation/app/libs/pytorch_android-release.aar``. Copy `pytorch_android_torchvision` (downloaded from `Pytorch Android Torchvision Nightly <https://oss.sonatype.org/#nexus-search;quick~torchvision_android/>`_) to the path ``ImageSegmentation/app/libs/pytorch_android_torchvision.aar``. Update the `dependencies` part of ``ImageSegmentation/app/build.gradle`` to
```
iseeyuan, line 43: Is it using "libraries from nightly build", instead of "from source"? Could we add "nightly build" specifically in the title, since it would be the suggested way for users to try lite interpreter?

iseeyuan: It's for Android nightly build. How about iOS?

brianjo:
I'm going to close this one as I landed this recipe in https://github.com/pytorch/tutorials/pull/1378. Thanks!

prototype_source/lite_interpreter.rst
cccclai: > Is it using "libraries from nightly build", instead of "from source"? Could we add "nightly build" specifically in the title, since it would be the suggested way for users to try lite interpreter?

It is only for torch vision, since I ran into some issues if I use the official one. 

> It's for Android nightly build. How about iOS?

It is a workaround to make lite interpreter built from source to work in Android. The workaround is to use torch vision from nightly, instead of the official one. iOS doesn't need this.

cccclai: Very good feedback. I am trying to think what is the best way to describe it..

PR #1372, opened 2021-02-23
Add array to control which sections get collapsed

Bgreen23:
This PR adds an array to the layout file. To collapse a section add the caption text to the array.
Ex: `var collapsedSections = ['Learning PyTorch', 'Audio'];`  
In that example, the Learning PyTorch and Audio sections will be collapsed and all other sections will be expanded by default


ydcjeff:
@brianjo @Bgreen23 the latest pytorch tutorials site still has left nav bar collapsed even after `collapsedSections` is null.

PR #1373, opened 2021-02-23
Update CUDA version to 10.2

ranman:


.circleci/config.yml
```diff
@@ -122,7 +122,7 @@ pytorch_tutorial_build_defaults: &pytorch_tutorial_build_defaults
       command: |
         set -e
 
-        export pyTorchDockerImageTag=291
+        export pyTorchDockerImageTag=9de29bef4a5dc0dd1dd19428d83e5a66a44a1ed2-d5ed57569b35aaa56850259a988130c0625addf2
```
seemethere, line 125: Typically the base images that we utilize for CI are just a commit SHA from `$(git rev-parse HEAD:.circleci/docker)` so they shouldn't include a `-${WORKFLOW_ID}` which is what I'm seeing here.

Basically any tag with a hyphen is reaped after 2 weeks in our build system by the ECR garbage collection script.

```suggestion
        export pyTorchDockerImageTag=9de29bef4a5dc0dd1dd19428d83e5a66a44a1ed2
```

brianjo:
Looks like we're really close on this. I'm rerunning the two failed workers. I think these are probably transient errors. I'll look deeper if they fail again.

brianjo:
@malfet Looks like a windows issue. Can you look at it.

brianjo:
Since this passes all but the two windows tests, I'm going to deploy this to allow us to start to test 1.8. Thanks!

PR #1386, opened 2021-03-03
Update spatial_transformer_tutorial.py

brianjo:
Fixes issue with 403 to MNIST.

### DIFF
--- a/beginner_source/fgsm_tutorial.py
+++ b/beginner_source/fgsm_tutorial.py
@@ -99,6 +99,11 @@ from torchvision import datasets, transforms
 import numpy as np
 import matplotlib.pyplot as plt
 
+from six.moves import urllib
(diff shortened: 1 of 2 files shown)


beginner_source/fgsm_tutorial.py
```diff
@@ -99,6 +99,11 @@
 import numpy as np
 import matplotlib.pyplot as plt
 
+from six.moves import urllib
```
seemethere, line 102: We should make sure to add a note and a link to https://github.com/pytorch/vision/issues/3497

```suggestion
# NOTE: This is a hack to get around "User-agent" limitations when downloading MNIST datasets
#       see, https://github.com/pytorch/vision/issues/3497 for more information
from six.moves import urllib
```

PR #1387, opened 2021-03-03
[fix] Wrong code formatting

z-a-f:
The code was not formatted properly, which caused the subsequent steps to fail


jlin27:
@brianjo  - Could you take a quick look for merge? Looks like a straight-forward change, but tagging cause I think y'all are in the middle of a release so flagging in cause you have a process setup

brianjo:
Here's the preview. 
https://deploy-preview-1387--pytorch-tutorials-preview.netlify.app/intermediate/quantized_transfer_learning_tutorial.html?highlight=quantized
Looks good. I'll merge it.

PR #1408, opened 2021-03-11
Small typo fix 'weigth' -> 'weight'

RitwikGupta:


brianjo:
Thanks for this Ritwik. There's an unrelated error that popped up and I'm going try to fix it in another PR then I'll bring this fix in. @RitwikGupta

PR #1414, opened 2021-03-15
Exclude audio_preprocessing_tutorial.py on Windows CI build

skyline75489:


brianjo:
Hey Chester, thanks for the PR. I've rebased and let's see if all of this works.

skyline75489:
Seems like there's a new failure "Key already registered with the same priority: GroupSpatialSoftmax". I'll talk to @guyang3532 about it and see if I fix it here.

skyline75489:
@brianjo The "Key already registered with the same priority: GroupSpatialSoftmax" failure was caused by installing *both* the CPU and GPU version of torch on Windows. I've fixed that in this PR 😃

brianjo:
This looks like it's working. I'm going to go ahead and merge it. Thanks so much for your work on it.

PR #1444, opened 2021-03-25
Parametrizations tutorial

Lezcano:
Creates the tutorial for the parametrizations functionality. This was discussed in the issue https://github.com/pytorch/pytorch/issues/7313 and implemented in the PR https://github.com/pytorch/pytorch/pull/33344

cc @albanD @IvanYashchuk @toshas


brianjo:
Looks like this requires PyTorch 1.9 which we don't build on yet. You can remove the _tutorial from the file name to add it without running the code. If you want to wait for 1.9 to publish, I have some other options for testing this.

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+    return X.triu() + X.triu(1).transpose(-1, -2)
+
+X = torch.rand(3, 3)
+A = symmetric(X)
+print(A)
+assert torch.allclose(A, A.T)
+
+###############################################################################
+# We can then use this idea to implement a linear layer with symmetric weights:
+class LinearSymmetric(nn.Module):
```
vadimkantorov, line 65: Less of a straw-man (for this simple case) would be a module deriving from Linear and adding a `weight` property

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+assert torch.allclose(layer.weight, symmetric(layer.parametrizations.weight.original))
+
+###############################################################################
+# Parametrizations are first-class citizens
+# -----------------------------------------
+# Since ``layer.parametrizations`` is an `nn.ModuleList`, it means that the parametrizations
+# are properly registered as submodules of the original module. As such, the same rules
+# for registering parameters in a module apply to register a parametrization.
+# For example, if a parametrization has parameters, these will be moved from CPU
+# to CUDA when calling ``model = model.cuda()``.
```
vadimkantorov, line 172: It may be good to add a note explaining how reparametrization magic is implemented under the hood

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+assert torch.allclose(X.T @ X, torch.eye(3))  # X is orthogonal
+
+###############################################################################
+# This may also be used to prune a parametrized module, or to reuse parametrizations. For example,
+# we may use the fact that the exponential of matrices maps the symmetric matrices to the
+# Symmetric Positive Definite (SPD) matrices, and the skew-symmetric matrices to the orthogonal
+# matrices. Using these two facts, we may reuse the parametrizations
+class MatrixExponential(nn.Module):
+    def forward(X):
+        return torch.matrix_exp(X)
```
vadimkantorov, line 222: Could `torch.matrix_exp` be directly used instead of `MatrixExponential()`?

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+        self.mask = torch.bernoulli(mask)
+
+    def forward(self, X):
+        return X * self.mask
+
+    def right_inverse(self, A):
+        return A
+
+###############################################################################
+# In this case, it is not true that ``forward(right_inverse(X)) == X``. This is
```
vadimkantorov, line 323: Would reparametrizations work for double-backward? A related question asked here: https://github.com/pytorch/pytorch/issues/55368

intermediate_source/parametrizations.py
Lezcano: That one's tricky. If you think about it, it'd go as something like this but with a bit more flare:
```python
import torch.nn as nn

class MyLin(nn.Linear):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._weight = self.weight  # Why does this line work?
        delattr(self, "weight")

    @property
    def weight(self):
        return self._weight

lin = MyLin(3, 4)
```
The line above works because of how `nn.Module` handles `__getitem__` and `__setitem__`. One could almost say that works "out of pure chance". What happens in that line is:
- It calls `__getattribute__`
- `__getattribute__` finds the property `self.weight` and calls it
- The property looks for `self._weight`. At that time it does not exist, so it raises an `AttributeError`.
- Since `__getattribute__` got an `AttributeError`,  `nn.Module.__getattr__` is called
- `nn.Module.__getattr__` finds the `self.weight` that was created in `nn.Linear.__init__` and returns it

This is quite a mess really. That's why I went for the simpler method here.

vadimkantorov: I had used `register_parameter('weight', None)` instead: https://gist.github.com/vadimkantorov/4f34fe60d2ef00e72dcad16512d224af

vadimkantorov: Ah, the trick is that you still want to access old `_weight`, while I didn't need it

Lezcano: That works for exactly the same convoluted reason, but changing `__getattr__` and `__getattribute__` with `__setattr__`  and `__setattribute__`. Again, it works, but it's tricky to know why it works.

intermediate_source/parametrizations.py
Lezcano: No, it could not. At the moment `Parametrizations` are defined to be `nn.Modules`, so they do not support the functional API.

I do not know whether it would be necessary to discuss this, as in the whole tutorial it's been made clear that a parametrization is just a plain `nn.Module`. What do you think @albanD ?

vadimkantorov: My reasoning: in many places in PyTorch one expects both functions and module objects work (especially in older areas of PyTorch). Whenever this is violated - quantization, module tracing etc, I'm always suspicious of undeclared magic :)

In this case, it begs the question, since the module here is just a wrapper

intermediate_source/parametrizations.py
Lezcano: That is a good point. At the moment that example but using
```python
import torch.nn.utils.parametrize as P
model = P.register_parametrization(torch.nn.Linear(2, 5), "weight", torch.nn.ReLU())
```
also breaks with the same error message. That being said, it smells like it should be a problem in the implementation of parametrizations. Perhaps @albanD can give a bit more insight on what's going on..

intermediate_source/parametrizations.py
Lezcano: I think that that would be too technical (it uses dynamically generated classes). I think something that could be done would be to stress more what happens after you call `register_parametrization`. In particular, the fact that it creates a `ModuleDict` under `module.parametrizations`, and each of those modules is a `ParametrizationList` and so on.
In fact, that's all `register_parametrization` is doing modulo the dynamically generated classes magic. This would also help clarifying why it can be used with `nn.Modules` but not with plain old functions.

vadimkantorov: I think it's better to mention it, then there wouldn't be a surprise of some generated class name if there is some exception. Also this would be a useful reminder for the user to not do any similar magics if they fear interference

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+using a function that maps matrices to orthogonal matrices. In the case of weight
+and spectral normalization, they divide the original parameter by its norm.
+
+More generally, all these examples use a function to put extra structure on the parameters.
+In other words, they use a function to constrain the parameters.
+
+In this tutorial, you will learn how to implement and use this pattern to write and
+put constraints on your model. Doing so is as easy as writing your own ``nn.Module``.
+
+Requirements
```
albanD, line 39: Is that section a common thing in tutorials? If not, a single sentence would be enough here I think.

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+------------
+``"torch>=1.9.0a0+7aeee28"``
```
albanD, line 41: I think that 1.9.0 would be enough here no?
I don't think users know exactly what the hash means and how to check if their nightly version is before or after this one hash.

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+import torch
+import torch.nn as nn
+import torch.nn.utils.parametrize as P
+
+def symmetric(X):
+    return X.triu() + X.triu(1).transpose(-1, -2)
+
+X = torch.rand(3, 3)
+A = symmetric(X)
+print(A)
```
albanD, line 60: Is the print useful here? Will it actually be rendered in the final tutorial once they run?

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+#    not very problematic for a linear layer, but imagine having to reimplement a CNN or a
+#    Transformer...
+# 2) It does not separate the layer and the parametrization.  If the parametrization were
+#    more difficult, we would have to rewrite its code for each layer that we want to use it
+#    in.
+# 3) It recomputes the parametrization everytime forward is called. If we used the layer
+#    several times during the forward pass, (imagine the recurrent kernel of an RNN) we would
+#    be recomputing the same ``A`` every time the layer is called.
+#
+# Parametrizations come to solve all these and other problems.
```
albanD, line 92: This sentence is actually attached to the above paragraph in the rendered version. Which doesn't look great.
Maybe we should move it as the first sentence of the next section?

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+print(f"Parametrized:\n{layer}")
+
+###############################################################################
+# We see that the ``Symmetric`` class has been registered under a ``parametrizations`` attribute.
+# This ``parametrizations`` attribute is an ``nn.ModuleList``, and it can be accessed as such
+print(layer.parametrizations.weight[0])
+
+###############################################################################
+# Note that each element in the `ModuleList` is itself a list, and we have to select the first
+# element of this list. It will be clear later the reason for this, when we see how to contactenate
```
albanD, line 148: Do we want to use chain here instead of concatenate?
Also typo on concatenate.

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+# paramtrizations.
```
albanD, line 149: parametrization

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+# Something that we may notice is that, if we print the parameters, we see that the
+# parameter ``weight`` has been moved
+print(dict(layer.named_parameters()))
+
+###############################################################################
+# It now sits under ``layer.parametrizations.weight.original``
+print(layer.parametrizations.weight.original)
+
+###############################################################################
+# Besides these two small differences, the parametrization is doing exactly the same
```
albanD, line 160: I think it would be nice to specify which are these two from the beginning of the section:
"When a module is parametrized, we find that the module has changed a bit" -> When a module is parametrized, we find that the module has changed in two ways" or something similar?

intermediate_source/parametrizations.py
albanD: I do agree that such explanation would be nice, but I'm not sure this tutorial is the right place for it.
A Note in the doc similar to the ones we have about autograd [herre](https://pytorch.org/docs/stable/notes/autograd.html) for example sound more appropriate.

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+-------------------------------------
+
+Assume that we want to have a square linear layer with symmetric weights, that is,
+with weights :math:`X` such that :math:`X = X^{\intercal}`. One way to do so is
+to copy the upper triangular part of the matrix into its lower triangular part
+"""
+
+import torch
+import torch.nn as nn
+import torch.nn.utils.parametrize as P
```
albanD, line 53: Is `P` really the best name here? should we use something more explicit in this tutorial?

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+print("Here, layer.weight is recomputed every time we call it")
+foo = layer.weight + layer.weight.T
+bar = layer.weight.sum()
+with P.cached():
+    print("Here, it is computed just the first time layer.weight is called")
+    foo = layer.weight + layer.weight.T
+    bar = layer.weight.sum()
+
+###############################################################################
+# Composing Parametrizations
```
albanD, line 193: We use many names for the same thing here, we should pick one I think between composing, chain or concatenate.

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+
+layer = nn.Linear(3, 3)
+P.register_parametrization(layer, "weight", Skew())
+P.register_parametrization(layer, "weight", CayleyMap(3))
+X = layer.weight
+assert torch.allclose(X.T @ X, torch.eye(3))  # X is orthogonal
+
+###############################################################################
+# This may also be used to prune a parametrized module, or to reuse parametrizations. For example,
+# we may use the fact that the exponential of matrices maps the symmetric matrices to the
```
albanD, line 217: While this sentence makes sense if you read it slowly, I feel like we could make it nicer.

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+        A = X.triu(1)
+        return A - A.transpose(-1, -2)
+
+    def is_skew(self, X):
+        return torch.allclose(X, -X.transpose(-1, -2))
+
+    def right_inverse(self, A):
+        if not self.is_skew(A):
+            raise ValueError(f"The provided matrix {A} is not skew-symmetric")
+        return A.triu(1)
```
albanD, line 261: Should we mention here that we only get the upper elements because these are the elements used in the forward?

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+
+###############################################################################
+# This initialization step can be written more succinctly as
+layer_orthogonal.weight = nn.init.orthogonal_(layer_orthogonal.weight)
+
+###############################################################################
+# The name of this method comes from the fact that we would often expect
+# that ``forward(right_inverse(X)) == X``. This is a direct way of rewriting that
+# the forward afer the initalization with value ``X`` should return the value ``X``.
+# This constraint is not enforced in the code. In fact, at times, it might be of
```
albanD, line 307: Maybe "This constraint is not enforced in the code" -> "This constraint is not strongly enforced in practice"

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+        return X * self.mask
+
+    def right_inverse(self, A):
+        return A
+
+###############################################################################
+# In this case, it is not true that ``forward(right_inverse(X)) == X``. This is
+# only true when the matrix ``A`` passed to ``right_inverse`` has zeros in the
+# same positions as the mask. Even then, if we assign a tensor to a pruned parameter,
+# it will comes as no surprise that tensor will be, in fact, pruned.:
```
albanD, line 326: double punctuation .:

intermediate_source/parametrizations.py
albanD: There is no reason it wouldn't.
Your implementation of the reparametrization will need to be double differentiable though.

albanD (commented):
Looks quite good to me.
Only phrasing and minor comments.

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,353 @@
...
+on deep models due to the complexity of the function being optimized.
+This is particularly problematic when working with ill-conditioned models.
+Examples of these are RNNs trained on long sequences and GANs. A number
+of techniques have been proposed in the recent years to regularize these
+models and improve their convergence. On recurrent models, it has been
+proposed to control the singular values of the recurrent kernel for the
+RNN to be well-conditioned. This can be achieved, for example, by making
+the recurrent kernel `orthogonal <https://en.wikipedia.org/wiki/Orthogonal_matrix>`_.
+Another way to regularize recurrent models is via
+"`weight normalization <https://pytorch.org/docs/stable/generated/torch.nn.utils.weight_norm.html>`_".
```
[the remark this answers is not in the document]
Lezcano: I put them in quotation marks to signal that they are a proper name referring to a particular method. Otherwise "weight normalisation" could mean any sort of normalisation that you put on the weights.

intermediate_source/parametrizations.py
Lezcano: I have seen this in the Pruning tutorial: https://pytorch.org/tutorials/intermediate/pruning_tutorial.html

Same with the next comment about the hash.

I agree that leaving one line with the tutorial and just the version 1.9.0 should be enough.

intermediate_source/parametrizations.py
Lezcano: I just wanted to show that A is indeed symmetric via a print, not only a assertion.

intermediate_source/parametrizations.py
Lezcano: I think concatenate is fine. I dropped the word "chain" from the docs in favour of concatenate. I think that either of them represents well what's going on here

intermediate_source/parametrizations.py
vadimkantorov: I guess it's also worth an explicit discussion somewhere

intermediate_source/parametrizations.py
Lezcano: Ooops. Done.

Lezcano:
@albanD I addressed the points that you raised and I corrected a few other things (the code does not break now... I had forgotten to check that...). Even then, there were no major changes in the text.

intermediate_source/parametrizations.py
albanD: That is indeed an interesting question.
But for now everything is expected to be a nn.Module and that sounds enough for me.
If we want to relax that in the future we might be able to do so but that would be out of scope of this tutorial.

intermediate_source/parametrizations.py
albanD: Ok, so it will actually print in the final version of the tutorial?

index.rst
```diff
@@ -320,6 +320,13 @@ Welcome to PyTorch Tutorials
    :link: beginner/hyperparameter_tuning_tutorial.html
    :tags: Model-Optimization,Best-Practice
 
+.. customcarditem::
+   :header: Parametrizations Tutorial
+   :card_description: Learn how to use torch.nn.utils.parametrizations to put constriants on your parameters (e.g. make them orthogonal, symmetric positive definite, low-rank...)
```
albanD, line 325: It's actually `nn.utils.parametrize` no?

albanD (commented):
Just a minor update and it looks mostly good for me.

intermediate_source/parametrizations.py
albanD: Not sure what you mean about the relu part though. This works fine on colab for me:
```python
# !pip uninstall --y torch
# !pip install --pre torch -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html

import torch
import torch.nn.utils.parametrize as P
model = P.register_parametrization(torch.nn.Linear(2, 5), "weight", torch.nn.ReLU())


inp = torch.rand(1, 2, requires_grad=True)

out = model(inp)

g = torch.autograd.grad(out.sum(), model.parameters(), create_graph=True)

print(g)

g[1].exp().sum().backward()

print(inp.grad)
```

index.rst
Lezcano: ooopsss!

intermediate_source/parametrizations.py
Lezcano: Yes. I have updated the tutorial with a comment in the source to make this intention clearer

intermediate_source/parametrizations.py
Lezcano: I meant that this example (which is a modification of that in that PR) breaks:
```python
import torch
import torch.nn.utils.parametrize as P
model = P.register_parametrization(torch.nn.Linear(2, 5), "weight", torch.nn.ReLU())
opt1 = torch.optim.SGD(model.parameters(), lr=1e-3)
opt2 = torch.optim.SGD(model.parameters(), lr=1e-3)

output = model(torch.randn(7, 2))
loss = output.abs().mean()

opt1.zero_grad(); loss.backward(retain_graph=True); opt1.step() # first propagation
opt2.zero_grad(); loss.backward(); opt2.step() # second
```
but it might be because `ReLU` is not twice differentiable?

intermediate_source/parametrizations.py
```diff
@@ -0,0 +1,386 @@
...
+# We use a similar parametrization, copying minus the upper triangular part into the
+# lower-triangular part
+class Skew(nn.Module):
+    def forward(self, X):
+        A = X.triu(1)
+        return A - A.transpose(-1, -2)
+
+
+cnn = nn.Conv2d(in_channels=5, out_channels=8, kernel_size=3)
+parametrize.register_parametrization(layer, "weight", Skew())
```
iamsimha, line 126: Should `layer` be replaced with `cnn`?

Lezcano: Absolutely. Thanks for the catch!

Issue #1442, opened 2021-03-25
example code problem: 


EngrStudent:
In this code:    
https://github.com/pytorch/tutorials/blob/master/recipes_source/recipes/saving_and_loading_models_for_inference.py

lines 83, 84, 86, 87 all try to use "F.relu" which is undefined. 

I think you need the following in your imports:    
import torch.nn.functional as F

krishnan-tech:
I think, we can use `nn.ReLU` instead of `F.relu`.

brianjo:
Thanks for the fix Krishnan!

PR #1446, opened 2021-03-26
Import F from torch.nn.functional

krishnan-tech:
Import `F` from `torch.nn.functional`, because `F.relu` is undefined

Solved problem from #1442


brianjo:
Thanks Krishnan!

PR #1452, opened 2021-04-01
fix dead link

arthursw:


brianjo:
Thanks you for the fix @arthursw!

PR #1455, opened 2021-04-03
Register all the parameters in the optimizer

ageron:
The code at the end registers only the parameters from `model.fc` in the optimizer, although the text underneath says: "Notice although we register all the parameters in the optimizer, the only parameters that are computing gradients (and hence updated in gradient descent) are the weights and bias of the classifier."

To be consistent with this explanation, we should be adding all the parameters from the model.


brianjo:
Thanks @ageron!

PR #1474, opened 2021-04-19
Update rpc_tutorial "Getting Started with Distributed RPC Framework"

H-Huang:
Clean up helper method which aren't need anymore and changed the hyperparameters (train steps, gamma) to allow the agent to train to completion

Currents docs: https://pytorch.org/tutorials/intermediate/rpc_tutorial.html
New docs preview: https://h-huang.github.io/tutorials/intermediate/rpc_tutorial.html


intermediate_source/rpc_tutorial.rst
```diff
@@ -335,17 +313,17 @@ available in the `API page <https://pytorch.org/docs/master/rpc.html>`__.
             rpc.init_rpc(AGENT_NAME, rank=rank, world_size=world_size)
 
             agent = Agent(world_size)
+            print(f"This will run until reward threshold of {agent.reward_threshold}"
+                    " is reached. Ctrl+C to exit.")
             for i_episode in count(1):
-                n_steps = int(TOTAL_EPISODE_STEP / (args.world_size - 1))
-                agent.run_episode(n_steps=n_steps)
```
zhaojuanmao, line 340: feel it is useful to understand how n_steps are divided/parallelized when I read this tutorial

zhaojuanmao (approved):
overall looks good, please see one minor comment

intermediate_source/rpc_tutorial.rst
H-Huang: good point, it is also mentioned in the descriptions so I will add it back in. The reason I changed it was because the default value was too small so the trainer wasn't making any progress; I will just increase the default val.

PR #1489, opened 2021-04-27
Parametrizations fix

Lezcano:
This PR fixes a couple formatting errors that the tutorial on parametrizations has. In particular, I forgot to put a `super().__init__()` in an `nn.Module`  and a section was being rendered as a comment.

I also took the chance point out explicitly in the tutorial that parametrizations are just Python properties, and I made some minor stylistic changes.

@albanD

### DIFF
--- a/intermediate_source/parametrizations.py
+++ b/intermediate_source/parametrizations.py
@@ -141,16 +141,11 @@ print(cnn.weight[2, 2])
 # 3) The unparametrized weight has been moved to ``module.parametrizations.weight.original``
 #
 # |
-# After parametrizing ``weight``, we see that, even though calling ``layer.weight`` returns
-# a tensor, ``layer.weight`` is not a tensor itself, but a property:
-layer = nn.Linear(3, 3)
-parametrization = Symmetric()
-parametrize.register_parametrization(layer, "weight", parametrization)
-print(layer.__class__.weight)
-
-###############################################################################
+# After parametrizing ``weight``, ``layer.weight`` is turned into a
+# `Python property <https://docs.python.org/3/library/functions.html#property>`_.
 # This property computes ``parametrization(weight)`` every time we request ``layer.weight``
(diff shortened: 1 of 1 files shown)


intermediate_source/parametrizations.py
```diff
@@ -132,30 +132,41 @@ def forward(self, X):
...
 #
-# 1) It has a new ``module.parametrizations`` attribute
+# 2) It has a new ``module.parametrizations`` attribute
 #
-# 2) The weight has been moved to ``module.parametrizations.weight.original``
+# 3) The unparametrized weight has been moved to ``module.parametrizations.weight.original``
 #
 # |
-# We may observe this first change by printing the module
+# After parametrizing ``weight``, we see that, even though calling ``layer.weight`` returns
```
albanD, line 144: I feel like this is a bit ambiguous.
And I don't think you want to go into the details of showing the property with `__class__` below.

Mentioning that `layer.weight` is now a property that contains the parametrized value of the weight. Sounds simpler.

intermediate_source/parametrizations.py
```diff
@@ -132,30 +132,41 @@ def forward(self, X):
...
+# a tensor, ``layer.weight`` is not a tensor itself, but a property:
 layer = nn.Linear(3, 3)
-print("Unparametrized:")
-print(layer)
+parametrization = Symmetric()
+parametrize.register_parametrization(layer, "weight", parametrization)
+print(layer.__class__.weight)
+
+###############################################################################
+# This property computes ``parametrization(weight)`` every time we request ``layer.weight``
```
albanD, line 152: This is a good description I think.

albanD (commented):
Looks mostly good.
Did you had a chance to run the script locally to make sure it all works well?

Lezcano:
I just corrected that part making it less technical, and adding a link to the definition of properties in the standard library.

And yes, I run the script. Now everything should be fine.

holly1238:
Thanks! :)

PR #1492, opened 2021-04-29
Update optimization_tutorial.py

Rajjaa:
The definition of the batch size was: the number of data samples seen by the model in each epoch. While in each epoch, the model sees the entire dataset. The batch defines the number of samples passed through the forward pass and used to update the parameters.


holly1238:
Closing this PR. This was fixed and merged in https://github.com/pytorch/tutorials/pull/1490.

PR #1523, opened 2021-05-14
update text tutorials

parmeet:
We would like to update the tutorials to accommodate changes in upcoming release (0.10.0) of torchtext.

Requires: PyTorch 1.9 and torchtext 0.10.0 

cc: @brianjo


beginner_source/translation_transformer.py
```diff
@@ -8,187 +8,154 @@
...
-valid_iter = DataLoader(val_data, batch_size=BATCH_SIZE,
-                        shuffle=True, collate_fn=generate_batch)
-test_iter = DataLoader(test_data, batch_size=BATCH_SIZE,
-                       shuffle=True, collate_fn=generate_batch)
+# Insert special symbols
+for symbol, index in special_symbols.items():
+    vocab_transform[SRC_LANGUAGE].insert_token(symbol, index)
+    vocab_transform[TGT_LANGUAGE].insert_token(symbol, index)
 
+# Set default index to return OOV token is queried
```
cpuhrsch, line 66: nit: This reads a bit weird in my mind. Maybe you mean something like "Return UNK_IDX if a token cannot be found"?

beginner_source/translation_transformer.py
```diff
@@ -8,187 +8,154 @@
...
+special_symbols = {'<unk>': UNK_IDX, '<pad>': PAD_IDX,'<bos>': BOS_IDX, '<eos>': EOS_IDX}
 
-train_iter = DataLoader(train_data, batch_size=BATCH_SIZE,
-                        shuffle=True, collate_fn=generate_batch)
-valid_iter = DataLoader(val_data, batch_size=BATCH_SIZE,
-                        shuffle=True, collate_fn=generate_batch)
-test_iter = DataLoader(test_data, batch_size=BATCH_SIZE,
-                       shuffle=True, collate_fn=generate_batch)
+# Insert special symbols
+for symbol, index in special_symbols.items():
```
cpuhrsch, line 62: What if any of these symbols already exist in the Vocab? I guess in this case we make a hard assumption that it won't, but I think we should mention what would happen otherwise etc.

beginner_source/translation_transformer.py
cpuhrsch: Similar to above we could also mention that the Vocab would throw an error if this method wasn't set and explain how to unset it again and check whether it's been set.

parmeet: Sounds good! Yes, got to work on text and comments a bit :). I plan to include more elaborate description on the behavior in the text part.

beginner_source/translation_transformer.py
parmeet: Sure, will clarify in the comments and perhaps write a special_symbols exists proof code (we can check if they exist we will just do reassign instead of insert).

beginner_source/translation_transformer.py
cpuhrsch: Ideally something like that would be covered in a central tutorial similar to torchaudio's [data augmentation tutorials](https://pytorch.org/tutorials/beginner/audio_preprocessing_tutorial.html#data-augmentation).

parmeet: > Ideally something like that would be covered in a central tutorial similar to torchaudio's [data augmentation tutorials](https://pytorch.org/tutorials/beginner/audio_preprocessing_tutorial.html#data-augmentation).

Yes, this seems to be a neat idea. We should plan to create tutorial on torchtext's  core utility and functionalities!

cpuhrsch (approved):

parmeet:
create new PR #1552 due to re-basing issues.

PR #1528, opened 2021-05-18
Update profiler recipe doc (1.9)

ilia-cher:
Summary:
Update the profiler recipe to use the new API and features

Test Plan:
make html-noplot


brianjo (commented):
Looks good. I'll test this against 1.9 RC and I'll merge this when we release 1.9.

PR #1529, opened 2021-05-20
Add loss_fn as an input to the test function

ValeriVoev:
The function `test()` depends on `loss_fn` just like `train()` does. It is better to explicitly provide `loss_fn` as an argument instead of relying on a global `test_fn` object. Also, this is more consistent with how `train()` is defined. And it also makes it more explicit that `train()` has a dependency on `optimizer`, while `test()` doesn't.


suraj813 (approved):
This is a good callout, thank you for making this change!

Issue #1300, opened 2020-12-30
Runtime Error: 'lengths' argument should be a 1D CPU int64 tensor, but got 1D cuda:0 Long tensor


bill2bill:
I've been following the chatbot tutorial (https://pytorch.org/tutorials/beginner/chatbot_tutorial.html#run-model) and discovered the code does not run out of the box for me. 

After training the model and inputting custom test data for the bot to respond I get the error:

`RuntimeError: 'lengths' argument should be a 1D CPU int64 tensor, but got 1D cuda:0 Long tensor`

I found a similar issue with a fix on the Pytorch Github issues, https://github.com/pytorch/pytorch/issues/43227

Their solution was to use the .cpu() on the length argument.

On line 683, of https://github.com/pytorch/tutorials/blob/master/beginner_source/chatbot_tutorial.py, I applied the .cpu() method to the variable input_sequence and that resolved my issue.

chaojihuihui:
well

krkv:
Ran into the same issue

holly1238:
Closing this issue. This issue was fixed and merged in https://github.com/pytorch/tutorials/pull/1535

PR #1535, opened 2021-05-24
Update chatbot_tutorial.py

krkv:
Fix lengths to cpu. See issue #1300


albanD (approved):
Good catch!
Thanks for the PR!

Could you rebase on top of master so that we can merge it please?

Issue #1486, opened 2021-04-25
Incorrect test loss calculation in Introduction to PyTorch tutorial


hxyue1:
I think the testing function in the tutorial calculates the test loss incorrectly. 

```
def test(dataloader, model):
    size = len(dataloader.dataset)
    model.eval()
    test_loss, correct = 0, 0
    with torch.no_grad():
        for X, y in dataloader:
            X, y = X.to(device), y.to(device)
            pred = model(X)
            test_loss += loss_fn(pred, y).item()
            correct += (pred.argmax(1) == y).type(torch.float).sum().item()
    test_loss /= size
    correct /= size
    print(f"Test Error: \n Accuracy: {(100*correct):>0.1f}%, Avg loss: {test_loss:>8f} \n")
```

If I'm not mistaken, the loss function ```nn.CrossEntropyLoss``` already averages the loss across the batch (in this case batch_size=64). So it doesn't make sense to divide the sum of the batch losses by the number of test observations, ```size = len(dataloader.dataset)``` = 10000. I think what should be done is to divide by the total number of batches instead, ```num_batches = len(dataloader)``` = 157.

This becomes obvious when you compare the training loss at various epochs with the training loss on the test set. Test loss is too small!

Epoch 1
-------------------------------
loss: **2.305201**  [    0/60000]
loss: **2.299808**  [ 6400/60000]
loss: **2.290298**  [12800/60000]
loss: **2.282613**  [19200/60000]
loss: **2.265882**  [25600/60000]
loss: **2.251907**  [32000/60000]
loss: **2.253182**  [38400/60000]
loss: **2.232780**  [44800/60000]
loss: **2.240818**  [51200/60000]
loss: **2.198148**  [57600/60000]
Test Error:
 Accuracy: 47.9%, Avg loss: **0.034639**
 
I first saw this in the Quickstart portion of the tutorial: https://pytorch.org/tutorials/beginner/basics/quickstart_tutorial.html, but the code is repeated in the Optimization section too: https://pytorch.org/tutorials/beginner/basics/optimization_tutorial.html.

PR #1536, opened 2021-05-24
Fix testloss calculation in Quickstart and Optimization tutorials

FaranIdo:
Fix incorrect calculation of test loss in Quickstart and Optimization tutorial, reported in #1486


suraj813 (approved):
Looks good to me! Thank you for making the correction :)

PR #1537, opened 2021-05-24
fix: fix xb not tensor issue

Franceshe:
fix xb not tensor issue which make it having trouble with multiplying with weights
Otherwise without the fixing code, got following error:
```
pytroch unsupported operand type(s) for @: 'numpy.ndarray' and 'Tensor' pytorch
```


Franceshe:
Just sign the CLA for contributing projects.

suraj813:
On running which line are you getting this error? From what I see, `xb` is a slice of `x_train` both of which are tensors. I'm not sure where are you passing a numpy array

PR #1544, opened 2021-05-27
"point-to-point" -> "collective" for describing all-reduce

andwgu:
Fix the comment from describing all-reduce as a point-to-point communication to describing all-reduce as a collective communication.


H-Huang (approved):
lgtm

PR #1559, opened 2021-06-08
Recipe for skipping parameter init

jbschlosser:
Demonstrates how to skip parameter init for PyTorch 1.9, including an example of the minimal module updates required to support it (i.e. a properly-implemented `device` kwarg).

Unfortunately, the helper function `torch.nn.utils.skip_init()` was merged just after the 1.9 release branch cut so it's not mentioned yet. Is it worth trying to get it cherry-picked in?

Edit: Changed to reference PyTorch 1.10.0 instead since that's when `torch.nn.utils.skip_init()` will come in; moved under `prototype_source`.


ezyang:
It feels like it would be better to wait to document this for when the combinator shows up, and then it's way simpler to explain.

ezyang (approved):
okey dokey

PR #1566, opened 2021-06-08
Update audio manipulation tutorial 

carolineechen:
part of torchaudio 0.9 release [checklist](https://github.com/pytorch/audio/issues/1527)

- update HTK support and metadata printing
- add resampling tutorial

### DIFF
--- a/beginner_source/audio_preprocessing_tutorial.py
+++ b/beginner_source/audio_preprocessing_tutorial.py
@@ -52,7 +52,6 @@ from botocore.config import Config
 import requests
 import matplotlib
 import matplotlib.pyplot as plt
-import numpy as np
 import pandas as pd
 import time
 from IPython.display import Audio, display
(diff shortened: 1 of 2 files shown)


beginner_source/audio_preprocessing_tutorial.py
```diff
@@ -52,6 +52,8 @@
 import requests
 import matplotlib
 import matplotlib.pyplot as plt
+import numpy as np
```
mthrok, line 55: Is this used?

carolineechen: this is needed for comparison to librosa, which takes an np array as input

mthrok: But I do not see any use of `np.XX`.

carolineechen: you're right, importing np is not necessary

PR #1572, opened 2021-06-10
Revert "Added cache for incremental build."

brianjo:
Reverts pytorch/tutorials#1184

This PR was old and I think it's causing trouble with the builds. Trying a revert to see if the build succeeds.


brianjo:
I believe this PR may have created an HTML folder in GH Pages and started copying files there instead of the proper root folder. This is the rogue folder: https://github.com/pytorch/tutorials/tree/gh-pages/html

PR #1577, opened 2021-06-13
PIL to read images in CustomImageDataset class

mustious:
using read_image() method raises an error: "TypeError: pic should be PIL Image or ndarray. Got <class 'torch.Tensor'>"


jicksaw:
@brianjo @fmassa Could you take another look at this?

It's still relevant, because the example on top of the tutorial (and in quickstart) uses `ToTensor()`:
```
training_data = datasets.FashionMNIST(
    root="data",
    train=True,
    download=True,
    transform=ToTensor()
)
```

Later in the tutorial the image is loaded with `read_image()`:
```
    def __getitem__(self, idx):
        img_path = os.path.join(self.img_dir, self.img_labels.iloc[idx, 0])
        image = read_image(img_path)
        label = self.img_labels.iloc[idx, 1]
        if self.transform:
            image = self.transform(image)
        if self.target_transform:
            label = self.target_transform(label)
        return image, label
```

The error message is a confusing for a newbie trying to patch together code for their first attempts.

brianjo:
I added Svetlana and Carl to this. I've started a rebase so that will be done when somebody takes a look at it.

svekars:
Thanks for bringing this up. Where does the error occur? I just ran this in Colab and I'm not seeing the error.

jicksaw:
The error only happens if you try making your own dataset only using the example code. Here's an example of something a beginner might code
```python
import os
import torch
import pandas as pd
from torchvision.io import read_image
from torch.utils.data import Dataset
from torchvision.transforms import ToTensor

import urllib.request
import csv

# Setup for example code
urllib.request.urlretrieve(
    "https://raw.githubusercontent.com/pytorch/tutorials/f26db6b46dba5a109707f397499db7a014768a8d/_static/img/panda.png",
    "panda.png",
)
annotations = pd.DataFrame([["panda.png", "panda"]], columns=["path", "label"])
annotations.to_csv("annotations.csv", index=False)

#
# Hypothetical code a beginner could patch together from examples
#

# Example code for custom dataset copied from tutorial
class CustomImageDataset(Dataset):
    def __init__(self, annotations_file, img_dir, transform=None, target_transform=None):
        self.img_labels = pd.read_csv(annotations_file)
        self.img_dir = img_dir
        self.transform = transform
        self.target_transform = target_transform

    def __len__(self):
        return len(self.img_labels)

    def __getitem__(self, idx):
        img_path = os.path.join(self.img_dir, self.img_labels.iloc[idx, 0])
        image = read_image(img_path)
        label = self.img_labels.iloc[idx, 1]
        if self.transform:
            image = self.transform(image)
        if self.target_transform:
            label = self.target_transform(label)
        return image, label


# Applying earlier example of using a dataset for custom dataset
cid = CustomImageDataset(
    annotations_file="annotations.csv",
    img_dir="",
    transform=ToTensor(),
    target_transform=None,
)

# TypeError: pic should be PIL Image or ndarray. Got <class 'torch.Tensor'>
cid.__getitem__(0)
```

PR #1580, opened 2021-06-15
More update to use mobile interpreter

cccclai:
1. Update the iOS gpu workflow to use mobile interpreter
2. Fix the header file in mobile interpreter tutorial
3. Remove the old lite interpreter tutorial


brianjo:
Thanks Chen! I'll merge that when it passes tests.

PR #1586, opened 2021-06-17
Correction on the use of "epoch"

aytekinar:
Made the switch from "epoch" to "mini-batches" to clarify the training process.

[ci skip]


aytekinar:
Usually, the term "epoch" is reserved for "_one full pass over the entire dataset_." In the example, we're processing only a (mini-)batch of 64 samples at each iteration, and hence, in my opinion, the correct term should have been simply "iteration."

aytekinar:
My bad... I was not careful enough with the `train` script. Dropping the PR because the term seems to be used properly...

PR #1594, opened 2021-06-29
Add custom function tutorial for fusing conv and bn

soulitzer:


intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,344 @@
...
+        else:
+            # If not training, we use running mean and var, which are constant wrt x
+            grad_out = grad_out / (ctx.sqrt_var[None, :, None, None] + ctx.eps)
+        # Conv2d backward
+        grad_input = F.conv2d(X.transpose(0, 1), grad_out.transpose(0, 1)).transpose(0, 1)
+        grad_X = F.conv_transpose2d(grad_out, conv_weight)
+        return grad_X, grad_input, None, None, None, None, None
+
+######################################################################
+# Now that the bulk of the work has been done, it is time to combine
```
soulitzer, line 175: Remove repeat

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,344 @@
+# -*- coding: utf-8 -*-
+"""
+Fusing Convolution and Batch Norm using Custom Function
+=======================================================
+
+Fusing adjacent convolution and batch norm layers together is typically an
+inference-time optimization to improve run-time. It is typically achieved
```
albanD, line 7: nit: repeated "typically"

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,344 @@
...
+
+######################################################################
+# When testing with gradcheck, it is important to use double precision
+weight = torch.rand(5, 3, 3, 3, requires_grad=True, dtype=torch.double)
+X = torch.rand(10, 3, 7, 7, requires_grad=True, dtype=torch.double)
+torch.autograd.gradcheck(Conv2D.apply, (X, weight))
+
+######################################################################
+# Batch Norm
+def expand(t):
```
albanD, line 70: The name is not the best here as there is already a torch.expand function that does something else

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,344 @@
...
+    # Helper function to unsqueeze the dimensions that we reduce over
+    return t[None, :, None, None]
+
+def batch_norm_backward(grad_out, X, sum, sqrt_var, N, eps):
+    # To simplify our derivation, we follow the chain rule and compute
+    # the following gradients before accumulating them all into a final grad_input
+    #  1) X of the numerator
```
albanD, line 77: These points are not super clear. What do you mean "of the numerator"?

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,344 @@
...
+X = torch.rand(10, 3, 7, 7, requires_grad=True, dtype=torch.double)
+torch.autograd.gradcheck(Conv2D.apply, (X, weight))
+
+######################################################################
+# Batch Norm
+def expand(t):
+    # Helper function to unsqueeze the dimensions that we reduce over
+    return t[None, :, None, None]
+
+def batch_norm_backward(grad_out, X, sum, sqrt_var, N, eps):
```
albanD, line 74: Why move this in a separate function?
If that's for re-use in the combined function, then you should do the same for conv

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,344 @@
...
+        ctx.eps = eps
+        ctx.sum = sum
+        ctx.N = N
+        ctx.sqrt_var = sqrt_var
+        mean = sum / N
+        denom = sqrt_var + eps
+        return (X - expand(mean)) / expand(denom)
+
+    @staticmethod
+    @once_differentiable
```
albanD, line 114: If we're saying this is once differentiable, we can "optimize" the code more agressively to re-use Tensors by doing most ops inplace.

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,344 @@
...
+        sum = X.sum(dim=(0, 2, 3))
+        var = X.var(unbiased=True, dim=(0, 2, 3))
+        N = X.numel() / X.size(1)
+        sqrt_var = torch.sqrt(var)
+        ctx.save_for_backward(X)
+        ctx.eps = eps
+        ctx.sum = sum
+        ctx.N = N
+        ctx.sqrt_var = sqrt_var
+        mean = sum / N
```
albanD, line 109: Here as well, you can show that you can do everything inplace to reduce memory usage

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,344 @@
...
+convolution batch norm pair can be a significant reduction.
+
+In this tutorial, we avoid this extra allocation by combining convolution
+and batch norm into a single layer (as a custom function). In the forward
+of this combined layer, we perform normal convolution and batch norm as-is,
+with the only difference being that we will only save the inputs to the convolution.
+To obtain the input of batch norm, which is necessary to backward through
+it, we recompute convolution forward again during the backward pass.
+
+For simplicity, in this tutorial we will not support the bias, stride, dilation, or
```
albanD, line 30: Also the saved statistics for batchnorm  no?

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,344 @@
...
+# Conv2D again in (2).
+class FusedConvBN2DFunction(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, X, conv_weight, running_mean, running_var, training, exp_avg_factor=0.1, eps=1e-6):
+        assert not running_mean.requires_grad and not running_var.requires_grad
+        assert X.ndim == 4
+        ctx.save_for_backward(X, conv_weight)  # (1) Only need to save this single buffer for backward!
+        X = F.conv2d(X, conv_weight)
+
+        if training:
```
albanD, line 137: Why handle this here if you don't in the previous example?

albanD: It would be simpler to just not handle the non-training case to simply fall back to the example above.

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,344 @@
...
+        100. * correct / len(test_loader.dataset)))
+
+use_cuda = torch.cuda.is_available()
+
+torch.manual_seed(123456)
+
+device = torch.device("cuda" if use_cuda else "cpu")
+
+train_kwargs = {'batch_size': 4096}
+test_kwargs = {'batch_size': 1024}
```
albanD, line 316: Why is the test batch size bigger than the train one? You should be able to test more since you don't save any buffer

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,344 @@
...
+    transforms.Normalize((0.1307,), (0.3081,))
+])
+dataset1 = datasets.MNIST('../data', train=True, download=True,
+                          transform=transform)
+dataset2 = datasets.MNIST('../data', train=False,
+                          transform=transform)
+train_loader = torch.utils.data.DataLoader(dataset1, **train_kwargs)
+test_loader = torch.utils.data.DataLoader(dataset2, **test_kwargs)
+
+model = Net(fused=True).to(device)
```
albanD, line 336: You could compare the fused model vs the non-fused one by measuring the peak memory usage for the whole training/testing with https://pytorch.org/docs/stable/generated/torch.cuda.max_memory_allocated.html?highlight=max_#torch.cuda.max_memory_allocated

albanD (commented):
Looks pretty good.
I didn't check the rendered version yet though.

intermediate_source/custom_function_conv_bn_tutorial.py
soulitzer: Yes for reuse - doing the same for conv

intermediate_source/custom_function_conv_bn_tutorial.py
soulitzer: Refactoring this section a bit

intermediate_source/custom_function_conv_bn_tutorial.py
soulitzer: This would be a big simplification indeed. Its probably fine since we can do batched testing here.

intermediate_source/custom_function_conv_bn_tutorial.py
soulitzer: Not sure I can do this one in-place because it would modify ctx.sum as well, which I wanted to save for backward

intermediate_source/custom_function_conv_bn_tutorial.py
soulitzer: This was what I needed!

intermediate_source/custom_function_conv_bn_tutorial.py
soulitzer: Tried it out a bit, and this seems a bit dangerous
- Modifying saved variable in-place means that it won't pass gradcheck - but may be that's expected for once differentiable functions.
- Modifying grad_out in-place seems okay on the surface, but we run into issues when the grad_out we get has memory overlap (this can happen if the next op is a sum, so backward is expand). Are there any other side effects you can think of?

Without modifying either X (the saved variable), or grad_out in-place, I can derive backward with a single extra NCHW-sized buffers, so at least for this example reducing another buffer does not reduce the peak memory usage anymore.
The bottleneck here should be the fact that we have two FusedBN and one FC layer saving large tensors and also grad_out being allocated. So in theory if we are the second FusedBN we are allowed 1 extra buffer anyway for backward.

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,338 @@
...
+# Batch Norm has two modes: training and eval mode. In training mode
+# the sample statistics are a function of the inputs, in eval mode,
+# we use the saved running statistics, which are not a function of the inputs
+# This makes non-training mode's backward significantly simpler. Below
+# We implement and test only the training mode case.
+def unsqueeze_all(t):
+    # Helper function to unsqueeze all the dimensions that we reduce over
+    return t[None, :, None, None]
+
+def batch_norm_backward(grad_out, X, sum, sqrt_var, N, eps, X_inplace=True):
```
soulitzer, line 83: Remove this

intermediate_source/custom_function_conv_bn_tutorial.py
albanD: You should never modify grad_out inplace!
And while you can modify saved variables, that means that you can only call backward once which is not great.

But there are a lot of possibilities for inplace. Things like `-((X - expand(sum) / N) * grad_out)` for example does 4 out of place op. You can do one out of place and 3 inplace.

intermediate_source/custom_function_conv_bn_tutorial.py
albanD: Yes sorry this comment is not pointing at the right line in the code.
The return statement below does contain some opportunities though

intermediate_source/custom_function_conv_bn_tutorial.py
soulitzer: Gotcha. I've rewritten it since the last review to do many things more things in-place. It should be only allocating a single extra buffer here now! Though it seems that in the backward at least, these optimizations can be done whether or not we specify the function to be once differentiable or not.

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,346 @@
...
+######################################################################
+# Testing with gradcheck
+a = torch.rand(1, 2, 3, 4, requires_grad=True, dtype=torch.double)
+torch.autograd.gradcheck(BatchNorm.apply, (a,), fast_mode=False)
+
+######################################################################
+# Now that the bulk of the work has been done, we can combine
+# them together. Note that in (1) we only save a single buffer
+# for backward, but this also means we recompute convolution forward
+# in (5). Also see that in (2), (3), (4), and (5), its the same
```
soulitzer, line 149: fix numbering

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,346 @@
...
+weight = torch.rand(5, 3, 3, 3, requires_grad=True, dtype=torch.double)
+X = torch.rand(10, 3, 7, 7, requires_grad=True, dtype=torch.double)
+torch.autograd.gradcheck(Conv2D.apply, (X, weight))
+
+######################################################################
+# Batch Norm has two modes: training and eval mode. In training mode
+# the sample statistics are a function of the inputs, in eval mode,
+# we use the saved running statistics, which are not a function of the inputs
+# This makes non-training mode's backward significantly simpler. Below
+# We implement and test only the training mode case.
```
soulitzer, line 79: Note to self to add punctuation and remove bad capitalization

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,346 @@
+# -*- coding: utf-8 -*-
+"""
+Fusing Convolution and Batch Norm using Custom Function
+=======================================================
+
+Fusing adjacent convolution and batch norm layers together is typically an
+inference-time optimization to improve run-time. It is usually achieved
+by eliminating the batch norm layer entirely and updating the weight
+and bias of the preceding convolution [0]. This technique is not applicable
+for training models, however.
```
albanD, line 10: Shouldn't however be at the beginning of the sentence he?

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,346 @@
...
+
+In this tutorial, we will show a different technique to fuse the two layers
+that can be applied during training. Rather than improved runtime, the
+objective of this optimization is to reduce memory usage.
+
+The idea behind this optimization is to see that both convolution and
+batch norm (as well as many other ops) need to save a copy of their input
+during forward for the backward pass. For large
+batch sizes, these saved inputs are responsible for much of your memory usage,
```
albanD, line 19: much -> most?

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,346 @@
...
+    return grad_X, grad_input
+
+class Conv2D(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, X, weight):
+        ctx.save_for_backward(X, weight)
+        return F.conv2d(X, weight)
+
+    # Use @once_differentiable by default unless we intend to double
+    # backward
```
albanD, line 61: nit this can be on the line above

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,346 @@
...
+For simplicity, in this tutorial we hardcode `bias=False`, `stride=1`, `padding=0`, `dilation=1`,
+and `groups=1` for Conv2D. For BatchNorm2D, we hardcode `eps=1e-3`, `momentum=0.1`,
+`affine=False`, and `track_running_statistics=False`. Another small difference
+is that we add epsilon in the denomator outside of the square root in the computation
+of batch norm.
+
+[0] https://nenadmarkus.com/p/fusing-batchnorm-and-conv/
+"""
+
+######################################################################
```
albanD, line 39: I don't know if you can add titles in this section. But it would be nice if we had different sections.

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,346 @@
...
+    #  2) 'grad of out wrt mean(X)' * 'grad of mean(X) wrt X'
+    #  3) 'grad of out wrt X in the numerator' * 'grad of X wrt X'
+    # We then rewrite the formulas to use as few extra buffers as possible
+    tmp = -((X - unsqueeze_all(sum) / N) * grad_out).sum(dim=(0, 2, 3))
+    d_denom = tmp / (sqrt_var + eps)**2  # d_denom = -num / denom**2
+    # It is useful to delete tensors when you no longer need them with `del`
+    # For example, we could've done `del tmp` here because we won't use it later
+    # In this case, it's not a big difference because tmp only has size of (C,)
+    # The important thing is avoid allocating NCHW-sized tensors unnecessarily
+    d_var = d_denom / (2 * sqrt_var)  # denom = torch.sqrt(var) + eps
```
albanD, line 99: I am not sure these lines don't actually allocate extra buffers?
Note that this is not a very strict requirement, but we're trying to show what to do to save memory, so I think we should make sure we see the biggest benefit possible.

albanD (commented):
Looks good!

The two main things are inplace ops and titles (if we can actually add some)

intermediate_source/custom_function_conv_bn_tutorial.py
soulitzer: It is possible!

intermediate_source/custom_function_conv_bn_tutorial.py
soulitzer: I think this one does, but `d_denom` only has size `(C,)` vs something like grad_input, X, grad_out which have  `(N, C, H, W)` so I figured it would be a good way to demonstrate that distinction as well (see code comment above). I end up only allocating one extra NCHW-sized tensor in the end (in addition to grad_out and X), which should be the theoretical best we can do given that we should not modify X or grad_out in-place. Had to pull off this weird hack as well - see below in step (3).

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,355 @@
...
+def batch_norm_backward(grad_out, X, sum, sqrt_var, N, eps):
+    # We use the formula: out = (X - mean(X)) / (sqrt(var(X)) + eps)
+    # in batch norm 2d's forward. To simplify our derivation, we follow the
+    # chain rule and compute the gradients as follows before accumulating
+    # them all into a final grad_input.
+    #  1) 'grad of out wrt var(X)' * 'grad of var(X) wrt X'
+    #  2) 'grad of out wrt mean(X)' * 'grad of mean(X) wrt X'
+    #  3) 'grad of out wrt X in the numerator' * 'grad of X wrt X'
+    # We then rewrite the formulas to use as few extra buffers as possible
+    tmp = -((X - unsqueeze_all(sum) / N) * grad_out).sum(dim=(0, 2, 3))
```
soulitzer, line 96: neg in-place after the sum

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,355 @@
...
+        N = X.numel() / X.size(1)
+        sqrt_var = torch.sqrt(var)
+        ctx.save_for_backward(X)
+        ctx.eps = eps
+        ctx.sum = sum
+        ctx.N = N
+        ctx.sqrt_var = sqrt_var
+        mean = sum / N
+        denom = sqrt_var + eps
+        return (X - unsqueeze_all(mean)) / unsqueeze_all(denom)
```
soulitzer, line 135: inplace

intermediate_source/custom_function_conv_bn_tutorial.py
```diff
@@ -0,0 +1,355 @@
...
+        else:
+            x = self.conv1(x)
+            x = self.bn1(x)
+        x = F.relu(x)
+        if self.fused:
+            x = self.convbn2(x)
+        else:
+            x = self.conv2(x)
+            x = self.bn2(x)
+        x = F.relu(x)
```
soulitzer, line 266: do this inplace

albanD (approved):
LGTM

Issue #1595, opened 2021-06-30
art transfer tutorial  uses incomplete model for training  


Hitansh159:
I read the [A neural Algorithm of artistic style](https://arxiv.org/abs/1508.06576) recently and while trying to implement I referred pytorch's tutorial but it used only top layers rather than full model as suggested by paper 

### Style image
![style image](https://user-images.githubusercontent.com/65333231/123945043-5bf7a480-d9bb-11eb-88d6-c4c7e9efb35d.png)

### Context image:
![context image](https://user-images.githubusercontent.com/65333231/123945174-7e89bd80-d9bb-11eb-94d6-fd44a76ac9a0.png)

## This is result from tutorial :
![Tutorial result](https://user-images.githubusercontent.com/65333231/123944829-23f06180-d9bb-11eb-8efa-f2854eb7fddb.png)
 

## This is the result after full model :
![my result](https://user-images.githubusercontent.com/65333231/123944625-ec81b500-d9ba-11eb-8692-aaea2c415c50.png)
 
As you can see my results captured context as well as style better than current version
the algorithm is expected to work with transfer learning from VGG19 model which will have 19 conv layers but the current tutorial only have 5 layers which is why results aren't up to the mark.

Changes:
- Changed in implantation of model 
- max pool to avg pool as suggested by author in paper   
- VGG takes 224*224 input image which is not the case with current tutorial.
- multiplying individual loss by a factor 
- due to normalisation on gram matrix style loss tends to zero     

I would like to contribute to this issue

Hitansh159:
I have also made an PR for it 
please review it and let me know if any changes are needed

PR #1596, opened 2021-06-30
changed implentation for model declaration for transfer learning

Hitansh159:
This PR is related to issue #1595 

Changes made are

-  Initially model just used top 5 layers that is change to use 19 layes as suggested in paper
-  Replaced maxPool with average pool as mentioned in paper 
                 'improves gradient flow and gives more appealing results'
-  Changed alpha beta ratio to 10^-4
-  Input size is changed to 224*224 as VGG take it as input dimensions
-  Individual loss is multiplied by weight depending upon total layers taken into consideration
-  Due to normalisation of gram matrix style losses tend to zero and the result is similar to content image


svekars:
@alexis-jacq I think you were the original author of this tutorial. Do you mind reviewing this change? Thanks!

PR #1602, opened 2021-07-08
[basics/quickstart_tutorial.py] call model.trian() before training

183amir:
Adds a `model.train()` call before training.
Because the `model.eval()` is called inside the `test()` function, the model trains in eval mode after the first eopch.

Even though `model.train()` has no effect here because the architecture does not contain batch norm or dropout layers,
A lot of people (myself included) copy code from this tutorial and calling `model.train()` before training is a good practice.

### DIFF
--- a/beginner_source/basics/quickstart_tutorial.py
+++ b/beginner_source/basics/quickstart_tutorial.py
@@ -139,6 +139,7 @@ optimizer = torch.optim.SGD(model.parameters(), lr=1e-3)
 
 def train(dataloader, model, loss_fn, optimizer):
     size = len(dataloader.dataset)
+    model.train()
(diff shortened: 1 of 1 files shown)


183amir:
The failed CI jobs are not related to the changes done here. It's because of a job timeout and an Internet connection error.

guoyejun:
https://github.com/pytorch/tutorials/issues/1624, I think it is better to accept this PR sooner, people will be confused by this quick start tutorial. (a new update may needed to explain in the document)

suraj813:
This was indeed a confusing bug, thank you @guoyejun and @183amir for calling it out and fixing it

suraj813 (approved):

PR #1603, opened 2021-07-08
Add double backward custom function tutorial

soulitzer:
Add `torchviz` to `requirements.txt`. Lets see if torchviz renders properly here


custom_function_double_backward.py
```diff
@@ -0,0 +1,206 @@
...
+######################################################################
+# When writing a custom autograd function to backward through twice,
+# it is important to know when operations performed in a custom function
+# are recorded by autograd, when they aren't, and most importantly, how
+# `save_for_backward` works with all of this.
+# 
+# Custom functions implicitly affects grad mode in two ways:
+# 1) During forward, autograd does not record any the graph for any
+#    operations performed within the forward function. When forward
+#    completes, the backward function of the custom function
```
albanD, line 23: nit: capital letter for Function

custom_function_double_backward.py
```diff
@@ -0,0 +1,206 @@
...
+# We can use torchviz to visualize the graph to see why this works
+import torchviz
+
+x = torch.tensor(1., requires_grad=True).clone()
+out = Square.apply(x)
+grad_x, = torch.autograd.grad(out, x, create_graph=True)
+
+# We can see that the gradient wrt to x, is itself a function of x (dout/dx = 2x)
+# And the graph of this function has been properly constructed
+torchviz.make_dot((grad_x, x, out), {"grad_x": grad_x, "x": x, "out": out})
```
albanD, line 72: Does this work?
On colab it only works if it is the last line of the cell :/

custom_function_double_backward.py
```diff
@@ -0,0 +1,206 @@
...
+# Because forward is performed in no-grad mode, if an intermediate result
+# of the forward pass is used to compute gradients in the backward pass
+# the backward graph of the gradients would not include the operations
+# that computed the intermediate result. This leads to incorrect gradients.
+class Sinh2(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, x):
+        expx = torch.exp(x)
+        expnegx = torch.exp(-x)
+        ctx.save_for_backward(expx, expnegx)
```
albanD, line 119: Do we want the "easy trap" case where someone saves expx on the ctx and show the two graphs? With a big warning on the Function that is wrong of course :D

custom_function_double_backward.py
```diff
@@ -0,0 +1,206 @@
...
+        x, = ctx.saved_tensors
+        return CubeBackward.apply(grad_out, x)
+
+class CubeBackward(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, grad_out, x):
+        ctx.save_for_backward(x, grad_out)
+        return cube_backward(grad_out, x)
+
+    @staticmethod
```
albanD, line 184: maybe make this one once differentiable so that triple gradient will properly fail under that assumption that the backward backward functions are not differentiable either.

albanD (approved):
The text looks good to me. Hopefully the rendering works as expected!

custom_function_double_backward.py
soulitzer: good point!

custom_function_double_backward.py
soulitzer: The graph should be the same right? So maybe we don't need an entire separate implementation shown. I think its good to mention this pitfall in a comment though.

albanD: No it won't 
In one, the Node will have only one output and only that will be used in backward.
In the second, you will see the two outputs and that both are actually properly linked in the double backward graph

soulitzer: Tried it out here: https://colab.research.google.com/drive/1sSVpP1Dmn4tYIXTCUmWJP3ON9gC4cT_v
The only case where I see it breaking is `ctx.result = torch.exp(x); return torch.exp(x)` which seems a little odd to do haha

albanD: Because your intermediary result is not an intermediary. I updated the notebook to make it fail.
And another custom function that is correct

soulitzer:
Looks like from the render that torchviz output does not display, so looks like we'll have to attach images ourselves.

 - Square:`image`
 - Exp: `image`
 - Sinh: `image`
 - Cube: `image`

custom_function_double_backward.py
soulitzer: ohhh crap I was thinking about the Exp function, didn't realize the comment was actually about Sinh

soulitzer:
@brianjo any idea what could be going wrong here? https://app.circleci.com/pipelines/github/pytorch/tutorials/4897/workflows/6908bc4d-4018-495c-ad17-88306d608a07/jobs/95127/steps. Looks like CUDA is running out of memory in an unrelated tutorial - yet we see master green https://app.circleci.com/pipelines/github/pytorch/tutorials?branch=master.

Could it just be because the master build machines have more cuda memory?

Just to be fully sure it is not our fault , I checked that we're still encountering the same failure with this commit even when I delete everything https://github.com/pytorch/tutorials/pull/1603/files/fc15597076b7ebd09965699f4d8808279101e11e. 

Maybe we can just manually land this?

brianjo:
> @brianjo any idea what could be going wrong here? https://app.circleci.com/pipelines/github/pytorch/tutorials/4897/workflows/6908bc4d-4018-495c-ad17-88306d608a07/jobs/95127/steps. Looks like CUDA is running out of memory in an unrelated tutorial - yet we see master green https://app.circleci.com/pipelines/github/pytorch/tutorials?branch=master.
> 
> Could it just be because the master build machines have more cuda memory?
> 
> Just to be fully sure it is not our fault , I checked that we're still encountering the same failure with this commit even when I delete everything https://github.com/pytorch/tutorials/pull/1603/files/fc15597076b7ebd09965699f4d8808279101e11e.
> 
> Maybe we can just manually land this?

Take a look at the readme.txt for information about submitting to the tutorials. This one looks like it's in root

> @brianjo any idea what could be going wrong here? https://app.circleci.com/pipelines/github/pytorch/tutorials/4897/workflows/6908bc4d-4018-495c-ad17-88306d608a07/jobs/95127/steps. Looks like CUDA is running out of memory in an unrelated tutorial - yet we see master green https://app.circleci.com/pipelines/github/pytorch/tutorials?branch=master.
> 
> Could it just be because the master build machines have more cuda memory?
> 
> Just to be fully sure it is not our fault , I checked that we're still encountering the same failure with this commit even when I delete everything https://github.com/pytorch/tutorials/pull/1603/files/fc15597076b7ebd09965699f4d8808279101e11e.
> 
> Maybe we can just manually land this?

No, if the pr is broken it will break the master build. Same machines. I'll see if I can rerun the test. Thanks!

soulitzer:
> Take a look at the readme.txt for information about submitting to the tutorials. This one looks like it's in root

Oh right, I realized my initial commit had the tutorial in root, but it should be in the correct directory now!

brianjo:
Hey @soulitzer, I just tried to run this on my local machine and I'm not seeing any output. Is there supposed to be some? If not, we should probably just make this an RST file so it renders as a document rather than a notebook.

soulitzer:
@brianjo good point, there is no output for this one, so rst is probably a better choice - though it is still strange why the test would be failing

soulitzer:
@brianjo Looks like tests are passing now! We should be good to go!

brianjo:
> @brianjo Looks like tests are passing now! We should be good to go!

Looks great! Thanks for your extra work on this. I'll merge it now. :)

PR #1618, opened 2021-07-27
Update ddp_tutorial.rst

bowangbj:
Relax GPU count to support devices with 2 GPUs

### DIFF
--- a/intermediate_source/ddp_tutorial.rst
+++ b/intermediate_source/ddp_tutorial.rst
@@ -265,8 +265,8 @@ either the application or the model ``forward()`` method.
         setup(rank, world_size)
 
         # setup mp_model and devices for this process
-        dev0 = rank * 2
-        dev1 = rank * 2 + 1
+        dev0 = (rank * 2) % world_size
+        dev1 = (rank * 2 + 1) % world_size
         mp_model = ToyMpModel(dev0, dev1)
         ddp_mp_model = DDP(mp_model)
 
@@ -285,9 +285,8 @@ either the application or the model ``forward()`` method.
 
     if __name__ == "__main__":
         n_gpus = torch.cuda.device_count()
-        if n_gpus < 8:
-          print(f"Requires at least 8 GPUs to run, but got {n_gpus}.")
-        else:
-          run_demo(demo_basic, 8)
-          run_demo(demo_checkpoint, 8)
-          run_demo(demo_model_parallel, 4)
+        assert n_gpus >= 2, f"Requires at least 2 GPUs to run, but got {n_gpus}"
+        world_size = n_gpus
+        run_demo(demo_basic, world_size)
+        run_demo(demo_checkpoint, world_size)
+        run_demo(demo_model_parallel, world_size)


bowangbj:
Hey Shen, Pritam, and Brian.

This is small change to make ddp tutorials accept dev server with >= 2 GPU instead of hard coding it to 8. PTAL

brianjo:
Preview looks good. https://deploy-preview-1618--pytorch-tutorials-preview.netlify.app/intermediate/ddp_tutorial.html I can merge anytime. Thanks Bo!

bowangbj:
> Preview looks good. https://deploy-preview-1618--pytorch-tutorials-preview.netlify.app/intermediate/ddp_tutorial.html I can merge anytime. Thanks Bo!

Thanks Brian a lot. Lets wait for Shen or Pritam to take a look before merging. Thanks a lot.

intermediate_source/ddp_tutorial.rst
```diff
@@ -265,8 +265,8 @@ either the application or the model ``forward()`` method.
         setup(rank, world_size)
 
         # setup mp_model and devices for this process
-        dev0 = rank * 2
-        dev1 = rank * 2 + 1
+        dev0 = (rank * 2) % world_size
```
mrshenli, line 268: Unfortunately, we cannot do this. This will make rank0 and rank 1 share the two GPUs, if there are just two GPUs. DDP and collective comm requires each process to exclusively work on GPUs, otherwise, the comm might hang.

mrshenli (changes requested):

brianjo (approved):

intermediate_source/ddp_tutorial.rst
bowangbj: The issue I got is like below:

I tried to run the code on 2 GPUs machine.
For rank0 -- dev0=0 dev1=1 
For rank1 --- dev0=2 dev2=3 

With that I got exception about wrong device ordinals since only cuda:0 and cuda:1 are valid.  But for rank2, it ends up with cuda:2 and cuda:3 

Any idea how to avoid such exception?

bowangbj:
Thanks Shen for your super quick review, please take a look at the inline comment on the error I got when running the example.

bowangbj:
Oops, there is not ready to merge. Sent out https://github.com/pytorch/tutorials/pull/1621/files to roll it back.

intermediate_source/ddp_tutorial.rst
mrshenli: This tutorial will require 8 GPUs to run, as highlighted in the note block in the summary section: "The code in this tutorial runs on an 8-GPU server, but it can be easily generalized to other environments."

Maybe we can error out when the number of GPUs is less than 8?

mrshenli: 2nd thought, I think only the model parallel part needs 8 GPUs. `demo_model_parallel` we can if just that one when the number of GPUs is less than 8

PR #1621, opened 2021-07-28
Rollback ddp_tutorial fix

bowangbj:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #1623
* __->__ #1621

Summary:

https://github.com/pytorch/tutorials/pull/1618 was merged unintentionally before all the cmts were resolved.
Also did a minior fix to skip demo parallel when ngpus is < 4

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/intermediate_source/ddp_tutorial.rst
+++ b/intermediate_source/ddp_tutorial.rst
@@ -265,8 +265,8 @@ either the application or the model ``forward()`` method.
         setup(rank, world_size)
 
         # setup mp_model and devices for this process
-        dev0 = (rank * 2) % world_size
-        dev1 = (rank * 2 + 1) % world_size
+        dev0 = rank * 2
+        dev1 = rank * 2 + 1
         mp_model = ToyMpModel(dev0, dev1)
         ddp_mp_model = DDP(mp_model)
 
@@ -285,7 +285,7 @@ either the application or the model ``forward()`` method.
 
     if __name__ == "__main__":
         n_gpus = torch.cuda.device_count()
-        assert n_gpus >= 2, f"Requires at least 2 GPUs to run, but got {n_gpus}"
+        assert n_gpus >= 8, f"Requires at least 2 GPUs to run, but got {n_gpus}"
         world_size = n_gpus
         run_demo(demo_basic, world_size)
         run_demo(demo_checkpoint, world_size)


intermediate_source/ddp_tutorial.rst
```diff
@@ -265,8 +265,8 @@ either the application or the model ``forward()`` method.
         setup(rank, world_size)
 
         # setup mp_model and devices for this process
-        dev0 = (rank * 2) % world_size
-        dev1 = (rank * 2 + 1) % world_size
+        dev0 = rank * 2
```
bowangbj, line 268: Copied the conversation from https://github.com/pytorch/tutorials/pull/1618:

Shen: 
Unfortunately, we cannot do this. This will make rank0 and rank 1 share the two GPUs, if there are just two GPUs. DDP and collective comm requires each process to exclusively work on GPUs, otherwise, the comm might hang.

Bo: 
The issue I got is like below:

I tried to run the code on 2 GPUs machine.
For rank0 -- dev0=0 dev1=1
For rank1 --- dev0=2 dev2=3

With that I got exception about wrong device ordinals since only cuda:0 and cuda:1 are valid. But for rank2, it ends up with cuda:2 and cuda:3

Any idea how to avoid such exception?

bowangbj:
Continued discussion of https://github.com/pytorch/tutorials/pull/1618 here.
Shen, Brian, PTAL

intermediate_source/ddp_tutorial.rst
mrshenli: Looks like only `demo_model_parallel` needs >= 4 GPUs. Maybe we can do the following?

1. skip `demo_model_parallel` when there are less than 4 GPUs
2. pass ngpus / 2 as world_size

mrshenli (commented):
Thanks for fixing!

intermediate_source/ddp_tutorial.rst
bowangbj: SG, changed as suggested

bowangbj:
Thanks Shen for the quick review, fixed as commented, PTAL

bowangbj:
Closing this PR which is subsumed by https://github.com/pytorch/tutorials/pull/1641

PR #1625, opened 2021-08-02
Add non-context manager tutorial for PyTorch Tensorboard plugin

guotuofeng:


brianjo (commented):
Mike, I'm seeing an error in the build. Can you try to run the file locally and make sure it's working as expected?

chaekit:
It seems like the CI is still failing?

guotuofeng:
> It seems like the CI is still failing?

It seems the tutorial not use the pytorch master code. Instead, it is using pytorch 1.9.0?

PR #1628, opened 2021-08-02
Update iOS GPU tutorial based on the lite interpreter support

husthyc:


xta0 (approved):
Thanks for updating the tutorial

Issue #1526, opened 2021-05-18
[Introduction to PyTorch] Bug on model definition part


sehyun03:
I found there's a bug on model definition part of following link.

https://pytorch.org/tutorials/beginner/basics/buildmodel_tutorial.html
https://pytorch.org/tutorials/beginner/basics/optimization_tutorial.html

The issue is on the following model definition line.

```
class NeuralNetwork(nn.Module):
    def __init__(self):
        super(NeuralNetwork, self).__init__()
        self.flatten = nn.Flatten()
        self.linear_relu_stack = nn.Sequential(
            nn.Linear(28*28, 512),
            nn.ReLU(),
            nn.Linear(512, 512),
            nn.ReLU(),
            nn.Linear(512, 10),
            nn.ReLU()
        )

    def forward(self, x):
        x = self.flatten(x)
        logits = self.linear_relu_stack(x)
        return logits
```

There's ReLU activation after the logit.
I think the final ReLU should be removed.

This part would be confusing for the beginners.

PR #1644, opened 2021-08-09
fix bug on model definition part of introduction

schnell3526:
This PR is related to issue https://github.com/pytorch/tutorials/issues/1526

A relu layer was inserted just before the cross entropy error was calculated, and it seemed that the loss was not calculated correctly, so I removed the relu layer.
Accordingly, the structure of the NeuralNetwork class in buildmodel_tutorial.py was also changed.

If you have time, I hope you will take a look.


suraj813:
Good catch! Thank you @sehyun03 and @schnell3526  for fixing this

suraj813 (approved):

PR #1657, opened 2021-08-23
Place evaluation data on device

Wesxdz:
The last part of the quickstart tutorial would not run by default and had the error
"Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu"
The issue could be particular to my local environment, but if this issue is applicable to others using CUDA , this is the solution


albanD (commented):
I am confused how this could work before if it fails now.
Also the fix should move classes and y around as well no?

PR #1664, opened 2021-08-30
Add text-to-speech tutorial

yangarbiter:


mthrok:
In tochaudio, we changed the way to instantiate pre-trained model, so the current state of this PR longer works. I will follow up with the updated version.

Issue #1666, opened 2021-09-03
Improvements for word vector tutorials.


Stevenjin8:
Hello!

I made a similar issue [here](https://github.com/pytorch/pytorch/issues/63728) but realized it was the wrong repo. There are just some unused variables in code examples.

PR #1667, opened 2021-09-03
Improvements to word embeddings tutorial.

Stevenjin8:
Following #1666, I think `CONTEXT_SIZE` should be used to preprocess data in the code examples. Also, I think it would be a good idea to lowercase/remove punctuation in the sample texts.

### DIFF
--- a/beginner_source/nlp/word_embeddings_tutorial.py
+++ b/beginner_source/nlp/word_embeddings_tutorial.py
@@ -208,10 +208,15 @@ This were to be new made when thou art old,
 And see thy blood warm when thou feel'st it cold.""".split()
 # we should tokenize the input, but we will ignore that for now
 # build a list of tuples.  Each tuple is ([ word_i-2, word_i-1 ], target word)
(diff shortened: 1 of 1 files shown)


beginner_source/nlp/word_embeddings_tutorial.py
```diff
@@ -208,10 +208,15 @@
 And see thy blood warm when thou feel'st it cold.""".split()
 # we should tokenize the input, but we will ignore that for now
 # build a list of tuples.  Each tuple is ([ word_i-2, word_i-1 ], target word)
-trigrams = [([test_sentence[i], test_sentence[i + 1]], test_sentence[i + 2])
-            for i in range(len(test_sentence) - 2)]
-# print the first 3, just so you can see what they look like
-print(trigrams[:3])
+ngrams = [
```
albanD, line 211: The comment above would need to be updated no?

Stevenjin8: yes

albanD (approved):
LGTM

PR #1675, opened 2021-09-13
Update quickstart_tutorial.py

puhuk:
Change iterable to iterator which seems to more clear.


brianjo:
I found [this description](https://www.geeksforgeeks.org/python-difference-iterable-iterator/#:~:text=Iterable%20is%20an%20object%2C%20which%20one%20can%20iterate%20over.&text=Iterator%20is%20an%20object%2C%20which,every%20iterable%20is%20an%20iterator.) on the net, but I'm not positive what applies here. @albanD thoughts?

albanD (changes requested):
My understanding is that the old naming is correct. What the DataLoader wraps is an object on which you can iterate by calling iter(). Not an iterator.
The Dataloader that you get is also an iterable and the iterator that is actually used is the `_BaseDataLoaderIter` instance that gets created when you try to iterate over the dataloader.

puhuk:
Thanks for feedback.
Let me close this PR and related issue :)

PR #1677, opened 2021-09-13
Add autograd not implemented to dispatcher tutorial

soulitzer:
- @albanD has mentioned that maybe `Fallback` isn't the best name since its not a real "fallback". It is probably best not to overload the the term, and instead have something like `REGISTER_AUTOGRAD_NOT_IMPLEMENTED_KERNEL`.
 - The exact API is not 100% determined, but starting to write this doc can help us think about what it will look like.
 - Any other places where this new macro can be linked?

### DIFF
--- a/advanced_source/dispatcher.rst
+++ b/advanced_source/dispatcher.rst
@@ -105,6 +105,50 @@ speaking, the structure of your registrations will look like this:
     that provides implementations for all basic operators on the XLA dispatch
     key.
 
+
+For operators that do not need autograd
+---------------------------------------
+
+In the next section, we will discuss how to add autograd support to an operator.
+But for the ops that do not need autograd support, the following line can be
+added to improve useability and make your op behave like PyTorch's built-in
+operators.
+
+.. code-block:: cpp
+
+  REGISTER_AUTOGRAD_NOT_IMPLEMENTED_FALLBACK(myops, "myadd");
+
+Including the above line will an register an Autograd kernel that appends a dummy
+`NotImplemented` node on forward (preserving the `require_grad`-ness of
+the inputs). On backward, the `NotImplemented` node raises an error. This
+can be helpful for debugging in larger models where previously it can be hard
+to pin-point exactly where the `requires_grad`-ness is lost during the forward pass.
+
+This macro **always** registers both the Autograd and ADInplaceOrView kernels
+whether or not your operator is view or inplace. As a consequence,
+**there may be an additional overhead for non-view-or-inplace ops in
+inference mode when the input is non-inference.** If your operator is neither
+view nor in-place, and this is an issue for your use case, consider only
+registering the Autograd kernel:
+
+.. code-block:: cpp
+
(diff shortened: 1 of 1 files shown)


albanD (commented):
Thanks for that update that looks quite nice!
I'll let @ezyang give his opinion on that as well.

soulitzer:
Hey @brianjo this is not quite ready to land yet. I forgot to make that more clear, sorry!
Is it safe to revert this?

advanced_source/dispatcher.rst
```diff
@@ -105,6 +105,50 @@ speaking, the structure of your registrations will look like this:
     that provides implementations for all basic operators on the XLA dispatch
     key.
 
+
+For operators that do not need autograd
```
soulitzer, line 109: Note to self: mark that this is 3.10(?) only

advanced_source/dispatcher.rst
```diff
@@ -105,6 +105,50 @@ speaking, the structure of your registrations will look like this:
...
+whether or not your operator is view or inplace. As a consequence,
+**there may be an additional overhead for non-view-or-inplace ops in
+inference mode when the input is non-inference.** If your operator is neither
+view nor in-place, and this is an issue for your use case, consider only
+registering the Autograd kernel:
+
+.. code-block:: cpp
+
+  TORCH_LIBRARY_IMPL(myops, Autograd, m) {
+    m.def(op, autogradNotImplementedFallback());
```
ezyang, line 137: This doesn't look right; should you be doing an `m.impl` here?

advanced_source/dispatcher.rst
```diff
@@ -105,6 +105,50 @@ speaking, the structure of your registrations will look like this:
...
+---------------------------------------
+
+In the next section, we will discuss how to add autograd support to an operator.
+But for the ops that do not need autograd support, the following line can be
+added to improve useability and make your op behave like PyTorch's built-in
+operators.
+
+.. code-block:: cpp
+
+  REGISTER_AUTOGRAD_NOT_IMPLEMENTED_FALLBACK(myops, "myadd");
```
ezyang, line 119: I'm not a big fan of the macro API here. (1) It's not namespaced with TORCH, so it is potentially ambiguous with other macros, and (2) it's pretty short to do this without a macro, right? So it seems better to not do the macro and make people write it out.

ezyang: I also still think that we ought to be registering a global fallback for all operators that defaults us to assuming that autograd is not implemented...

soulitzer: > it's pretty short to do this without a macro, right? So it seems better to not do the macro and make people write it out.

It will be like 6 lines, so not too bad. I'd be okay with that. Alternatively, we'd prefix this macro with `TORCH_`.

> I also still think that we ought to be registering a global fallback for all operators that defaults us to assuming that autograd is not implemented

Yeah this seems reasonable. Creating an issue to discuss the implications of this in more detail...

Also, since this has been merged (accidentally), I will make the changes to the tutorial in a separate PR

PR #1679, opened 2021-09-15
Update profiler tutorial for adding GLOO support

guyang3532:
In latest pytorch1.10, distributed backend Gloo is supported for profiler.


brianjo:
Thanks @guyang3532. I've labeled it for the 1.10 release.

PR #1682, opened 2021-09-15
Add a "Export/Load Model in TorchScript Format" section in the saving/loading models tutorial

leopoldmaillard:
The [Saving and loading models](https://pytorch.org/tutorials/beginner/saving_loading_models.html) tutorial is the preferred destination for beginners who want to load a saved model.

However, it does not mention the TorchScript format that enables to load a model for inference without defining its class in the code. It is a requested feature (see [here](https://stackoverflow.com/questions/59287728/saving-pytorch-model-with-no-access-to-model-class-code) and [here](https://discuss.pytorch.org/t/loading-pytorch-model-without-a-code/12469/7)) and TorchScript seems to be the best way to achieve it. From my own experience, the existing tutorial can be confusing for users coming from TensorFlow whose ```.pb```format enables to run inference without access to the model class.

TorchScript is great and I think it is worth mentioning in the beginners tutorials. I tried to make the new section as simple as possible but do not hesitate to request more details.

This is my very first contribution, any feedback would be much appreciated !

Have a nice day 😃


leopoldmaillard:
@brianjo in case you've missed it 😃

PR #1685, opened 2021-09-16
Update dispatcher tutorial for autograd not implemented kernel

soulitzer:
Address @ezyang's comments from https://github.com/pytorch/tutorials/pull/1677

 - `def` -> `impl`
 - macro -> manually write it out

This update actually makes it easier for users to avoid the trap of registering the ADInplaceOrView kernel by default which could cause perf issues.


ezyang (approved):

ezyang:
tbh the UX here is not very good and worth spending some time on, though this doesn't block the PR

PR #1734, opened 2021-10-27
Fix thumbnail aspect ratio

mthrok:
Replacing the thumbnail image for ASR tutorial so that it will be rendered properly.


brianjo:
@mthrok Moto, could you look at this and fix the conflict if needed?

PR #1739, opened 2021-11-05
Convert docstring to literal and fix math rendering typo

weston-barger:
https://github.com/pytorch/pytorch/issues/67772#issue-1043805621

Added missing \ to `\pi`. 

Converted docstring of `beginner_source/examples_autograd/polynomial_custom_function.py ` to literal because \f and \r were being resolved as escapes before math rendering.


soulitzer (approved):
Looks good, thanks for the fix!

weston-barger:
Thanks @soulitzer! This is my first github pull request :). Is there anything else that I need to do?

soulitzer:
Nope, I think we should be good to go. Just waiting for the tests to run again before it can be merged.

PR #1746, opened 2021-11-18
Add forward AD tutorial

soulitzer:


intermediate_source/forward_ad_tutorial.py
```diff
@@ -0,0 +1,167 @@
...
+    def jvp(ctx, gI):
+        gO = gI * ctx.result
+        # If the tensor stored in ctx will not also be used in the backward pass,
+        # one can manually free it using ``del``
+        del ctx.result
+        return gO
+
+fn = Fn.apply
+
+primal = torch.randn(10, 10, dtype=torch.double, requires_grad=True)  # Fix this?
```
soulitzer, line 149: Gradcheck should not require inputs to require grad if we are not checking backward AD. See:  https://github.com/pytorch/pytorch/pull/69004

albanD: Not sure this is true.
We do use requires_grad to know which input/outputs are differentiable.

Maybe we should not, but today we do require this, and just removing the check doesn't sound like the right solution.

intermediate_source/forward_ad_tutorial.py
```diff
@@ -0,0 +1,166 @@
...
+    # considered to have a zero-filled tangent of the same shape.
+    plain_tensor = torch.randn(10, 10)
+    dual_output = fn(dual_input, plain_tensor)
+
+    # Unpacking the dual returns a namedtuple with ``primal`` and ``tangent``
+    # as attributes
+    jvp = fwAD.unpack_dual(dual_output).tangent
+
+assert fwAD.unpack_dual(dual_output).tangent is None
+output = fwAD.unpack_dual(dual_output)
```
albanD, line 68: What is this line showing? Should we have another line below?

intermediate_source/forward_ad_tutorial.py
```diff
@@ -0,0 +1,166 @@
...
+    # primal with another tensor of the same size, which we call the tangent.
+    # If the layout of the tangent is different from that of the primal,
+    # The values of the tangent are copied into a new tensor with the same
+    # metadata as the primal. Otherwise, the tangent itself is used as-is.
+    #
+    # It is also important to note that the dual tensor created by
+    # ``make_dual`` is a view of the primal.
+    dual_input = fwAD.make_dual(primal, tangent)
+    assert dual_input._base is primal
+    assert fwAD.unpack_dual(dual_input).tangent is tangent
```
albanD, line 56: Should we have another example with a transposed tangent to show that the tangent is not always being re-used?

intermediate_source/forward_ad_tutorial.py
```diff
@@ -0,0 +1,166 @@
...
+# `nn.Parameter`s. As a workaround, one must register the dual tensor
+# as a non-parameter attribute of the module.
+
+import torch.nn as nn
+
+model = nn.Linear(5, 5)
+input = torch.randn(16, 5)
+
+params = {name: p for name, p in model.named_parameters()}
+tangents = {name: torch.rand_like(p) for name, p in params.items()}
```
albanD, line 85: Do you really need to create these ahead of time? You can just loop over `model.named_parameters()` below no?

intermediate_source/forward_ad_tutorial.py
```diff
@@ -0,0 +1,166 @@
+# -*- coding: utf-8 -*-
+"""
+Forward-mode Auto Differentiation
```
albanD, line 3: Automatic Differentiation or Auto Diff or AD

intermediate_source/forward_ad_tutorial.py
```diff
@@ -0,0 +1,166 @@
...
+in a Jacobian-vector product). When a input, which we call "primal", is
+associated with a "direction" tensor, which we call "tangent", the
+resultant new tensor object is called a "dual tensor" for its connection
+to dual numbers[0].
+
+As the forward pass is performed, if any input tensors are dual tensors,
+extra computation is performed to propogate this "sensitivity" of the
+function.
+
+[0] https://en.wikipedia.org/wiki/Dual_number
```
albanD, line 25: This should be at the bottom no?

intermediate_source/forward_ad_tutorial.py
soulitzer: You're correct about the inputs, but looks like gradcheck currently doesn't actually need the outputs to have requires grad:
```python
import torch
import torch.autograd.forward_ad as fwAD

def detach_backward(x):
  _, tangent = fwAD.unpack_dual(x)
  if not tangent:   # non-dual-level friendly path
    return x.detach()
  return fwAD.make_dual(x.detach(), tangent)

class Func(torch.autograd.Function):
  @staticmethod
  def forward(ctx, x):
    return x * 2

  @staticmethod
  def jvp(ctx, gI):
    return gI * 3  # Wrong, lets see if gradcheck catches this

t = torch.tensor(1., requires_grad=True, dtype=torch.double)

def fn(x):
  out1 = detach_backward(Func.apply(x))
  out2 = x.clone()  # only necessary to bypass the no differnetiable outputs check
  return out1, out2

with fwAD.dual_level():
  dual = fwAD.make_dual(t, torch.rand_like(t))
  out = fn(dual)
  assert not out[0].requires_grad

torch.autograd.gradcheck(fn, (t,), check_forward_ad=True,
  check_backward_ad=False, check_batched_grad=False, check_undefined_grad=False)
```

I don't know whether the fact non-require-grad outputs are checked is useful though. At least for custom functions, it looks like the output still requires_grad even if only `jvp` is implemented, i..e, does a function that behaves like `detach_backward` actually exist in nature?

albanD: > does a function that behaves like detach_backward actually exist in nature?

Ignoring things like `detach`, I think the answer is no. If a function is differentiable, it will be for both forward and backward.

intermediate_source/forward_ad_tutorial.py
soulitzer: I tried this at first, but ran into the issue of deleting entries while iterating

intermediate_source/forward_ad_tutorial.py
soulitzer: Hmm, don't remember. I will just remove it for now lol

albanD:
Note that this needs an API that is only in master so the build will fail.
You can rename the file so that it doesn't end with "_tutorial" such that it won't run. We can do a follow up PR for 1.11 that moves it back to make it run again.

albanD (approved):
sgtm

PR #1758, opened 2021-12-06
Fix forward AD tutorial directory and add to index

soulitzer:
Currently this tutorial is not being rendered at all. This PR places the tutorial in correct directory, adds it to the 'frontend API' index. We also note in the title/description that forward AD is beta (for now), so that users won't be mislead by the current lack of coverage.

After this merges, we can create another PR that updates the name (marked with the correct label).


albanD (approved):
SGTM
CI failure is unrelated

PR #1772, opened 2021-12-20
Fix quantization tutorials (imports, syntax, and style)

andrewor14:
Summary: This commit fixes the quantization tutorials such that
they can be run smoothly by the user.

Test Plan: Ran the updated tutorials without problem.

Reviewers: jerryzh168

Subscribers: jerryzh168, supriyar


jerryzh168 (approved):

prototype_source/fx_graph_mode_ptq_static.rst
```diff
@@ -294,6 +291,7 @@ Utility functions related to ``qconfig`` can be found in the `qconfig <https://g
   
 .. code:: python  
   
+    from torch.ao.quantization.quantize_fx import prepare_fx
```
jerryzh168, line 294: should we move imports to the beginning?

prototype_source/fx_graph_mode_ptq_static.rst
```diff
@@ -75,7 +75,7 @@ Download the `torchvision resnet18 model <https://github.com/pytorch/vision/blob
     import sys  
     import torch.quantization
```
jerryzh168, line 76: we should probably remove torch.quantization now

jerryzh168: also in L95 of this file:
```
from torch.quantization import get_default_qconfig, quantize_jit
```
we should remove quantize_jit

prototype_source/fx_graph_mode_ptq_static.rst
```diff
@@ -75,7 +75,7 @@ Download the `torchvision resnet18 model <https://github.com/pytorch/vision/blob
     import sys  
     import torch.quantization   
 
-    # Setup warnings    
+    # Set up warnings
     import warnings 
     warnings.filterwarnings(    
         action='ignore',
```
jerryzh168, line 81: please also update all `torch.quantization` in this file to `torch.ao.quantization`, another example is in L87

advanced_source/dynamic_quantization_tutorial.py
```diff
@@ -98,9 +98,9 @@ def __len__(self):
 class Corpus(object):
     def __init__(self, path):
         self.dictionary = Dictionary()
-        self.train = self.tokenize(os.path.join(path, 'train.txt'))
-        self.valid = self.tokenize(os.path.join(path, 'valid.txt'))
-        self.test = self.tokenize(os.path.join(path, 'test.txt'))
+        self.train = self.tokenize(os.path.join(path, 'wiki.train.token'))
+        self.valid = self.tokenize(os.path.join(path, 'wiki.valid.token'))
+        self.test = self.tokenize(os.path.join(path, 'wiki.test.token'))
```
malfet, line 103: Why are you doing those changes? Where those files are supposed to be coming from?

malfet (changes requested):
This breaks CI, namely because it references non-existing data files

advanced_source/dynamic_quantization_tutorial.py
andrewor14: These were the paths when I downloaded the dataset myself. Without these changes I was not able to run the tutorials. I guess CI downloads the files differently somehow

andrewor14: To unblock this PR maybe we should just revert these changes and investigate separately. How does that sound @jcaip?

malfet: Hmm, downloaded from where? For tutorials build, download locations are defined in (plan to move them to json file, but this is where they are for now)
https://github.com/pytorch/tutorials/blob/4918fe4bf3df96b233ca134c985ffb09ea45f92c/Makefile#L93-L94

jcaip: I was using a different wikitext version which has changed the naming convention of those files, which is causing the error. I will update the internal tutorial guide and revert the filepaths, which I think should fix everything. Thanks for the help @malfet!

PR #1761, opened 2021-12-09
Update forward AD tutorial to use _tutorial suffix for 1.11

soulitzer:
Do not land until 1.11

PR #1776, opened 2021-12-21
Update forward-ad tutorial to mention the functional API

zou3519:
I'm writing tutorials for functorch. Right now the plan is to (1) put a
mention of functorch.jvp into the PyTorch forward-mode AD tutorial (to
consolidate information and so that users aren't confused about why
there are "two ways" or "two tutorials" on how to do forward-mode AD)
and (2) write up a "various higher-order autodiff use cases" tutorial.

This PR implements (1).

Test Plan:
- I tested this locally.


intermediate_source/forward_ad_usage.py
```diff
@@ -172,5 +172,75 @@ def jvp(ctx, gI):
                          check_backward_ad=False, check_undefined_grad=False,
                          check_batched_grad=False)
 
+######################################################################
+# Functional API
```
soulitzer, line 176: Should users be concerned about these APIs changing?

intermediate_source/forward_ad_usage.py
```diff
@@ -172,5 +172,75 @@ def jvp(ctx, gI):
...
+# --------------------------------------------------------------------
+# We also offer a higher-level functional API in functorch
+# for computing Jacobian-vector products that you may find simpler to use
+# depending on your use case.
+#
+# The benefit of the functional API is that there isn't a need to understand
+# or use the lower-level dual tensor API and that you can compose it with
+# other functorch transforms (like vmap); the downside is that it offers you
```
soulitzer, line 184: I suppose that will be covered in the advanced tutorial? Maybe we want to link to that here, once that is ready.

intermediate_source/forward_ad_usage.py
```diff
@@ -172,5 +172,75 @@ def jvp(ctx, gI):
...
+y = torch.randn(10, 10)
+
+import functools
+new_fn = functools.partial(fn, y=y)
+primal_out, tangent_out = ft.jvp(new_fn, (primal,), (tangent,))
+
+######################################################################
+# Using the functional API with Modules
+# --------------------------------------------------------------------
+# To use ``nn.Module`` with functorch.jvp to compute jvps with respect to the
```
soulitzer, line 220: nit: JVPs (or writing it out as Jacobian-vector product?) just so that its different than lower case `jvp`, the API

soulitzer (approved):
Thanks for the addition! Added some small comments

intermediate_source/forward_ad_usage.py
```diff
@@ -172,5 +172,75 @@ def jvp(ctx, gI):
...
+# The benefit of the functional API is that there isn't a need to understand
+# or use the lower-level dual tensor API and that you can compose it with
+# other functorch transforms (like vmap); the downside is that it offers you
+# less control.
+#
+# Note that the remainder of this tutorial will require functorch
+# (https://github.com/pytorch/functorch) to run. Please find installation
+# instructions at the specified link.
+
+import functorch as ft
```
soulitzer, line 191: I think this means that I will need to add `functorch` as a dependency in https://github.com/pytorch/tutorials/pull/1761. Or you could also add it in this PR I suppose, but this file isn't being run anyway as of yet.

intermediate_source/forward_ad_usage.py
zou3519: The APIs are technically "beta" but I don't see us changing the API in a BC-breaking way at all because that's what JAX offers. I can put "(beta)" next to it to make it clearer

intermediate_source/forward_ad_usage.py
zou3519: Compositions of transforms will be covered in the other tutorial (or tutorials) that I am writing, I'm not sure how many. I can link to them when they're done

intermediate_source/forward_ad_usage.py
zou3519: Good catch. Yes, we'll need to add functorch as a dependency

intermediate_source/forward_ad_usage.py
```diff
@@ -172,5 +172,75 @@ def jvp(ctx, gI):
...
+primal1 = torch.randn(10, 10)
+tangent1 = torch.randn(10, 10)
+
+def fn(x, y):
+    return x ** 2 + y ** 2
+
+# Here is a basic example to compute the JVP of the above function.
+# The jvp(func, primals, tangents) returns func(*primals) as well as the
+# computed jvp. Each primal must be associated with a tangent of the same shape.
+primal_out, tangent_out = ft.jvp(fn, (primal0, primal1), (tangent0, tangent1))
```
albanD, line 204: I am bit confused on the functorch APIs which ones return functions vs eagerly compute the result.
Is there any doc that explain why both are used depending on the function?

zou3519: - jvp eagerly computes the result
- vjp eagerly computes the output and returns a vjp_fn
- All the other APIs (grad, vmap, jacrev, jacfwd, hessian) return a function
See https://pytorch.org/functorch/functorch.html for more details.

The main rationale for this is to align the API with JAX's so that it is easy to port code. In general the design philosophy is:
- return a function when possible
- for vjp, it doesn't make sense to return a function that computes the output AND the vector-jacobian-product, because if we were to run this function multiple times we would only need to compute the output once
- for jvp: I don't have a good reason why it is like this.

albanD: Consistency with jax sounds good.
Is it also consistent with hips/autograd?

zou3519: I'm not familiar with HIPS/autograd. Reading their README it looks like they only support one operation, `grad`, and that it is higher order (and consistent with jax.grad and functorch.grad): https://github.com/HIPS/autograd

brianjo:
Hold until Mar 11.

intermediate_source/forward_ad_usage.py
zou3519: Dependency added.

zou3519:
Looks good to me now:
image

zou3519:
> https://app.circleci.com/pipelines/github/pytorch/tutorials/5727/workflows/2b735c50-3f85-4c81-b6d2-0a25cd50a8a3/jobs/112461 

This looks flaky, do I rebase this before attempting to merge?

brianjo:
> > https://app.circleci.com/pipelines/github/pytorch/tutorials/5727/workflows/2b735c50-3f85-4c81-b6d2-0a25cd50a8a3/jobs/112461
> 
> This looks flaky, do I rebase this before attempting to merge? @brianjo

Yeah, I needed to merge a fix, so I rebased again on that. If it's clean I'll merge it. Thanks!

PR #1779, opened 2021-12-26
Fixed moving tensor to GPU when not supported

pranavgade20:
I have an old GPU, so although I have CUDA, torch does not support it. So although `torch.cuda.is_available()` returns True, moving the tensor to GPU causes errors later when I try to read/write to it. I did some poking around, and it looks like `(torch.device('cuda').index is not None)` returns false for me. Alternatively, hasattr(torch, 'cuda_version') returns false as well. I'm not sure if this is the best(or correct) way to figure out cuda compatibility though.


albanD:
Hi,

Wouldn't it be simpler for you to just hide your GPU from the process altogether by having the following env variable `CUDA_VISIBLE_DEVICES=''` ?

pranavgade20:
Yup, but considering that this is a 'beginner' tutorial, there should be some way of determining if your GPU works so that people don't waste time(like I did) being very confused about why it fails in a weird way. i.e., a comment somewhere that mentions cases where you'd like to `CUDA_VISIBLE_DEVICES=''` would be incredibly helpful.

albanD:
Agreed, what about something like:
```python
# We move our tensor to the GPU if available
if torch.cuda.is_available():
    if torch.device('cuda').index is None:
        raise RuntimeError("You have cuda installed but this version is too old and not supported by "
                                         "PyTorch binary. You can hide this device from PyTorch by setting the env "
                                         "variable CUDA_VISIBLE_DEVICES='' before running this script.")
    tensor = tensor.to('cuda')
```

Does that sounds better?

pranavgade20:
Yep, LGTM!

albanD:
Perfect! Could you update this PR?

albanD (approved):
Thanks!

@brianjo good to go!

brianjo:
> Thanks!
> 
> @brianjo good to go!

Thanks Alban. I'm going to make sure this builds cleanly and then I'll merge it. :)

suraj813 (approved):

suraj813 (changes requested):
A) `torch.device('cuda').index` will always return None. `torch.device('cuda:0').index` returns 0.

B) I'm not sure the condition `torch.device('cuda:0').index is None` is appropriate for what you're trying to do, as it doesn't specifically check for incompatibility in CUDA versions. 

I can't check this since I don't have an older CUDA machine. Do you mind sharing a screenshot where `torch.cuda.is_available() == True` and `torch.device('cuda:0').index is None`?

albanD:
It is always None indeed.
I guess a good old try/except would be better here?

PR #1811, opened 2022-01-26
set cudnn.benchmark to True for speed up training process

developer0hye:
set cudnn.benchmark to True for speed up training process

I followed other official pytorch example code.
https://github.com/pytorch/examples/blob/master/imagenet/main.py#L200


albanD (approved):
Thanks for the update!

brianjo:
> 

Thanks for the review Alban!

PR #1815, opened 2022-02-03
update 'TODO link' with working link for Conv-BN FX Fusion tutorial

lessw2020:
updating the residual TODO: link to working, intended link for the BN w FX tutorial.


lessw2020:
Before:
torch_fx_todo_link

**After**:
bn_fx_link_fixed

I'm not clear that the circleCI  failure has anything to do with the actual fix.

jamesr66a (approved):
Awesome, thanks! The failure does indeed look unrelated, but cc @brianjo (or whoever knows best about this) to look

PR #1816, opened 2022-02-03
Update awscli dep

saulshanabrook:
This PR updates the awscli dependency so that this project can be installed alongside a recent version of rich, which requires a more recent version of colorama than was allowed by the old awcli version.

I choose the oldest version of awscli that is compatible, to be minimally invasive.


saulshanabrook:
Ah I see that this is then incompatible with docutils 0.16. Even the most recent version of awscli requires `docutils>=0.10,<0.16` which is incompatible with install command.

PR #1831, opened 2022-02-23
normalize pathnames to fix issue 1830

### LINKED ISSUE #1830: test and val dataset included in train in speech_command_classification_with_torchaudio_tutorial.py
In [speech_command_classification_with_torchaudio_tutorial.py](https://github.com/pytorch/tutorials/blob/master/intermediate_source/speech_command_classification_with_torchaudio_tutorial.py), the test and validation datasets are included in the training dataset.

When doing:
https://github.com/pytorch/tutorials/blob/bb3523fe06116c6671319d476f5d2725d0406c58/intermediate_source/speech_command_classification_with_torchaudio_tutorial.py#L95-L96

And checking for intersection of speakers:
```python
train_speaker_ids = set([ s[3] for s in train_set])
test_speaker_ids = set([ s[3] for s in test_set])
intersection_of_speakers = train_speaker_ids & test_speaker_ids
print(len(train_speaker_ids))
print(len(test_speaker_ids))
print(len(intersection))
```
It gives

```
2618
250
250
```
Hence the 250 speakers in test are in the training dataset. Same for thing for validation set.

algat:
fixes #1830 

With this commit, train, test and validation dataset are disjoint:

Verification: when doing
https://github.com/pytorch/tutorials/blob/bb3523fe06116c6671319d476f5d2725d0406c58/intermediate_source/speech_command_classification_with_torchaudio_tutorial.py#L95-L96

And checking for intersection of speakers:
```python
train_speaker_ids = set([ s[3] for s in train_set])
test_speaker_ids = set([ s[3] for s in test_set])
intersection_of_speakers = train_speaker_ids & test_speaker_ids
print(len(train_speaker_ids))
print(len(test_speaker_ids))
print(len(intersection))
```
It gives

```
2618
250
0
```
Which is what we want.


mthrok (approved):
@algat Thanks for the fix. It looks good to me.

Even a better solution is to simply use the `subset` argument of the constructor https://pytorch.org/audio/stable/datasets.html#speechcommands
The argument did not exist when the tutorial was firstly introduced.

PR #1847, opened 2022-03-08
Updated text classification tutorial to work with torchtext 0.12.0 release

Nayef211:
- Added `iter` around `AG_NEWS` dataset call since datapipes don't implement a `next` method
- Updated `train_iter` output when calling `next`
- Update vocab transform output


parmeet (approved):
Overall LGTM! I just changed the base to 1.11-RC-TEST to test changes with the latest since now we have dependency on torchdata cc: @brianjo

Nayef211:
@brianjo I'm wondering if you guys can add torchdata as a dependency for the tutorials repo since torchtext datasets, starting from `0.12.0`, require torchdata in order to function.

PR #1850, opened 2022-03-08

yaxuan1102:
I added an if...else... condition under `filterPair()` to determine whether 'eng' would be the input language or not, so that the code would also work well under the setting of `reverse=False`.
Please consider the proposed changes if they are reasonable :)


svekars:
@parmeet  @hudeven can you please take a look?

PR #1851, opened 2022-03-09
Fixing timing scheme in cpp_extension tutorial

stevenwalton:
Print statement incorrectly stated `us` when it was in `100ms` units.
For clarity I changed it to `s` units and removed the multiplication.
This could have intended to be averaged but total time works just as
well.

### DIFF
--- a/advanced_source/cpp_extension.rst
+++ b/advanced_source/cpp_extension.rst
@@ -688,7 +688,7 @@ We'll start with the C++ file, which we'll call ``lltm_cuda.cpp``, for example:
 
   // C++ interface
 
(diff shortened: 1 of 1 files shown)


stevenwalton:
Note here the second commit. I don't actually get warnings when compiling the code using `gates.type()` but the define directive said that `type()` was depreciated. I can make the changes to `option()` if that is actually the correct default.

mrshenli (approved):
LGTM!

mrshenli:
Hey @brianjo 

The "urllib.error.HTTPError: HTTP Error 500: Internal Server Error" error seems irrelevant to this PR. Shall we merge?

brianjo:
Let me try to kick that worker first. :)

PR #1857, opened 2022-03-12
Remove awscli requirement

malfet:
Not sure why its needed


mthrok:
Actually, torchaudio I/O tutorial uses it. https://pytorch.org/tutorials/beginner/audio_preprocessing_tutorial.html Should we remove that portion of tutorial?

PR #1859, opened 2022-03-12
add explicit instructions to install torchdata

parmeet:
torchtext datasets starting from version 0.12 depend on TorchData. For 0.12 release, installing torchtext doesn't automatically install torchdata as it is an optional dependency, hence providing explicit message to install torchdata for all the tutorials that access torchtext datasets.


Nayef211 (approved):
LGTM

PR #1903, opened 2022-04-30
updated loading data recipe to use latest api of torchaudio.datasets.YESNO

dineshladi:
- `download` parameter is set to `True` so that data is downloaded
- `transform` and `target_transform` parameters are not supported anymore. Hence, removed from the recipe. 
- `root` parameter is undefined in the recipe. Hence `root` parameter value is set to `'./'` so that data is downloaded to the current folder. 
- remove transform parameter related help info in the document

### DIFF
--- a/recipes_source/recipes/loading_data_recipe.py
+++ b/recipes_source/recipes/loading_data_recipe.py
@@ -87,8 +87,6 @@ torchaudio.datasets.YESNO(
 # useful info on the other parameters:
 
 # * ``download``: If true, downloads the dataset from the internet and puts it in root directory. If dataset is already downloaded, it is not downloaded again.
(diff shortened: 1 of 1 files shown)


msaroufim:
Can you run the new command, show me the pytorch version and paste that it works here? Over time I think we should move to make our tutorials testable @svekars

dineshladi:
torch version
```
import torch
import torchaudio

print(f"torch version: {torch.__version__}")
print(f"torchaudio version: {torchaudio.__version__}")

# torch version: 1.11.0
# torchaudio version: 0.11.0
```

```
torchaudio.datasets.YESNO(
     root='./',
     url='http://www.openslr.org/resources/1/waves_yesno.tar.gz',
     folder_in_archive='waves_yesno',
     download=True)

# <torchaudio.datasets.yesno.YESNO at 0x1100d9dc0>
``

malfet:
@msaroufim tutorials that end with `.py` are already testable, so...

PR #1909, opened 2022-05-04
generate .circleci/config.yml from template

### LINKED ISSUE #1898: Generate .circleci/config.yml from template
Right now, PR and master build definitions are completely disconnected. It would be nice to either merge them or generate from template.
Also, `build_manager` job (one that is assembles all tutorials outputs into one place), should be executed at the very end, rather than busy-wait for the workers to finish

cc @malfet @janeyx99

clee2000:
fixes #1898
- generate the config.yml for circleci so that master and pr are in sync
- attempted to keep as much of the file the same, but some new lines got deleted and the ordering changed
- add requires for build manager so it doesnt spin wait


janeyx99:
You're missing the Fixes keyword in front of 1898 just fyi :p

.circleci/regenerate.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def indent(indentation, data_list):
+    return ("\n" + " " * indentation).join(
+        yaml.dump(data_list, default_flow_style=False).splitlines()
+    )
+
+
+def jobs(pr_or_master, num_workers=20, indentation=2):
+    w = {}
```
janeyx99, line 27: What do w and d stand for?

.circleci/regenerate.py
```diff
@@ -0,0 +1,100 @@
...
+    needs_gpu_nvidia_small_multi = [3, 9, 12, 13]
```
janeyx99, line 28: Do we know why 3, 9, 12, and 13 need nvidia small multi? Same as the second line

.circleci/regenerate.py
clee2000: i stole it from vision, where i think it stood for workflow? and then i just kept using it for all of them
d is dict
i will get slightly better names

.circleci/regenerate.py
clee2000: no clue, i took them from the original config.yml, i assume its because these shards end up running tutorials that need those machines but i cant find info about which tutorials specifically need them

when i do sharding w/ some sort of meta data ill probably figure it out

janeyx99 (approved):
Looks fine so far, but I think more logic can potentially be delegated to the template and less to the python script in the future

Issue #1899, opened 2022-04-21
Better sharding mechanism


malfet:
Right now sharding is done in round-robin manner per folder, see https://github.com/pytorch/tutorials/blob/master/.jenkins/build.sh
Which creates a very unbalanced experience as some folders have 20+ tutorials, but many have as few as 3-4 tutorials, which is guaranteed to schedule more tutorials build in first shards and leave later shards with just 1 or 2 tutorials to build

This can be addressed by enumerating all tutorials and dividing them between shards evenly

clee2000:
its already round robin sharded, we just get really unlucky b/c some tutorials take 20 min to build and they end up on the same shard

should do meta data thing like multigpu to keep track of tutorials that take a long time to run and shard based on that

clee2000:
as stated in #1910 
> there are three files (can see them in the metadata file) that are long running (20-30 minutes) and are responsible for the unbalance in the shards. The long run time comes from running a training loop, so we should see what can be done about not running this training loop everytime (perhaps caching the result or reducing the number of epochs)

svekars:
Any update on this by any chance?

PR #1910, opened 2022-05-04
[ci] convert sharding from bash to python

### LINKED ISSUE #1905: label tutorials as multigpu vs not
Sometimes tutorials CI breaks because tutorials get sharded to the wrong machine so we should label tutorials that require multigpu machine vs. not

cc @malfet @janeyx99

clee2000:
fixes #1905
helps w/ #1899
Changes:
- convert sharding from bash to python
- create a metadata file to list long running jobs + jobs that require special machines
- shard based on time
  - default to 1 min if not a super long time
- only 0th worker uses gpu.nvidia.small.multi

There are three files (can see them in the metadata file) that are long running (20-30 minutes) and are responsible for the unbalance in the shards.  The long run time comes from running a training loop, so we should see what can be done about not running this training loop everytime (perhaps caching the result or reducing the number of epochs)

### DIFF
--- a/.circleci/regenerate.py
+++ b/.circleci/regenerate.py
@@ -25,8 +25,10 @@ def indent(indentation, data_list):
 
 def jobs(pr_or_master, num_workers=20, indentation=2):
     jobs = {}
+
+    # all tutorials that need gpu.nvidia.small.multi machines will be routed
+    # by get_files_to_run.py to 0th worker
     needs_gpu_nvidia_small_multi = [0]
-    needs_gpu_nvidia_medium = []
     jobs[f"pytorch_tutorial_{pr_or_master}_build_manager"] = {
         "<<": "*pytorch_tutorial_build_manager_defaults"
     }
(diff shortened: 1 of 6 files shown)


.jenkins/get_files_to_run.py
```diff
@@ -0,0 +1,79 @@
...
+
+def main():
+    num_shards = int(os.environ.get("NUM_WORKERS", 20))
+    shard_num = int(os.environ.get("WORKER_ID", 0))
+
+    all_files = get_all_files()
+    files_to_run = calculate_shards(all_files, num_shards=num_shards)[shard_num]
+    remove_other_files(all_files, files_to_run)
+    stripped_file_names = list(map(lambda x: Path(x).stem, files_to_run))
+    print(" ".join(stripped_file_names))
```
janeyx99, line 75: Could these be separated by new lines for better readability in the logs?

janeyx99 (approved):

.jenkins/get_files_to_run.py
clee2000: I added a newline so if you run the python script on its own there are new lines, but the output gets swallowed up by FILES_TO_RUN and printing that doesn't seem to include the new line unfortunately

.circleci/regenerate.py
```diff
@@ -25,8 +25,8 @@ def indent(indentation, data_list):
 
 def jobs(pr_or_master, num_workers=20, indentation=2):
     jobs = {}
-    needs_gpu_nvidia_small_multi = [3, 9, 12, 13]
-    needs_gpu_nvidia_medium = [14, 15]
+    needs_gpu_nvidia_small_multi = [0]
+    needs_gpu_nvidia_medium = []
```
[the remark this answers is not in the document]
clee2000: removed

PR #1918, opened 2022-05-14
Update introyt1_tutorial.py

DokySp:
Hello. I found the issue on [Introduction to PyTorch](https://pytorch.org/tutorials/beginner/introyt/introyt1_tutorial.html) - [Datasets and Dataloaders](https://pytorch.org/tutorials/beginner/introyt/introyt1_tutorial.html#datasets-and-dataloaders) example. To run this code, I struggled with RuntimeError because of multiprocessing issue. Also I couldn't see the result of pyplot image. So I modify some code like below.
 - Add `plt.show()` in `imshow()`
 - Add `freese_support()`

Reference of RuntimeError
https://stackoverflow.com/questions/50701690/pytorch-tutorial-error-training-a-classifier

### DIFF
--- a/beginner_source/introyt/introyt1_tutorial.py
+++ b/beginner_source/introyt/introyt1_tutorial.py
@@ -357,6 +357,7 @@ trainloader = torch.utils.data.DataLoader(trainset, batch_size=4,
 
 import matplotlib.pyplot as plt
 import numpy as np
+from multiprocessing import freeze_support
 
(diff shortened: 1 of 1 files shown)


malfet:
Thank you for your contribution. Can you please mention in the PR description (and perhaps in tutorial code as well), that `freeze_support` is only needed if/when running tutorial on Windows, according to https://docs.python.org/3.8/library/multiprocessing.html#multiprocessing.freeze_support

PR #1922, opened 2022-05-19
Update buildmodel_tutorial.py

alecbuetow:
I messed up the last pull request's formatting, info is still the same


malfet:
Not sure I understand the comment, content of two patches seems the same, isn't it? Closing

PR #1925, opened 2022-05-22
Update memory_format_tutorial.py

mingfeima:
Add perf gain data for CPU Channels last part.


svekars (approved):

intermediate_source/memory_format_tutorial.py
```diff
@@ -240,6 +241,11 @@
 # ``alexnet``, ``mnasnet0_5``, ``mnasnet0_75``, ``mnasnet1_0``, ``mnasnet1_3``, ``mobilenet_v2``, ``resnet101``, ``resnet152``, ``resnet18``, ``resnet34``, ``resnet50``, ``resnext50_32x4d``, ``shufflenet_v2_x0_5``, ``shufflenet_v2_x1_0``, ``shufflenet_v2_x1_5``, ``shufflenet_v2_x2_0``, ``squeezenet1_0``, ``squeezenet1_1``, ``vgg11``, ``vgg11_bn``, ``vgg13``, ``vgg13_bn``, ``vgg16``, ``vgg16_bn``, ``vgg19``, ``vgg19_bn``, ``wide_resnet101_2``, ``wide_resnet50_2``
 #
 
+######################################################################
+# The following list of models has the full support of Channels last and showing 26%-76% perf gains on Intel(R) Xeon(R) Platinum 8380 CPU (codenamed Ice Lake):
```
malfet, line 245: As perf gain range is so wide, I wonder whether specifying specific CPU model is important? Should we say something like "perf gains on IceLake or newer CPUs?")

mingfeima: Good point! specifying specific CPU model is not very important since the channels last performance gain should be equally effective.

PR #1936, opened 2022-06-06
Add FX Numeric Suite Core APIs tutorial

vkuzo:
Summary:

Adds a simple tutorial on using the FX numeric suite core APIs on MobileNetV2.

Test plan:

```
make html-noplot
// renders well on Mac OS
```

### DIFF
--- a/prototype_source/fx_numeric_suite_tutorial.py
+++ b/prototype_source/fx_numeric_suite_tutorial.py
@@ -1,0 +1,231 @@
+# -*- coding: utf-8 -*-
+"""
+PyTorch FX Numeric Suite Core APIs Tutorial
+==============================
+
+Introduction
+------------
+
+Quantization is good when it works, but it’s difficult to know what's wrong
+when it doesn't satisfy the accuracy we expect. Debugging the accuracy issue
+of quantization is not easy and time consuming.
+
+One important step of debugging is to measure the statistics of the float model
+and its corresponding quantized model to know where are they differ most.
+We built a suite of numeric tools called PyTorch FX Numeric Suite Core APIs in
+PyTorch quantization to enable the measurement of the statistics between
+quantized module and float module to support quantization debugging efforts.
+Even for the quantized model with good accuracy, PyTorch FX Numeric Suite Core
+APIs can still be used as the profiling tool to better understand the
+quantization error within the model and provide the guidance for further
+optimization.
+
+PyTorch FX Numeric Suite Core APIs currently supports models quantized through
+both static quantization and dynamic quantization with unified APIs.
+
+In this tutorial we will use MobileNetV2 as an example to show how to use
+PyTorch FX Numeric Suite Core APIs to measure the statistics between static
+quantized model and float model.
+
+Setup
+^^^^^^
+We’ll start by doing the necessary imports:
+"""
+
+##############################################################################
+
+# Imports and util functions
+
+from __future__ import print_function, division, absolute_import
+import copy
+import torch
+import torchvision
+import torch.quantization
+import torch.ao.ns._numeric_suite_fx as ns
+import torch.quantization.quantize_fx as quantize_fx
+
+import matplotlib.pyplot as plt
+from tabulate import tabulate
+
+torch.manual_seed(0)
+plt.style.use('seaborn-whitegrid')
+
+
+# a simple line graph
+def plot(xdata, ydata, xlabel, ylabel, title):
+    _ = plt.figure(figsize=(10, 5), dpi=100)
+    plt.xlabel(xlabel)
+    plt.ylabel(ylabel)
+    plt.title(title)
+    ax = plt.axes()
+    ax.plot(xdata, ydata)
+    plt.show()
+
+##############################################################################
+# Then we load the pretrained float MobileNetV2 model, and quantize it.
+
+
+# create float model
+mobilenetv2_float = torchvision.models.quantization.mobilenet_v2(
+    pretrained=True, quantize=False).eval()
+
+# create quantized model
+qconfig_dict = {
+    '': torch.quantization.get_default_qconfig('fbgemm'),
+    # adjust the qconfig to make the results more interesting to explore
+    'module_name': [
+        # turn off quantization for the first couple of layers
+        ('features.0', None),
+        ('features.1', None),
+        # use MinMaxObserver for `features.17`, this should lead to worse
+        # weight SQNR
+        ('features.17', torch.quantization.default_qconfig),
+    ]
+}
+# Note: quantization APIs are inplace, so we save a copy of the float model for
+# later comparison to the quantized model. This is done throughout the
+# tutorial.
+mobilenetv2_prepared = quantize_fx.prepare_fx(
+    copy.deepcopy(mobilenetv2_float), qconfig_dict)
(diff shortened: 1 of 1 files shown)


prototype_source/fx_numeric_suite_tutorial.py
```diff
@@ -0,0 +1,231 @@
...
+"""
+PyTorch FX Numeric Suite Core APIs Tutorial
+==============================
+
+Introduction
+------------
+
+Quantization is good when it works, but it’s difficult to know what's wrong
+when it doesn't satisfy the accuracy we expect. Debugging the accuracy issue
+of quantization is not easy and time consuming.
```
svekars, line 11:
```suggestion
of quantization is not easy and time-consuming.
```

prototype_source/fx_numeric_suite_tutorial.py
```diff
@@ -0,0 +1,231 @@
...
+    'module_name': [
+        # turn off quantization for the first couple of layers
+        ('features.0', None),
+        ('features.1', None),
+        # use MinMaxObserver for `features.17`, this should lead to worse
+        # weight SQNR
+        ('features.17', torch.quantization.default_qconfig),
+    ]
+}
+# Note: quantization APIs are inplace, so we save a copy of the float model for
```
svekars, line 85:
```suggestion
# Note: quantization APIs are in place, so we save a copy of the float model for
```

prototype_source/fx_numeric_suite_tutorial.py
```diff
@@ -0,0 +1,231 @@
...
+# later comparison to the quantized model. This is done throughout the
+# tutorial.
+mobilenetv2_prepared = quantize_fx.prepare_fx(
+    copy.deepcopy(mobilenetv2_float), qconfig_dict)
+datum = torch.randn(1, 3, 224, 224)
+mobilenetv2_prepared(datum)
+# Note: there is a long standing issue that we cannot copy.deepcopy a
+# quantized model. Since quantization APIs are inplace and we need to use
```
svekars, line 93:
```suggestion
# quantized model. Since quantization APIs are in place and we need to use
```

prototype_source/fx_numeric_suite_tutorial.py
```diff
@@ -0,0 +1,231 @@
...
+#
+# .. figure:: /_static/img/compare_output.png
+#
+# The add_loggers/extract_logger_info API knows how to extract activations
+# from any layer with a torch.Tensor return type. It works for dynamic
+# quantization as well as PTQ/QAT.
+
+# Compare unshadowed activations
+
+# Create a new copy of the quantized model, because we cannot copy.deepcopy
```
svekars, line 174:
```suggestion
# Create a new copy of the quantized model, because we cannot `copy.deepcopy`
```

prototype_source/fx_numeric_suite_tutorial.py
```diff
@@ -0,0 +1,231 @@
...
+plot(
+    [x[0] for x in mobilenetv2_wt_to_print],
+    [x[4] for x in mobilenetv2_wt_to_print],
+    'idx',
+    'sqnr',
+    'weights, idx to sqnr'
+)
+
+##############################################################################
+# also print out the SQNR, so we can inspect the layer name and type
```
svekars, line 153:
```suggestion
# Also print out the SQNR, so we can inspect the layer name and type:
```

prototype_source/fx_numeric_suite_tutorial.py
```diff
@@ -0,0 +1,231 @@
...
+mobilenetv2_quantized = quantize_fx.convert_fx(
+    copy.deepcopy(mobilenetv2_prepared))
+
+##############################################################################
+# 1. Compare the weights of float and quantized models
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# The first analysis we can do is comparing the weights of fp32 model and int8
+# model by calculating the SQNR between each pair of weights.
+#
+# The extract_weights API knows how to extract weights from linear, convolution
```
svekars, line 106:
```suggestion
# The `extract_weights` API can extract weights from linear, convolution
```

prototype_source/fx_numeric_suite_tutorial.py
```diff
@@ -0,0 +1,231 @@
...
+plot(
+    [x[0] for x in mobilenet_v2_act_to_print],
+    [x[4] for x in mobilenet_v2_act_to_print],
+    'idx',
+    'sqnr',
+    'unshadowed activations, idx to sqnr',
+)
+
+##############################################################################
+# also print out the SQNR, so we can inspect the layer name and type
```
svekars, line 227:
```suggestion
# Also print out the SQNR, so we can inspect the layer name and type:
```

prototype_source/fx_numeric_suite_tutorial.py
malfet: Agreed about decorating extract_weights with backticks, but not sure if can is the right verb here. "can be used to"  perhaps (or simply "API extracts weights")
```suggestion
# The `extract_weights` API can be used to extract weights from linear, convolution
```

malfet (changes requested):
Please remove future imports not needed in Python-3 environment, and use back-ticks to highlight codeblocks in documentation

vkuzo:
thanks folks! I will make all of the suggested changes.

prototype_source/fx_numeric_suite_tutorial.py
vkuzo: keeping it as is, lmk if this needs to change

prototype_source/fx_numeric_suite_tutorial.py
vkuzo: keeping it as is, lmk if this needs to change

prototype_source/fx_numeric_suite_tutorial.py
```diff
@@ -0,0 +1,231 @@
...
+Setup
+^^^^^^
+We’ll start by doing the necessary imports:
+"""
+
+##############################################################################
+
+# Imports and util functions
+
+from __future__ import print_function, division, absolute_import
```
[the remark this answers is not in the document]
vkuzo: this was copy-pasted from a similar tutorial, good catch, thanks

prototype_source/fx_numeric_suite_tutorial.py
```diff
@@ -0,0 +1,231 @@
...
+        ('features.1', None),
+        # use MinMaxObserver for `features.17`, this should lead to worse
+        # weight SQNR
+        ('features.17', torch.quantization.default_qconfig),
+    ]
+}
+# Note: quantization APIs are inplace, so we save a copy of the float model for
+# later comparison to the quantized model. This is done throughout the
+# tutorial.
+mobilenetv2_prepared = quantize_fx.prepare_fx(
```
hx89, line 87: I think Jerry has made a change to prepare_fx() to require example input, may update to incorporate that change?

vkuzo: @hx89 , yes, the example_inputs change will be in v1.13, we will update the tutorial in the RC for v1.13 with those changes.  This one is for v1.12.

hx89 (approved):

PR #1937, opened 2022-06-07
added beginner tutorial for linear regression

jsnowdon99:


svekars:
@jsnowdon99 thanks for contributing to PyTorch tutorials! Can we please move this to the recipes/ as this is more of a bite-size content? Also, one suggestion to make it more educational is to break the single code block into multiple code blocks with descriptions on what's happening in each block?

beginner_source/linear_regression_tutorial.py
```diff
@@ -0,0 +1,75 @@
...
+================
+**Author:** Jack Snowdon
+
+In this tutorial, we will generate a linear regression with PyTorch to find
+the best-fitting line in a linear relationship between an independent and 
+dependent variable. Linear regressions are popular among ML projects used
+for predictive analysis.
+"""
+
+from torch import nn
```
albanD, line 12: This can be cleaned up to
```python
import torch
from torch import nn, tensor
```

beginner_source/linear_regression_tutorial.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # Backward pass, update optimizer parameter given current gradient.
+    loss.backward()
+    optimizer.step()
+
+
+# After training, test with a novel x_value
+x = tensor([[5.0]])
+y_pred = model(x)
+print("Prediction:", y_pred.data[0][0])
```
albanD, line 75: Don't use `.data`, you can use `.detach()`.

svekars:
Closing for now as we have not heard from the original author in a while. Feel free to reopen, if needed.

Issue #1868, opened 2022-03-23
TypeError: ZipperIterDataPipe instance doesn't have valid length


Gare-Ng:
Just copy the code in "LANGUAGE TRANSLATION WITH NN.TRANSFORMER AND TORCHTEXT" and try run it on my own PC, it pop up this error and it seems do not have solutions on the Internet now.

hezhongyu:
I have the same question. It seems you cannot get the length of the variable "train_iter". Tracing it, I found the problem may occur in built-in function(__len__()) of torch.utils.data.datapipes.iter.combining.ZipperIterDataPipe, but I have no idea how to fix it.

Krb686:
@Gare-Ng @hezhongyu I noticed the same issue today.  If you just remove the " / len(train_dataloader)" and " / len(val_dataloader)" in the train_epoch and evaluate functions respectively, the model will run.  I still haven't quite figured out if that's logically correct though.  The model ran and spit out "A group of people stand in front of an igloo"

lixiaojun2914:
The problem could be solved by: 
`return losses / len(list(train_dataloader))`

jiluojiluo:
> 

yes,it works,thanks.

jiluojiluo:
> yes, it works,thanks



> The problem could be solved by: `return losses / len(list(train_dataloader))`

PR #1954, opened 2022-06-22
Fixes issue #1868 regarding TypeError: ZipperIterDataPipe

otioss:
This fixes issue #1868:

> TypeError: ZipperIterDataPipe instance doesn't have valid length


hudeven (approved):
LGTM

beginner_source/translation_transformer.py
```diff
@@ -303,7 +303,7 @@ def train_epoch(model, optimizer):
         optimizer.step()
         losses += loss.item()
 
-    return losses / len(train_dataloader)
+    return losses / len(list(train_dataloader))
```
hudeven, line 306: It might be better to fix the `__len__` of ZipperIterDataPipe in torchdata: https://github.com/pytorch/pytorch/blob/ea72a0991c3422d8f314acdf8b911de42a6b4c1e/torch/utils/data/datapipes/iter/combining.py#L554

cc: @ejguan

ejguan: It's better to append a `LenthSetterDataPipe` at the end of each Dataset within TorchText with a known length. For `Zipper` in torchdata, it's for general usage that requires the prior `DataPipe` provides valid length.

hudeven: cc: @Nayef211

Nayef211: @ejguan if my understanding is correct, calling `list(train_dataloader)` would cause all data in the dataset to materialize right? I thought this is something we'd want to avoid given the fact that this can lead to OOMs for very large datasets.

> It's better to append a LenthSetterDataPipe at the end of each Dataset within TorchText with a known length.

If we do this, can we avoid the data materialization issue I mentioned above?

ejguan: Correct, we shouldn't do materialization here. If this Dataset has a known length, we can add a DataPipe:
```py
class LenSetter(IterDataPipe):
    def __init__(self, length):
        self.length = length

    def __len__(self):
        self.length = lenth
```
This would prevent any pre-materialization before data iterated. `IterDataPipe` itself doesn't guarantee to provide a `__len__` function because there might DataPipe operations introduce the dynamic length (`filter`)

Nayef211: Gotcha, just created an issue to track this https://github.com/pytorch/text/issues/1943. Will merge this PR for now

PR #1955, opened 2022-06-22
Updated URLs for Multi30k dataset in translation_transformer tutorial

Nayef211:
## Description
- We need to programmatically modify the URLs for the multi30k dataset since the links to the original dataset are broken. 
- We plan to followup with a PR in torchtext that directly updates the URLs in source code, however this PR is still needed to make the dataset work for the 0.13 release (since the torchtext PR will not make it into the latest release)
- Added a commented line to install torchdata (a required dependency to work with torchtext datsets)


parmeet (approved):
LGTM! Thanks @Nayef211 for the fix.

PR #1957, opened 2022-06-29
Disable functorch test

malfet:


malfet:
Pushed directly to trunk

PR #1973, opened 2022-07-13
Create Fast Transformer Inference with Better Transformer

mikekgfb:
Add a tutorial demonstrating the use of Better Transformer for fast transformer inference


msaroufim (approved):

msaroufim:
Looks good but make sure to also update the index - here's an example https://github.com/pytorch/tutorials/pull/1969/files

PR #1977, opened 2022-07-14
Fix math render in Autograd Saved Tensors Hooks Tutorial

awgu:
The example shows `y = a * b`. The original equation does not render because of a space between `:math:` and the formula. The original equation also has an extraneous `\mapsto`.
Screen Shot 2022-07-14 at 11 42 13 AM


zhaojuanmao (approved):

awgu:
@svekars This is ready to merge.

PR #1995, opened 2022-08-03
Add gh action to build tutorial age

svekars:
This GH action will run periodically to create a .csv file with the time tutorial files were updated last. This stats in combination with other analytics will help us prioritize our work on tutorials.

### DIFF
--- a/.github/workflows/page-age.yml
+++ b/.github/workflows/page-age.yml
@@ -36,4 +36,4 @@ jobs:
         with: 
           branch: main
           directory: ./stats
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.STATS_TOKEN }}
--- a/.github/workflows/page-age.yml
+++ b/.github/workflows/page-age.yml
@@ -1,0 +1,39 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Page age
+
+# Controls when the workflow will run
+on:
+  schedule:
+    # Run this once per day, towards the end of the day for keeping the most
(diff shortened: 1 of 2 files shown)


malfet (changes requested):
The script is good to have, but what is the point of checking this file in? Isn't it always a function of github repo at the current moment?

I.e. what's wrong with just checking in the script and running it when those stats are needed?

.github/workflows/page-age.yml
```diff
@@ -0,0 +1,39 @@
...
+          fetch-depth: 0
+      - name: Make script executable
+        run: chmod +x .github/workflows/page-age-script.sh
+      - name: Run script
+        run: .github/workflows/page-age-script.sh
+      # Runs a set of commands using the runners shell
+      - name: push
+        uses: actions-x/commit@v6    
+        with: 
+          branch: master
```
malfet, line 37: Can we push it to a different branch (say ghpages)

.github/workflows/page-age.yml
```diff
@@ -0,0 +1,39 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Page age
+
+# Controls when the workflow will run
+on:
+  schedule:
+    # Run this once per day, towards the end of the day for keeping the most
+    # recent data point most meaningful (hours are interpreted in UTC).
+    - cron: "55 23 * * 0"
```
malfet, line 10: Page age only gets non-trivial updates, when there are changes to the repo, should we perhaps change the trigger to:
```
on:
  push:
    - maser
```

PR #2002, opened 2022-08-12
Fix same memory location error

### LINKED ISSUE #1742: ddp_pipeline
I ran the code as is on the cluster. Gives

`RuntimeError: unsupported operation: some elements of the input tensor and the written-to tensor refer to a single memory location. Please clone() the tensor before performing the operation.`

What could be wrong? Also, is there any way to run this code in Jupyter? By the way, the Colab [notebook](https://colab.research.google.com/github/pytorch/tutorials/blob/gh-pages/_downloads/8976a0b7cba4d8c4bc2a28205b91a7da/ddp_pipeline.ipynb) doesn't run and gives and error

`Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/r/roman-koshkin/miniconda3/envs/tranformer/lib/python3.8/multiprocessing/spawn.py", line 116, in spawn_main
    exitcode = _main(fd, parent_sentinel)
  File "/home/r/roman-koshkin/miniconda3/envs/tranformer/lib/python3.8/multiprocessing/spawn.py", line 126, in _main
    self = reduction.pickle.load(from_parent)
AttributeError: Can't get attribute 'run_worker' on <module '__main__' (built-in)>
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/r/roman-koshkin/miniconda3/envs/tranformer/lib/python3.8/multiprocessing/spawn.py", line 116, in spawn_main
    exitcode = _main(fd, parent_sentinel)
  File "/home/r/roman-koshkin/miniconda3/envs/tranformer/lib/python3.8/multiprocessing/spawn.py", line 126, in _main
    self = reduction.pickle.load(from_parent)
AttributeError: Can't get attribute 'run_worker' on <module '__main__' (built-in)>`

mabdullahsoyturk:
Fixes #1742 as suggested by @vpj


mabdullahsoyturk:
@suraj813 doesn't seem to be active @svekars

malfet:
Pinged Pritam about the review

malfet:
Since this snippet has been copy-n-pasted from https://pytorch.org/tutorials/beginner/transformer_tutorial.html we probably need to change it there as well. Also wouldn't the same can be achieved by using `self.register_parameter`? (though I though we need grads on that encoder)

mrshenli (approved):
Stamp to unblock, but it seems to me the longer term solution should be correctly supporting buffers?

mabdullahsoyturk:
> Since this snippet has been copy-n-pasted from https://pytorch.org/tutorials/beginner/transformer_tutorial.html we probably need to change it there as well. Also wouldn't the same can be achieved by using `self.register_parameter`?

Yes, it could be achieved by using `self.register_parameter` but I am not sure how it's any better. 

> Stamp to unblock, but it seems to me the longer term solution should be correctly supporting buffers?

I agree. It seems like the cause of the problem is the [synchronization of the buffers](https://github.com/pytorch/pytorch/blob/master/torch/nn/parallel/distributed.py#L1598).

svekars:
@malfet @mrshenli so what's the final on this? wait for a better solution or ok to merge for now?

malfet (approved):
Fine to merge as is, will do a small followup to change it to register parameter.

PR #2006, opened 2022-08-14
Add new tutorial: Multi-Objective NAS with Ax

Balandat:
Adds a new tutorial for how to perform multi-objective Neural Architecture Search with Ax. The tutorial walks the user through a case study that shows how to use Bayesian optimization in order to find Pareto-efficient tradeoffs between model complexity and performance in a sample-efficient fashion. The tutorial uses PyTorch lightning and TorchX to run a number of actual training jobs in the background, and uses Ax+BoTorch to automate the multi-objective optimization. 

Built website output: [Multi-Objective NAS with Ax — PyTorch Tutorial [BUILD].zip](https://github.com/pytorch/tutorials/files/9334218/Multi-Objective.NAS.with.Ax.PyTorch.Tutorial.BUILD.zip)
 
cc @dme65, @d4l3k


beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
+# -*- coding: utf-8 -*-
+"""
+Multi-Objective NAS with Ax
+==================================================
+
+**Authors:** `David Eriksson <https://github.com/dme65>`__,
+`Max Balandat <https://github.com/Balandat>`__,
+and the Adaptive Experimentation team at Meta.
+
+In this tutorial we show how to use `Ax <https://ax.dev/>`__ to run
```
svekars, line 10:
```suggestion
In this tutorial, we show how to use `Ax <https://ax.dev/>`__ to run
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+multi-objective neural architecture search (NAS) for a simple neural
+network model on the popular MNIST dataset. While the underlying
+methodology would typically be used for more complicated models and
+larger datasets, we opt for a tutorial that is easily runnable
+end-to-end on a laptop in less than 20 minutes.
+
+In many NAS applications, there is a natural tradeoff between multiple
+objectives of interest. For instance, when deploying models on-device
+we may want to maximize model performance (e.g. accuracy), while
```
svekars, line 19:
```suggestion
we may want to maximize model performance (for example, accuracy), while
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+
+######################################################################
+# Defining the TorchX App
+# -----------------------
+#
+# Our goal is to optimize the PyTorch Lightning training job defined
+# in `mnist_train_nas.py <https://github.com/pytorch/tutorials/tree/master/beginner_source/mnist_train_nas.py>`__.
+# To do this using TorchX, we write a helper function that takes in
+# the values of the architcture and hyperparameters of the training
+# job and creates a TorchX AppDef (see the
```
svekars, line 52:
```suggestion
# job and creates a `TorchX AppDef <https://pytorch.org/torchx/latest/basics.html>`__
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# `TorchX docs <https://pytorch.org/torchx/latest/basics.html>`__
```
svekars, line 53:
```suggestion
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# for more detail) with the appropriate settings.
```
svekars, line 54:
```suggestion
# with the appropriate settings.
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+#
+# Ax’s `Runner <https://ax.dev/api/core.html#module-ax.core.runner>`__
+# abstraction allows writing interfaces to various backends.
+# Ax already comes with Runner for TorchX, and so we just need to
+# configure it. For the purpose of this tutorial we run jobs locally
+# in a fully asynchronous fashion.
+#
+# In order to launch them on a cluster, you can instead specify a
+# different TorchX scheduler and adjust the configuration appropriately
+# (e.g. if you have a Kubernetes cluster, you just need to change the
```
svekars, line 126:
```suggestion
# For example, if you have a Kubernetes cluster, you just need to change the
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# ---------------------
+#
+# Ax’s `Runner <https://ax.dev/api/core.html#module-ax.core.runner>`__
+# abstraction allows writing interfaces to various backends.
+# Ax already comes with Runner for TorchX, and so we just need to
+# configure it. For the purpose of this tutorial we run jobs locally
+# in a fully asynchronous fashion.
+#
+# In order to launch them on a cluster, you can instead specify a
+# different TorchX scheduler and adjust the configuration appropriately
```
svekars, line 125:
```suggestion
# different TorchX scheduler and adjust the configuration appropriately.
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# (e.g. if you have a Kubernetes cluster, you just need to change the
+# scheduler from local_cwd to kubernetes).
```
svekars, line 127:
```suggestion
# scheduler `from local_cwd` to `kubernetes`.
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# with the goal of maximizing the validation accuracy and minimizing
+# the number of model parameters. The latter represents a simple proxy
+# of model latency, which is hard to estimate accurately for small ML
+# models (in an actual application we would benchmark the latency while
+# running the model on-device).
+#
+# In our example TorchX will run the training jobs in a fully asynchronous
+# fashion locally and write the results to the ``log_dir`` based on the trial
+# index (see the ``trainer()`` function above). We will define a metric
+# class that is a ware of that logging directory. By subclassing
```
svekars, line 242: aware?

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# `TensorboardCurveMetric <https://ax.dev/tutorials/multiobjective_optimization.html>`__
+# we get the logic to read and parse the tensorboard logs for free.
```
svekars, line 244:
```suggestion
# we get the logic to read and parse the Tensorboard logs for free.
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+    # still running. We don't use this in the current tutorial, but Ax
+    # utilizes this to implement trial-level early-stopping functionality.
+    @classmethod
+    def is_available_while_running(cls):
+        return False
+
+
+######################################################################
+# Now we can instatiate the metrics for accuracy and the number of
+# model parameters Here `curve_name` is the name of the metric in the
```
svekars, line 273:
```suggestion
# model parameters. Here `curve_name` is the name of the metric in the
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+#
+# The way to tell Ax what it should optimize is by means of an
+# `OptimizationConfig <https://ax.dev/api/core.html#module-ax.core.optimization_config>`__.
+# Here we use a ``MultiObjectiveOptimizationConfig`` as we will
+# be performing multi-objective optimization.
+#
+# Additionally, Ax supports placing constraints on the different
+# metrics by specifying objective thresholds, which bound the region
+# of interest in the outcome space that we want to explore. For this
+# ecample we will constrain the validation accuracy to be at least
```
svekars, line 303:
```suggestion
# example, we will constrain the validation accuracy to be at least
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+
+######################################################################
+# Creating the Ax Experiment
+# --------------------------
+#
+# In Ax, the `Experiment <https://ax.dev/api/core.html#module-ax.core.experiment>`__
+# object is the object that stores all the information about the problem
+# setup.
+#
+# ..tip:
```
svekars, line 333:
```suggestion
# .. tip:
```
It won't generate properly otherwise.

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+#
+# A `GenerationStrategy <https://ax.dev/api/modelbridge.html#module-ax.modelbridge.generation_strategy>`__
+# is the abstract representation of how we would like to perform the
+# optimization. While this can be customized (if you’d like to do so, see
+# `this tutorial <https://ax.dev/tutorials/generation_strategy.html>`__),
+# in most cases Ax can automatically determine an appropriate strategy
+# based on the search space, optimization config, and the total number
+# of trials we want to run.
+#
+# Typically, Ax will choose to evauate a number of random configurations
```
svekars, line 361:
```suggestion
# Typically, Ax chooses to evaluate a number of random configurations
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+  )
+
+
+######################################################################
+# Configuring the Scheduler
+# -------------------------
+#
+# The `Scheduler` (TODO: link) acts as the loop control for the optimization.
+# It communicates with the backend to launch trials, check their status,
+# and retrieve results (in the case of this tutorial it is simply reading
```
svekars, line 383:
```suggestion
# and retrieve results. In the case of this tutorial, it is simply reading
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# and parsing the locally saved logs, bug in a remote execution setting
```
svekars, line 384:
```suggestion
# and parsing the locally saved logs. In a remote execution setting,
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# it would call APIs). The following illustration from the Ax
```
svekars, line 385:
```suggestion
# it would call APIs. The following illustration from the Ax
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# summarizes how the Scheduler interacts with external systems used to run
+# trial evaluations:
+#
+# .. image:: ../../_static/img/ax_scheduler_illustration.png
+#
+#
+# The ``Scheduler`` requires the ``Experiment`` and the ``GenerationStrategy``.
+# A set of options can be passed in via ``SchedulerOptions``. Here, we
+# configure the number of total evaluations as well as ``max_pending_trials``,
+# the maximum number of trials that should run concurrently (in our
```
svekars, line 396:
```suggestion
# the maximum number of trials that should run concurrently. In our
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# local setting this is the number of training jobs running as individual
```
svekars, line 397:
```suggestion
# local setting, this is the number of training jobs running as individual
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# processes, while in a remote execution setting this would be the number
```
svekars, line 398:
```suggestion
# processes, while in a remote execution setting, this would be the number
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# of machines you want to use in parallel).
```
svekars, line 399:
```suggestion
# of machines you want to use in parallel.
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+)
+
+
+######################################################################
+# Running the optimization
+# ------------------------
+#
+# Now that everything is configured, we can let Ax run the optimization
+# in a fully automated fashion. The Scheduler will periodially check
+# the logs for the status of all currenty running trials, and if a
```
svekars, line 420:
```suggestion
# the logs for the status of all currently running trials, and if a
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+######################################################################
+# Evaluating the results
+# ----------------------
+#
+# We can now inspect the result of the optimization using helper
+# functions and visualizations included with Ax.
+
+######################################################################
+# First, we generate a dataframe with a summary of the results
+# of the experiment. Each row in this dataframe corresponds ot a
```
svekars, line 438:
```suggestion
# of the experiment. Each row in this dataframe corresponds to a
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# trial (i.e. a training job that was run), and contains information
```
svekars, line 439:
```suggestion
# trial (such as a training job that was run), and contains information
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+df = exp_to_df(experiment)
+df.head(10)
+
+
+######################################################################
+# We can also visualize the Pareto frontier of tradeoffs between the
+# validation accuracy and the number of model parameters.
+#
+# .. tip::
+#   Ax uses Ploly to produce interactive plots, which allow you to
```
svekars, line 456:
```suggestion
#   Ax uses Plotly to produce interactive plots, which allow you to
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+#   do things like zoom, crop, or hover in order to view details
+#   of components of the plot. Try it out, and take a look at the
+#   `visualization tutorial <https://ax.dev/tutorials/visualizations.html>`__
+#   if you'd like to learn more).
+#
+# The final optimization results are shown in the Figure below where
```
svekars, line 462:
```suggestion
# The final optimization results are shown in the figure below where
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# the color corresponds to the iteration number for each trial.
+# We see that our method was able to successfully explore the
+# trade-offs and found both large models with high validation
+# accuracy as well as small models with compartively lower
```
svekars, line 466:
```suggestion
# accuracy as well as small models with comparatively lower
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+#
+
+from ax.service.utils.report_utils import _pareto_frontier_scatter_2d_plotly
+
+_pareto_frontier_scatter_2d_plotly(experiment)
+
+
+######################################################################
+# To better understand what our surrogate models have learned about
+# the black box objectives, we can take a look at the Leave-one-out
```
svekars, line 477:
```suggestion
# the black box objectives, we can take a look at the leave-one-out
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# cross validation results. Since our models are Gaussian Processes,
+# they not only provide point predictions but also uncertainty estimates
+# about these predictions. A good model means
```
svekars, line 480:
```suggestion
# about these predictions. A good model means
```
Should "A good model means" be deleted?

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+#
+# As the figures below show, the model size (num_params) metric is
```
svekars, line 482:
```suggestion
# As the figures below show, the model size (`num_params`) metric is
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+# much easier to model than the validation accuracy (val_acc) metric.
```
svekars, line 483:
```suggestion
# much easier to model than the validation accuracy (`val_acc`) metric.
```

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+
+cv = cross_validate(model=gs.model)  # The surrogate model is stored on the GenerationStrategy
+compute_diagnostics(cv)
+
+interact_cross_validation_plotly(cv)
+
+
+######################################################################
+# We can also make contour plots to better understand how the different
+# objectives depend on two of the input parameters. In the figure below
```
svekars, line 498:
```suggestion
# objectives depend on two of the input parameters. In the figure below,
```

svekars (commented):
Thanks, @Balandat! Looks great! I added a few editorial suggestions and there seems to be a problem with data loading. From the [log](https://app.circleci.com/pipelines/github/pytorch/tutorials/6323/workflows/d16f58d8-073a-4130-a37b-977fb839a214/jobs/124222?invite=true#step-104-4547): 

```
Aug 14 22:51:14 ax.exceptions.core.NoDataError: Observed data is required for generation step #1 (model MOO), but fetched data was empty. Something is wrong with experiment setup -- likely metrics do not implement fetching logic (check your metrics) or no data was attached to experiment for completed trials.
```

Balandat:
> Thanks, @Balandat! Looks great! I added a few editorial suggestions and there seems to be a problem with data loading. From the [log](https://app.circleci.com/pipelines/github/pytorch/tutorials/6323/workflows/d16f58d8-073a-4130-a37b-977fb839a214/jobs/124222?invite=true#step-104-4547):
> 
> ```
> Aug 14 22:51:14 ax.exceptions.core.NoDataError: Observed data is required for generation step #1 (model MOO), but fetched data was empty. Something is wrong with experiment setup -- likely metrics do not implement fetching logic (check your metrics) or no data was attached to experiment for completed trials.
> ```

Hmm this is interesting. I've never ran into this issue before, this seems to either a problem with running the training jobs or at least with accessing their logs. A few different folks are able to run this locally, maybe it's an issue with the environment / permissions of the circleci runner? Is there a way to run this as a docker image or ssh into the job to understand what's going on?

index.rst
```diff
@@ -492,14 +492,21 @@ What's new in PyTorch tutorials?
...
+
 .. customcarditem::
    :header: Grokking PyTorch Intel CPU Performance from First Principles
    :card_description: A case study on the TorchServe inference framework optimized with Intel® Extension for PyTorch.
    :image: _static/img/thumbnails/cropped/generic-pytorch-logo.png
    :link: intermediate/torchserve_with_ipex
    :tags: Model-Optimization,Production
 
+.. customcarditem::
+   :header: Multi-Objective Neural Architechture Search with Ax
```
d4l3k, line 504: nit: Architecture*

index.rst
```diff
@@ -492,14 +492,21 @@ What's new in PyTorch tutorials?
...
+   :card_description: Learn how to use Ax to search over architectures find optimal tradeoffs between accuracy and latency.
+   :image: _static/img/ray-tune.png
+   :link: beginner/ax_multiobjective_nas_tutorial.html
+   :tags: Model-Optimization,Best-Practice
```
d4l3k, line 508: might be nice to add an Ax and TorchX tag here, seems like that's what TorchRec does

beginner_source/ax_multiobjective_nas_tutorial.py
Balandat: I think you need double ` for this to render as code
```suggestion
# scheduler from ``local_cwd`` to ``kubernetes``).
```

beginner_source/ax_multiobjective_nas_tutorial.py
Balandat: yes
```suggestion
# class that is aware of that logging directory. By subclassing
```

beginner_source/ax_multiobjective_nas_tutorial.py
Balandat:
```suggestion
# trial (that is, a training job that was run), and contains information
```

beginner_source/ax_multiobjective_nas_tutorial.py
Balandat: This was incomplete, thanks for catching this.

beginner_source/ax_multiobjective_nas_tutorial.py
```diff
@@ -0,0 +1,524 @@
...
+    return specs.AppDef(
+        name="trainer",
+        roles=[
+            specs.Role(
+                name="trainer",
+                image=TORCHX_IMAGE,
+                entrypoint="python",
+                args=[
+                    "-c",
+                    train_source,
```
d4l3k, line 93: any reason you're passing the train_source instead of just passing the script path?

might also be good to use the python builtin component to reduce amount of TorchX knowledge required https://pytorch.org/torchx/main/components/utils.html#torchx.components.utils.python

beginner_source/ax_multiobjective_nas_tutorial.py
Balandat:
```suggestion
# about these predictions. A good model means that the predicted means
# (the points in the figure) are close to the 45 degree line and that the
# confidence intervals cover the 45 degree line with the expected frequency
# (here we use 95% confidence intervals, so we would expect them to contain
# the true observation 95% of the time).
```

beginner_source/ax_multiobjective_nas_tutorial.py
Balandat:
```suggestion
# As the figures below show, the model size (``num_params``) metric is
```

beginner_source/ax_multiobjective_nas_tutorial.py
Balandat:
```suggestion
# much easier to model than the validation accuracy (``val_acc``) metric.
```

beginner_source/ax_multiobjective_nas_tutorial.py
Balandat: > any reason you're passing the train_source instead of just passing the script path?

Hmm good point - this is likely a remnant from the original colab setup where the script code was defined in the notebook. Let me see if I can simplify this. 

> might also be good to use the python builtin component to reduce amount of TorchX knowledge required

good idea

index.rst
Balandat: Architechture is a cool word though in the context of NAS
```suggestion
   :header: Multi-Objective Neural Architecture Search with Ax
```

index.rst
Balandat:
```suggestion
   :tags: Model-Optimization,Best-Practice,Ax,TorchX
```

svekars:
@malfet any thoughts on this: https://app.circleci.com/pipelines/github/pytorch/tutorials/6327/workflows/dc48542e-9102-4927-870d-1c202c77e079/jobs/124310?invite=true#step-104-4588

beginner_source/ax_multiobjective_nas_tutorial.py
Balandat: > might also be good to use the python builtin component to reduce amount of TorchX knowledge required

Actually, I'm not sure this is easily possible. The current `trainer()` function does some dynamic parsing of the trial index to generate the log path that is then passed to the training script. AFAICT it seem that the python builtin component static and wouldn't allow for this parsing. The other option would of course be to pass only the base log path to the script and add an additional `trial_idx` cmd line arg, but ideally doing that wouldn't be necessary. Is there any way to pass dynamic transformations as args to the builtin python component?

Balandat: Nvm I don't think this question makes much sense now that I think about it, I can just redefine `trainer` in the same way.

Balandat:
@danielrjiang just pointed out to me that the trials are completing in one second, which is not good - it means that the training actually fails. I'll dig into why that may be the case (likely some issue with the environment or the relative location of the script file...).

@d4l3k is there a straightforward way to view the logs kept by TorchX in a programmatic way (without using the command line tool `torch log`)? Then I can catch the exception raised here and print the logs to debug what's going on.

beginner_source/mnist_train_nas.py
```diff
@@ -0,0 +1,170 @@
...
+logger.log_metrics({"val_acc": val_accuracy})
+
+# Log the number of model parameters
+num_params = trainer.model.num_params
+logger.log_metrics({"num_params": num_params})
+
+logger.save()
+
+# Print outputs
+eprint(f"train time: {train_time}, val acc: {val_accuracy}, num_params: {num_params}")
```
malfet, line 170: Just curious, why this needs to go to `stderr`?

beginner_source/mnist_train_nas.py
```diff
@@ -0,0 +1,170 @@
...
+)
+parser.add_argument(
+    "--hidden_size_2", type=int, required=True, help="hidden size layer 2"
+)
+parser.add_argument("--learning_rate", type=float, required=True, help="learning rate")
+parser.add_argument("--epochs", type=int, required=True, help="number of epochs")
+parser.add_argument("--dropout", type=float, required=True, help="dropout probability")
+parser.add_argument("--batch_size", type=int, required=True, help="batch size")
+
+args = parser.parse_args()
```
malfet, line 46: Nit
```suggestion
def parse_args():
    parser = argparse.ArgumentParser(description="train mnist")
    parser.add_argument(
        "--log_path", type=str, required=True, help="dir to place tensorboard logs from all trials"
    )
    parser.add_argument(
        "--hidden_size_1", type=int, required=True, help="hidden size layer 1"
    )
    parser.add_argument(
        "--hidden_size_2", type=int, required=True, help="hidden size layer 2"
    )
    parser.add_argument("--learning_rate", type=float, required=True, help="learning rate")
   parser.add_argument("--epochs", type=int, required=True, help="number of epochs")
    parser.add_argument("--dropout", type=float, required=True, help="dropout probability")
    parser.add_argument("--batch_size", type=int, required=True, help="batch size")
    return parser.parse_args()

args = parse_args()
```

beginner_source/mnist_train_nas.py
malfet, line 170: Can this be refactored into main routine and guarded with `if __name__ == "main":`?
```suggestion
logger = pl_loggers.TensorBoardLogger(args.log_path)

eprint(args.log_path)

# Train the model and log time ⚡
start = time.time()
trainer.fit(model=mnist_model)
end = time.time()
train_time = end - start
logger.log_metrics({"train_time": end - start})

# Compute the validation accuracy once and log the score
with io.capture_output() as captured:
    val_accuracy = trainer.validate()[0]["val_acc"]
logger.log_metrics({"val_acc": val_accuracy})

# Log the number of model parameters
num_params = trainer.model.num_params
logger.log_metrics({"num_params": num_params})

logger.save()

# Print outputs
eprint(f"train time: {train_time}, val acc: {val_accuracy}, num_params: {num_params}")
```

Balandat: I don't think it needs to, this was probably copied over from where this was initially set up. I can change it to just use stdout.

malfet:
Our CI system (and .py to -> .ipynb converter) is currently implemented around the idea, that single file corresponds to single tutorial. As this tutorial comprises of two files, it:
- will result in generating 2 "tutorials"
- It accidentally break CI system as different parts of tutorials are rendered in different shard, and all tutorials not targeting are temporarily rendered empty

Both problems can be easily fixed by blocklisting stripping code/rendering of auxilary file, but I'm curious how Jupyther notebook will look like in this case, and whether it is a good experience from user point of view, when they can not see part of the tutorial code on a single page.

Balandat:
Thanks for catching the issue!

I'm linking the training script in the tutorial file, so it should be pretty straightforward for the user to view it. But I do take your point about that potentially being confusing / not as clear as it could be if it's not shown on the same page, so I'd be happy to consider changing this if it makes sense.

Note that we can just pass the source code of the training script as a string to the TorchX runner. So one option could be to define the code directly in the notebook. The issue with that is that if it's a string it won't have syntax highlighting etc. and so it would be hard to read. Maybe there is some fancy way to extract the source code of a specific cell into a string (though since this is a .py file I guess the concept of a cell doesn't even exist yet prior to the parsing)?

I'm not super familiar with what one could do here using rst functionality, maybe there is a way to render the separate py file in some minipage as part of the tutorial?

malfet (approved):
LGTM, though we might want to move it to different folder later

Issue #2010, opened 2022-08-15
Add a caution comment for installing OpenCV in Raspberry Pi


YunjinJo:
https://github.com/pytorch/tutorials/blob/86aca7dc14f3189320c0d6d59ec8cf539049093c/intermediate_source/realtime_rpi.rst#L70-L72  
If installing OpenCV in Raspberry Pi using the command ```pip install opencv-contrib-python```, it takes about 2 hours. And nothing shows in the terminal until installation finishes. People who follow this tutorial could misconceive their Raspberry Pi stop working.  
I think it would be better if a caution comment in the code block.

svekars:
Resolved with  #2011

PR #2011, opened 2022-08-15
Add a caution comment for installing OpenCV in Raspberry Pi

YunjinJo:
Issue #2010 
Add a caution comment in the code block to prevent a user from misconceiving.


intermediate_source/realtime_rpi.rst
```diff
@@ -68,6 +68,7 @@ PyTorch and all the other libraries we need have ARM 64-bit/aarch64 variants so
 .. code:: shell
 
     $ pip install torch torchvision torchaudio
+    # Installing OpenCV will take about 2 hours and nothing shows in the terminal until it finishes.
```
svekars, line 71: @d4l3k is this expected?

d4l3k: That seems way longer than expected. @YunjinJo  what RPi are you using? Are you using 64bit arm or 32bit? And it's the install that takes a long time not the download?

d4l3k (approved):

intermediate_source/realtime_rpi.rst
d4l3k: Though, if it builds from the source distribution that might be reasonable... 

Seems like other folks are seeing the same thing: https://raspberrypi-guide.github.io/programming/install-opencv

There's some prebuild rpi wheels at https://www.piwheels.org/project/opencv-contrib-python/ but only for 32bit

d4l3k:
Here's the issues for adding prebuilt aarch64 wheels to opencv python 

https://github.com/opencv/opencv-python/issues/473
https://github.com/opencv/opencv-python/issues/562

d4l3k:
Looks like opencv-python has a aarch64 prebuilt library -- wonder if we should switch to that instead of contrib https://pypi.org/project/opencv-python/#files

d4l3k:
Since we're only using `cv2.VideoCapture` and that's available in opencv-python does seem like we should just switch to that (or at least recommend it)

@YunjinJo have you tried just using `opencv-python`?

YunjinJo:
> Since we're only using `cv2.VideoCapture` and that's available in opencv-python does seem like we should just switch to that (or at least recommend it)
> 
> @YunjinJo have you tried just using `opencv-python`?

@d4l3k  
I tested on my RPI (RPI 4 model B 8GB, RPI OS 64-bit) using `opencv-python` and it works well. I think it would be better to use `opencv-python` than `opencv-contrib-python`.

svekars (approved):

PR #2023, opened 2022-08-30
Add missing backslash in math formula

paulhaider:
Fix math formula


malfet (approved):
Thank you for the fix

PR #2024, opened 2022-08-31
Fix error in comment in tensors_tutorial.py

Sh-Ac:
Line 152, z3 is said to have the same value as z1, z2, which is not true.


malfet (changes requested):
What do you mean it's not true? z1, z2 and z3 would have the same values

Sh-Ac:
Oops, my bad. Everything was fine after all. Apologies for the distraction.

Issue #263, opened 2018-06-22
DQN is broken


zspasztori:
Hi,

I am running your implementation of DQN.
Tried several runs, but it does not seem to converge with the given hyperparameters.
Tried higher amount of episodes in the magnitude of 1000, but there is no convergence.

Also the ploting has problems, you need to move the display part before sleep.
![dqn_tutorial](https://user-images.githubusercontent.com/12113563/41774606-b7066c82-7620-11e8-938e-87b726209518.png)

cc @vmoens @nairbv

johan-gras:
![figure_2](https://user-images.githubusercontent.com/30185046/43036611-58247f14-8d05-11e8-850a-126350f514b1.png)

Same issue here.
Running the exact code from the tutorial for 10 000 episodes and the algorithm doesn't converge.
I've been testing lots of differents hyperparameters without success.

cebenso2:
I am experiencing the same issue.

leonardblier:
I also have the same issue.

danaugrs:
Same:

![pytorch-dqn-example](https://user-images.githubusercontent.com/5261448/51807454-1e344600-226e-11e9-8f3f-e066ec3b255d.png)

vmikulik:
+1 to the plotting code not working (on MacOS, using Jupyter). Swapping display and clear_output solved the issue for me.

mansur007:
Has anyone solved this problem?
I am having it too...

zhihaocheng:
same too

mansur007:
> Has anyone solved this problem?
> I am having it too...

If you anyone is still looking for a working DQN - check out mine: https://github.com/mansur007/gym_cartpole_dqn

josiahls:
After making my own DQN. My main concern:
- Why is the target network being updated every few **episodes**. What DQN paper is that supposed to be related to? Surely you should be doing per step, like every 5-10 steps update the target network.

As noted:
![image](https://user-images.githubusercontent.com/19930483/65838095-76c05500-e2cd-11e9-8327-21fcf49437d5.png)
[1] Mnih, Volodymyr, et al. "Playing atari with deep reinforcement learning."
            arXiv preprint arXiv:1312.5602 (2013).

Key quote "parameters from the previous iteration θi−1 are held fixed when optimising the loss function" they do not say previous episode, they say previous iteration.

**I recommend changing your target updates to every few iterations.**

[...]

PR #2026, opened 2022-09-05
DQN tutorial fixes

SiftingSands:
Gym's domain name has changed since https://github.com/pytorch/tutorials/pull/1928#issue-1247029039

Minor updates to Gym calls, which would produce errors with the latest version of Gym (0.25.2). (possibly fixes https://github.com/pytorch/tutorials/issues/1432#issue-837212016)

Still doesn't train properly as noted in https://github.com/pytorch/tutorials/issues/1755#issue-1071634106 , but at least it runs now as shown in the figure below (didn't touch any other settings in the tutorial).

Figure_2

PR #2030, opened 2022-09-07
[WIP] Improve training of DQN tutorial

SiftingSands:
Following up the discussion from https://github.com/pytorch/tutorials/pull/2026

I still need to do multiple runs to get a semblance of the statistics of # episodes vs duration for both the original and my changes. The slight increase in model capacity still only uses ~1.5 GB of VRAM, so it should be pretty accessible and training is still relatively quick.

Here's the reward history for one run of these tweaks when I was doing a bunch of trial and error (spent an embarrassing amount of time tweaking hyperparameters and rewards).

duration_only_reward_03

@vmoens Feel free to change (or completely discard) anything based on your findings. I haven't tried tweaking anything else in the training pipeline.


vmoens (commented):
Looks like we're on the good track!

One thing I'm not super familiar / comfortable with is the batch norm. In general many don't use it in RL. In this case I would at least turn it off for the select_action function, ie. for data collection (call policy.eval() before and policy.train() after).
For the target network I would call requires_grad_(False) rather than detaching and I think we can avoid calling eval on it (the target net could be in train no?), hence batch_norm would still be "active" there.

SiftingSands:
See if that last commit is what you had in mind. I decided to use torch.no_grad() instead of setting requires_grad=False all of the params of the target network. I agree, it seems logical to have the batch norm behavior be the same for both the target and policy networks for loss calculation. However, I wonder if the target network only updating every 10 episodes makes that a moot point.  I got it running in the background anyway.

FYI I'm only doing 10 runs for each "config", so I'm sure the mean results (when I get around to making those graphs) will still be noisy. We could try removing batch norm altogether as well.

The only other thought I had right now was changing the replay memory sampling from uniform to something learned or a heuristic, but I'm on the fence due to the added complexity for a tutorial.

vmoens:
> The only other thought I had right now was changing the replay memory sampling from uniform to something learned or a heuristic, but I'm on the fence due to the added complexity for a tutorial

I don't think that a prioritized RB will change much and that will be a lot of trouble for a tutorial.
But we could update the target net more progressively with little effort (using a moving average for instance, i.e. soft update). A decay of 0.002 is typical but we can go a bit higher than that depending on how easy the problem is to solve.

> We could try removing batch norm altogether as well
I would be in favour of that! 

IMO we should also check if the data as is is properly normalized, I don't think it is. We could compute the stats over, say, 100 images for each channels (i.e. a mean of 3 values and a std of 3 values) and normalize the data at every step with that to get something close to normally distributed.

SiftingSands:
I decided to setup some experiment tracking infrastructure (test driving mlflow), so things are more manageable. I ran 5 configs so far, but only doing 10 runs had very noisy results as expected.

For example, here's a plot of mean +/- 1 sigma and median +/- 25%. `cfg_01` is from the same settings as the first 4 commits here and the same settings that produced the graph in the original post. `baseline` is the original tutorial parameters.
mean_stdev
quartiles

Instead of that mess, I decided to smooth the medians to make multiple runs readable. Here I'm using https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.ewm.html with `alpha = 0.01` (heavy smoothing)
Here's an example comparing median vs smoothed
smooth_example

So, here's the comparison between all of my runs so far. Gonna test out soft updates and input normalization afterwards.
* `cfg_01_vmoensBN` is the most recent commit with changes suggested at https://github.com/pytorch/tutorials/pull/2030#pullrequestreview-1099758509
* `cfg_01_BNdelete` has all batch norm ops removed
* I ran `baseline_BNdelete` after seeing how poor performance was after removing batch norm, and I wanted a sanity check.
initial_results

vmoens:
Feels like i'm giving pretty bad suggestions haha
It would be nice to know how significant those differences are. From what I see from the plot above it could be that the differences are not significant, am I right to assume that?
It could be nice to have a sense of what is the loss value, the value of the gradients etc.
On my side I'm working on a torchrl tutorial based on the same example, so perhaps I'll have some insight to share about all this soon.

pseudo-rnd-thoughts:
@vmoens @SiftingSands Hey, Im one of the maintainers of gym and just saw this PR as I noted we needed to update it with the last changes in v26 so thanks for the PR

Could it be updated to v26 instead of v25? We should specify what gym version this tutorial uses.
If so, then the `gym.make` on line 77 will need to be updated to `env = gym.make('CartPole-v0', render_mode='rgb_array').unwrapped`

I also had a couple of questions about the current tutorial that you might know the answer to
1. Why is the environment unwrapped from `gym.make`? This seems to be to remove the time limit wrapper
2. Is the cartpole environment being imagined as an infinite horizon problem with no time limit (this might be the solution to question 1). If so, we should mention this assumption as it is not true for a majority of problems otherwise, we should add the termination variable to maths equations.

Couple of issues
1. "Install using `pip install gym`" on line 49 should be `pip install gym[classic_control]` to install pygame automatically which is necessary for the rendering
2. "It makes rewards from the uncertain far future less important for our agent than the ones in the near future that it can be fairly confident about." should note that this also encourages agents to collect reward closer in time than equivalent rewards temporally future away
4. `\delta = Q(s, a) - (r + \gamma \max_a Q(s', a))` the `max_a` should probably be `max_a'` to avoid confusion with `Q(s, a)`.
5. "Below, `num_episodes` is set small. You should download the notebook and run lot more epsiodes, such as 300+ for meaningful duration improvements." on 449 is incorrect currently as `num_episodes=1000`

Suggested upgrades
1. Add docstring explaining the `BATCH_SIZE`, `GAMMA` and `TARGET_UPDATE` variables
2. Use `env.action_space.sample()` instead of `random.randrange(n_actions)` in line 350
3. I believe you are using `param.grad.data.clamp_(-1, 1)` on line 437 is gradient clipping. Maybe add a note of this as it is common in RL.
4. Change `done` to `terminated` on line 467 as this is the true meaning of the variable
5. Line 470 does reward shaping, however, to my knowledge, there isn't an explanation of this anywhere

SiftingSands:
> Could it be updated to v26 instead of v25? We should specify what gym version this tutorial uses. If so, then the `gym.make` on line 77 will need to be updated to `env = gym.make('CartPole-v0', render_mode='rgb_array').unwrapped`
> 
> I also had a couple of questions about the current tutorial that you might know the answer to
> 
>     1. Why is the environment unwrapped from `gym.make`? This seems to be to remove the time limit wrapper
> 
>     2. Is the cartpole environment being imagined as an infinite horizon problem with no time limit (this might be the solution to question 1). If so, we should mention this assumption as it is not true for a majority of problems otherwise, we should add the termination variable to maths equations.

`unwrapped` was in Adam's original tutorial, so I didn't bother changing it. I did notice the time limit being unbounded, when there were durations over 200 even though cartpole-v0 was specified. 

> 
> Couple of issues
> 
>     1. "Install using `pip install gym`" on line 49 should be `pip install gym[classic_control]` to install pygame automatically which is necessary for the rendering
> 
>     2. "It makes rewards from the uncertain far future less important for our agent than the ones in the near future that it can be fairly confident about." should note that this also encourages agents to collect reward closer in time than equivalent rewards temporally future away
> 
>     3. `\delta = Q(s, a) - (r + \gamma \max_a Q(s', a))` the `max_a` should probably be `max_a'` to avoid confusion with `Q(s, a)`.
> 
>     4. "Below, `num_episodes` is set small. You should download the notebook and run lot more epsiodes, such as 300+ for meaningful duration improvements." on 449 is incorrect currently as `num_episodes=1000`
> 
> 
> Suggested upgrades
> 
>     1. Add docstring explaining the `BATCH_SIZE`, `GAMMA` and `TARGET_UPDATE` variables
> 
>     2. Use `env.action_space.sample()` instead of `random.randrange(n_actions)` in line 350
> 
>     3. I believe you are using `param.grad.data.clamp_(-1, 1)` on line 437 is gradient clipping. Maybe add a note of this as it is common in RL.
> 
>     4. Change `done` to `terminated` on line 467 as this is the true meaning of the variable
> 
>     5. Line 470 does reward shaping, however, to my knowledge, there isn't an explanation of this anywhere

Thanks for the suggestions on the documentation! It all seems reasonable to me, but I'll defer to @vmoens on further discussion of the doc changes.
The increase in episodes and reward shaping are very much experimental, as this PR is really just brainstorming (hence the WIP).

vmoens:
Hi @pseudo-rnd-thoughts 
Thanks for those suggestions. Indeed it's important that our RL tutorials use gym as it should.
We'll make the changes you suggested.

I don't think we should unwrap the environment, that does not seem like something we want to do.

The number of episodes being run is indeed puzzling in this tutorial and I guess most user wonder why so few are proposed. We should definitely increase that.

For the gradient clippping I believe we should use `torch.nn.utils.clip_grad_value_()` instead, since it's a PyTorch tutorial let's use pytorch tools!

+1 on the done to terminated change.

In a nutshell there is work to do! 

@SiftingSands feel free to make those edits in your PR, I can give a hand if needed.

SiftingSands:
I decided to try out offline normalization like what's done for ImageNet trained models (e.g. Normalize((0.485, 0.456, 0.406), (0.229, 0.224, 0.225))). We could do a moving average and std dev calculation as you suggested, but maybe this is 99% of the benefit. 

Regardless, I tracked the average of the mean and std dev of the "state" over 2000 episodes in the figure below (x axis is actually "steps" instead of episodes). I think the sharp movements are due to the beginning of a new episode. I was hoping the std dev would level out, but I expect it to stay within the range of [0.006, 0.007]. The mean is, unsurprisingly, effectively zero, because the "state" is a frame difference.

state_stat_history

At the end of 2000, episodes these are the numerical results for each channel. The transform that I ended up using was `normalize(state, mean=[0,0,0], std=[0.00656, 0.0106, 0.0152])`
```
Channel 0 mean: 1.7123858279475807e-06, std: 0.006567582970755784
Channel 1 mean: 1.1795560506103564e-06, std: 0.010622139538439322
Channel 2 mean: 8.115399832946285e-07, std: 0.015237523709654841
```

I ran the input normalization on the baseline and my changes from the initial post, `cfg_01`. Looks like a slight improvement, but maybe that will evaporate if ran for more episodes and repeats? Same sentiment as `baseline` vs `cfg_01` to answer you from
 https://github.com/pytorch/tutorials/pull/2030#issuecomment-1243649994
>From what I see from the plot above it could be that the differences are not significant, am I right to assume that?

(switched y axis to linear scaling and increased the font size a bit, but it should show up fine if you click on the figure to expand)
offlineNorm_results

@vmoens Find any improvements on your end? I can still look into soft updates, since that should be a pretty simple change. I've been saving the best model from each run, so I still gotta run them in the environment and see if they actually perform.

pseudo-rnd-thoughts:
Hey, is there any timeline on this PR and what are the aims of the PR as that isn't clear by the initial comment. 

If we re-add the TimeLimit wrapper with CartPole-v1 (defaults to 500), then an easy aim could be for a script that will train a neural network to achieve on average X rewards? Im not sure what a reasonable reward is
This seems like a very helpful tutorial for users and I worry that making a tutorial with every fancy tool RL researchers twhich  will encourage tools that are not helpful for alot of settings to become standard.

vmoens:
> Hey, is there any timeline on this PR and what are the aims of the PR as that isn't clear by the initial comment.
> 
> If we re-add the TimeLimit wrapper with CartPole-v1 (defaults to 500), then an easy aim could be for a script that will train a neural network to achieve on average X rewards? Im not sure what a reasonable reward is This seems like a very helpful tutorial for users and I worry that making a tutorial with every fancy tool RL researchers twhich will encourage tools that are not helpful for alot of settings to become standard.

I agree that the tuto is useful as is but it looks bad if it does not train... I would advise against anything fancy (like reward shaping) but for instance proper normalization of the observations or good usage of gym's API seem to be crucial points.

To narrow the scope, our goal is basically to have a version of the tuto that trains out of the box (which was far from being the case before).

SiftingSands:
@vmoens Here's the results for soft updates of the target network using \tau = 0.002. Updates were done at every step instead of a predefined # of episodes. The baseline run now includes offline input normalization. I reran the baseline, because I updated Gym to v0.26.1 and updated the API calls, so the results are a bit different from the plot from yesterday.

softupdates

I didn't bother running it with my changes, since the reward shaping was quite a hack. 3/10 repeated runs w/ soft updates had episodes with a duration > 400, which is the best I've seen yet. However, the duration vs episode history takes an unfortunate noise dive. I might try tweaking a few parameters in the next day or so, but I won't be able to contribute too much more on this in the near term.

vmoens:
@SiftingSands I got it running using TorchRL, I'll make a PR on https://github.comfacebookresearch/rl by EOD and ping you with it

SiftingSands:
@vmoens Thanks for all the hard work on that tutorial! My understanding is that the current implementation *roughly* on par with your td(0) results? The overall trend of the training history of episode lengths is very reminiscent of the current results. 

As you mentioned, the addition of td(\lambda) is the most significant performance improvement. However, the current tutorial's replay buffer only holds state transitions instead of trajectories. Correct me if I'm wrong, but a td(\lambda) implementation would require this tutorial to have a replay buffer of trajectories instead of state transitions? If so, I'm afraid that I can't commit to making the necessary updates at this time. However, I can open another PR for the minor gym API and doc changes, and someone else can take the reigns from here.

pseudo-rnd-thoughts:
Somehow did not realise this before but why are we doing this as a vision-based version of cartpole? 
If this is meant to be an introduction to reinforcement learning with Gym and neural network based DQN, this seems like a strange first tutorial.

Could we not do an initial tutorial using the raw state observations with Neural networks based DQN then make an advanced tutorial for Atari or Car Racing, problems that are only vision based

SiftingSands:
> I think that here we're bound by the constraints of what the tutorial is built on initially. For instance, we're using CartPole (where the reward is always 1, i.e. in some sense it is sparse, as only the done/terminated signal is informative), which isn't the most easy task to solve in a few iterations (not saying it's hard but it requires more trials, especially with td(0)). I'm also surprised by the choice to do this with pixels. I get that this is more appealing but it's much harder than learning from vector states, and again for a tutorial that needs to be solved in a few thousands of episodes I'd be surprised if we could do any better than what we're doing here.

Agreed, this level of performance is probably all we can hope with vanilla DQN on pixels (trying to stay in the "spirit" of the original tutorial). I was initially expecting better results within a 1000 episodes (tutorial starts off w/ 50 and only suggests 300+), because this was within the official PyTorch docs.

@pseudo-rnd-thoughts Vision based cartpole is a "legacy" choice from the original author of the tutorial. I would defer to @vmoens on the decision to pivot to a state vector input.

@vmoens I can do the 3rd option you outlined and drop a mention to your torchrl tutorial. Just let me know if I should do that, and I'll open up a new PR. Unfortunately, that's all I can commit to at this time. (Although, switching to a non-pixel input should be very straightforward and likely achieve good performance within a 1000 episodes?)

vmoens:
Let me see with the authors of the tutorial just to make sure we're not rushing into anything but in general I'd be supportive to switch to a state-based tutorial. 
Having a tuto that does not learn is counterproductive IMO.

SiftingSands:
> Let me see with the authors of the tutorial just to make sure we're not rushing into anything but in general I'd be supportive to switch to a state-based tutorial. Having a tuto that does not learn is counterproductive IMO.
> 
> EDIT: upon reflection I think that we should move to a state-based tuto and make the PR. Do you want to work on that @SiftingSands or should I take care of it? You've already done a lot so no pressure, I'd be happy to take it over.

Sounds like a plan. I'll push all of my changes to the PR this weekend. Switched to the state vector input w/ a 3 layer MLP, used AdamW instead of RMSProp, and minor changes to gamma, epsilon decay, tau (kept soft updates) -> got duration plateauing to 500 steps within 150 episodes with some annoying dips that quickly recover.

(didn't bother making a fancy chart for this one, episode duration vs episode #)
newplot(1)

pseudo-rnd-thoughts:
@SiftingSands thanks for the changes. 
Given that we have changed the problem from an infinite horizon problem to a finite horizon (time limited) problem.
I think that we should change some of the maths to reflect termination and truncation. 
We also need a note on the gym version used

SiftingSands:
All of my changes have been committed. It should have all of the suggested changes from @pseudo-rnd-thoughts except for elaborating on the changes to a finite time horizon. @vmoens I'll have to pass that part onto you (I assume you can make edits unless you're not designated as a 'maintainer'). 

Here's a few final plots just for fun.

'baseline' hyperparameters:
BATCH_SIZE = 128
GAMMA = 0.999
EPS_START = 0.9
EPS_END = 0.05
EPS_DECAY = 200
TAU = 0.002
RMSProp w/ default settings

'final' hyperparameters:
BATCH_SIZE = 128
GAMMA = 0.99
EPS_START = 0.9
EPS_END = 0.05
EPS_DECAY = 1000
TAU = 0.005
LR = 0.0001 for AdamW w/ amsgrad=True

Decided to remove smoothing of the median duration / episode over 10 repeated runs in this plot. Colored 'ranges' denote min and max duration. Still have some runs with those 'dips', but oh well.
final_comparison

Proof that the plotting in the tutorial still works, and I guess that I got an unlucky seed during this run.
big

intermediate_source/reinforcement_q_learning.py
```diff
@@ -30,51 +30,44 @@
...
 import torch
 import torch.nn as nn
 import torch.optim as optim
 import torch.nn.functional as F
-import torchvision.transforms as T
 
-
-env = gym.make('CartPole-v0', new_step_api=True, render_mode='single_rgb_array').unwrapped
+# Wrap the environment to limit the maximum number of steps per episode
+env = TimeLimit(gym.make('CartPole-v1'), max_episode_steps=500)
```
pseudo-rnd-thoughts, line 70: This is unnecessary to add the `TimeLimit`, it is already done in the background within `gym.make`
You should be able to do `gym.make("CartPole-v1")`

intermediate_source/reinforcement_q_learning.py
```diff
@@ -433,44 +362,38 @@ def optimize_model():
...
+    # Initialize the environment and get it's state
+    state, _ = env.reset()
+    state = torch.tensor(state, dtype=torch.float32, device=device).unsqueeze(0)
     for t in count():
-        # Select and perform an action
         action = select_action(state)
-        _, reward, done, _, _ = env.step(action.item())
+        observation, reward, terminated, truncated, _ = env.step(action.item())
         reward = torch.tensor([reward], device=device)
+        if truncated:
```
pseudo-rnd-thoughts, line 390: This is not true, truncation != termination
`done = termination or truncation` in terms of if to reset the environment

This blog post will hopefully be made more public soon https://github.com/pseudo-rnd-thoughts/gym/issues/1

pseudo-rnd-thoughts (commented):
I don't believe that the truncation implementation is done correctly.

Is there a way of viewing the final page?

intermediate_source/reinforcement_q_learning.py
SiftingSands: Fixed

intermediate_source/reinforcement_q_learning.py
SiftingSands: Should the replay buffer still be updated on the `truncation` signal with `next_state = torch.tensor(observation, dtype=torch.float32, device=device).unsqueeze(0)`? Performance on this toy problem is similar in either case, but I assume one is more technically correct?

```
done = terminated or truncated
if terminated:
    next_state = None
else:
    next_state = torch.tensor(observation, dtype=torch.float32, device=device).unsqueeze(0)
# Store the transition in memory
memory.push(state, action, next_state, reward)
```

vs

```
done = terminated or truncated
if done:
    next_state = None
else:
    next_state = torch.tensor(observation, dtype=torch.float32, device=device).unsqueeze(0)
# Store the transition in memory
memory.push(state, action, next_state, reward)
```

pseudo-rnd-thoughts: The first one is correct. Thanks for doing that

vmoens (approved):
LGTM, thanks @SiftingSands for taking care of this!

svekars:
The preview page gives me 404 for some reason: https://deploy-preview-2030--pytorch-tutorials-preview.netlify.app/intermediate/reinforcement_q_learning.html

SiftingSands:
> The preview page gives me 404 for some reason: https://deploy-preview-2030--pytorch-tutorials-preview.netlify.app/intermediate/reinforcement_q_learning.html

I wonder if there's a formatting issue with one of my changes to the docstrings that is causing this failure. Do we have any logs to find the culprit? My best guess is maybe line 318 
```
# `soft update <https://arxiv.org/pdf/1509.02971.pdf>`__ controlled by
```

intermediate_source/reinforcement_q_learning.py
```diff
@@ -386,11 +314,9 @@ def plot_durations():
...
 # loss. By definition we set :math:`V(s) = 0` if :math:`s` is a terminal
 # state. We also use a target network to compute :math:`V(s_{t+1})` for
-# added stability. The target network has its weights kept frozen most of
-# the time, but is updated with the policy network's weights every so often.
-# This is usually a set number of steps but we shall use episodes for
-# simplicity.
-#
+# added stability. The target network is updated at every step with a 
+# `soft update <https://arxiv.org/pdf/1509.02971.pdf>`__ controlled by 
+# the hyperparameter ``TAU``, which was previously defined.
```
vmoens, line 319: @SiftingSands Previously there was always a new line ending the comment blocks -- I wonder if that's the reason why the html rendering is broken. Would you mind adding the empty comment line (`#`) back?

SiftingSands: Haha it looks like we typed at the same time and found the same culprit. I'll give that a shot

vmoens: I still can't see it... :/

intermediate_source/reinforcement_q_learning.py
```diff
@@ -480,17 +403,21 @@ def optimize_model():
 
         # Perform one step of the optimization (on the policy network)
         optimize_model()
+
+        # Soft update of the target network's weights
+        # θ′ ← τ θ + (1 −τ )θ′
```
vmoens, line 408: could it be that special characters are not supported? Did we have these before?

SiftingSands: That was not present before. I'll just remove that line.

vmoens:
We still have this issue... I will investigate, not sure what's happening

SiftingSands:
> We still have this issue... I will investigate, not sure what's happening

I might try some brute force debugging at the expense of a messy commit history. This is really the only thing I see in the deploy logs. It appears for other tutorials, so I'm not sure this is the issue.
```
3:49:08 PM: /opt/build/repo/index.rst:780: WARNING: toctree contains reference to nonexisting document 'intermediate/text_to_speech_with_torchaudio'
3:49:08 PM: /opt/build/repo/index.rst:797: WARNING: toctree contains reference to nonexisting document 'beginner/translation_transformer'
3:49:08 PM: /opt/build/repo/index.rst:812: WARNING: toctree contains reference to nonexisting document 'intermediate/reinforcement_q_learning'
3:49:08 PM: /opt/build/repo/index.rst:821: WARNING: toctree contains reference to nonexisting document 'intermediate/flask_rest_api_tutorial'
3:49:08 PM: /opt/build/repo/index.rst:821: WARNING: toctree contains reference to nonexisting document 'advanced/super_resolution_with_onnxruntime'
```

SiftingSands:
@vmoens Alright, those last 3 commits were sanity checks on the rest of the docstring changes. As expected, none of them resolved the issue. 

Here's one thing that I noticed. Currently, this tutorial is not under the "Reinforcement Learning" tab in the left sidebar.
image

However, the build from #2026 shows the sidebar properly filled.
image

vmoens:
I noticed that too but it was already the case with your previous PR, and the rendering was ok...

SiftingSands:
@vmoens Hmm, in that case I'm not familiar enough with the build process to know where to look next. For now, it doesn't look like the docstring changes are the issue. I have a feeling it's something minor that we're missing.
@svekars Could you provide any assistance, or know if anyone else familiar with the webpages could help?

vmoens:
> @vmoens Hmm, in that case I'm not familiar enough with the build process to know where to look next. For now, it doesn't look like the docstring changes are the issue. I have a feeling it's something minor that we're missing. @svekars Could you provide any assistance, or know if anyone else familiar with the webpages could help?

When looking at [this](https://app.circleci.com/pipelines/github/pytorch/tutorials/6582/workflows/181b6049-0def-4aeb-9046-f6d76ff60e62/jobs/129067) it seems that multiple pages are not found, all unrelated to ours
```
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:725: WARNING: toctree contains reference to nonexisting document 'beginner/basics/intro'
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:741: WARNING: toctree contains reference to nonexisting document 'beginner/introyt/trainingyt'
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:741: WARNING: toctree contains reference to nonexisting document 'beginner/introyt/captumyt'
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:780: WARNING: toctree contains reference to nonexisting document 'intermediate/text_to_speech_with_torchaudio'
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:797: WARNING: toctree contains reference to nonexisting document 'beginner/translation_transformer'
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:812: WARNING: toctree contains reference to nonexisting document 'intermediate/reinforcement_q_learning'
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:821: WARNING: toctree contains reference to nonexisting document 'intermediate/flask_rest_api_tutorial'
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:821: WARNING: toctree contains reference to nonexisting document 'advanced/super_resolution_with_onnxruntime'
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:833: WARNING: toctree contains reference to nonexisting document 'intermediate/fx_conv_bn_fuser'
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:842: WARNING: toctree contains reference to nonexisting document 'intermediate/forward_ad_usage'
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:868: WARNING: toctree contains reference to nonexisting document 'beginner/profiler'
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:868: WARNING: toctree contains reference to nonexisting document 'beginner/hyperparameter_tuning_tutorial'
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:868: WARNING: toctree contains reference to nonexisting document 'intermediate/parametrizations'
Sep 29 00:20:54 /var/lib/jenkins/workspace/index.rst:888: WARNING: toctree contains reference to nonexisting document 'advanced/ddp_pipeline'
```

IMO the problem is either (1) in the caches used by circleCI or (2) in a previous commit on `main` that broken stuff upstream. I don't think it is related to this PR.

SiftingSands:
> IMO the problem is either (1) in the caches used by circleCI or (2) in a previous commit on `main` that broken stuff upstream. I don't think it is related to this PR.
> 
> EDIT @svekars other PRs have the same problem too, if you try to see the intro tutorial you'll see that it's broken on multiple PRs. We should figure this out!

That's the same feeling that I have as well. Just let me know if I should wait or don't wait to undo those last three commits until the root cause is found. A silver lining is that I found a typo and a few other minor docstring changes to make while putting everything under the microscope.

vmoens:
> That's the same feeling that I have as well. Just let me know if I should wait or don't wait to undo those last three commits until the root cause is found. A silver lining is that I found a typo and a few other minor docstring changes to make while putting everything under the microscope.

The first commit (adding the missing `#`) was ok, the others are probably superfluous. IMO you can discard them

SiftingSands:
> > That's the same feeling that I have as well. Just let me know if I should wait or don't wait to undo those last three commits until the root cause is found. A silver lining is that I found a typo and a few other minor docstring changes to make while putting everything under the microscope.
> 
> The first commit (adding the missing `#`) was ok, the others are probably superfluous. IMO you can discard them

@vmoens The old commits have been removed (glad that my force push didn't make a giant mess). I also changed the diagram in the tutorial to reflect the soft updates. I also included the drawio source file, so it can easily edited in the future. I tried to recreate the original, but it's not a perfect replica.

intermediate_source/reinforcement_q_learning.py
```diff
@@ -30,30 +30,24 @@
...
-the current screen patch and the previous one. This will allow the agent
-to take the velocity of the pole into account from one image.
 
 **Packages**
 
 
 First, let's import needed packages. Firstly, we need
 `gym <https://github.com/openai/gym>`__ for the environment
-(Install using `pip install gym`).
+(Install using `pip install gym`). Developed on v0.26.1 of gym.
```
svekars, line 45: Can we please remove: `Developed on v0.26.1 of gym.` We are updating version of gym to 0.25 which should be sufficient to run this tutorial.

SiftingSands: I would have thought that this only works w/ v0.26 and above because of the truncation vs termination logic. Did this actually run on v0.25? This is mentioned in the `Breaking backward compatibility` section of https://github.com/openai/gym/releases/tag/0.26.0

svekars: @malfet any thoughts?

malfet: Why do we want to make this tutorial exclusive to latest gym version? I.e. I was able to run the tutorial in its current form using both gym-0.25 and gym-0.26, what is wrong with keeping it backward compatible with older version (which by the way is currently preinstalled in Google Colab)

Is there a way to test if ipyb generated from this tutorial is runnable in Colab without any weaks (which imo is an important feature of tutorials, isn't it?)

SiftingSands: I originally added this note due to a comment from @pseudo-rnd-thoughts. I don't have a strong opinion, but I understood that v0.26 had major and breaking changes, so we may want to steer users away from those older gym versions. @pseudo-rnd-thoughts might be able to elaborate further as one of the maintainers of gym.

pseudo-rnd-thoughts: Hi, Gym v26 implemented a number of breaking API changes that while v25 has implemented, by default are turned off. Therefore, to run this tutorial, I believe you are going to have to modify several parameters in `gym.make`, `env.reset`, etc. 
Whatever version of the tutorial we need to make it clear what Gym version we are developing for due to the breaking changes noted above.

malfet: Anybody knows of the plans of https://github.com/Kautenja/nes-py to be compatible with gym v26? Also, any hint when Google Colab can change its default version from v25 to v26?

intermediate_source/reinforcement_q_learning.py
```diff
@@ -438,44 +368,38 @@ def optimize_model():
...
 #
-# Below, `num_episodes` is set small. You should download
-# the notebook and run lot more epsiodes, such as 300+ for meaningful
-# duration improvements.
+# Below, `num_episodes` to 1000, but you should see the model constantly
+# achieve 500 steps within 600 training episodes.
 #
 
-num_episodes = 50
+num_episodes = 1000
```
malfet, line 388: Did we measured how much it would take to finish compared to the original one?

vmoens: I guess that now that things are not pixel-based we can (slightly) recuce that, what do you think @pseudo-rnd-thoughts  @SiftingSands ?
But I don't think 50 episodes will be enough

SiftingSands: I vaguely remember 1000 episodes taking less than 10 minutes with the latest changes here (on a 2080ti). I want to say 50 episodes on the original one took 1-2 minutes, but that lead to obviously insufficient performance.

intermediate_source/reinforcement_q_learning.py
malfet: Sure, but v0.26 still supports compatible mode, so one can tweak the tutorial pretty easily, i.e. something like that should work:
```
env = gym_super_mario_bros.make("SuperMarioBros-1-1-v0", render_mode='rgb', apply_api_compatibility=True)
```

malfet: https://github.com/pytorch/tutorials/pull/2075

pseudo-rnd-thoughts: An initial look at the tutorial that solution would only work for v0.25 and v0.26 so it would be simpler to just have a single version supported. In addition, at a quick look I think the new step API implementation is done incorrectly. I will see if I can update it soon

pseudo-rnd-thoughts:
@SiftingSands @vmoens @malfet What is stopping this PR from merging?

vmoens:
> @SiftingSands @vmoens @malfet What is stopping this PR from merging? 

I'm personally happy with it

malfet:
@pseudo-rnd-thoughts  imo following things:
- It conflicts with trunk (likely due to this change, which by the way also makes it compatible with latest gym: https://github.com/pytorch/tutorials/pull/2073/files)
- It break CI test sanity
- It would not work out of box on Google Collab 
- Before merging it would be good to see how much increase of `num_eposides` from 50 to 1000 would affect tutorial training running time on Collab (I've asked this question 9 days ago, but haven't seen reply yet)

malfet:
> I'm personally happy with it

@vmoens then go ahead and merge it as is :)

pseudo-rnd-thoughts:
@malfet How big an issue is the CI failing? 
Im happy to make the update such that the code works for v25 and v26
I run the script on colab, which installs gym v25 by default as you said but when I update to v26, then the scripts works
I found that it takes about 21 minutes on colab, not fast but it took 500 episodes to reach max reward. So we could reduce the number of episodes to 600 / 700 episodes probably. Secondly, after training the figure disappears which is annoying and we should probably fix that at some point.

SiftingSands:
> @pseudo-rnd-thoughts imo following things:
> 
>     * It conflicts with trunk (likely due to this change, which by the way also makes it compatible with latest gym: https://github.com/pytorch/tutorials/pull/2073/files)
> 
>     * It break CI test sanity
> 
>     * It would not work out of box on Google Collab
> 
>     * Before merging it would be good to see how much increase of `num_eposides` from 50 to 1000 would affect tutorial training running time on Collab (I've asked this question 9 days ago, but haven't seen reply yet)

FYI, I did reply to your training run time question last week : https://github.com/pytorch/tutorials/pull/2030#discussion_r990395666

malfet:
> How big an issue is the CI failing?

@pseudo-rnd-thoughts not sure I understand the question. Tutorials are using sphinx/sphinx-gallery to convert `.py` to `.rst`/`.ipynb` and then finally render them as html. If this process fails then nothing gets published

> Im happy to make the update such that the code works for v25 and v26

As I've mentioned before, this has already been done by https://github.com/pytorch/tutorials/pull/2073

> I found that it takes about 21 minutes on colab,

Don't you think it's a bit too long? Perhaps one can change `num_episodes` to be dependent on the backend - if `CUDA/XPU` then it could be `500`, but if `cpu` it should remain at 50 (just kicked the notebook to benchmark it)

pseudo-rnd-thoughts:
Looking at reasons the CI is failing is due to gym <= v25 is expected and this tutorial requires v26

SiftingSands (commented):
@pseudo-rnd-thoughts Didn't we make a point to remove the unwrapped call on the environment? Not sure we need to specify the render mode, since this is no longer image based.

pseudo-rnd-thoughts:
Yes, my proposed code would be
```python
if gym.__version__[:4] == '0.26':
    env = gym.make('CartPole-v1')
elif gym.__version__[:4] == '0.25':
    env = gym.make('CartPole-v1', new_step_api=True)
else:
    raise ImportException(f"Requires gym v25 or v26, actual version: {gym.__version__}")
```

SiftingSands:
> Yes, my proposed code would be
> 
> ```python
> if gym.__version__[:4] == '0.26':
>     env = gym.make('CartPole-v1')
> elif gym.__version__[:4] == '0.25':
>     env = gym.make('CartPole-v1', new_step_api=True)
> else:
>     raise ImportException(f"Requires gym v25 or v26, actual version: {gym.__version__}")
> ```

Great! Yeah, it looks like that was the one change that got overwritten due to https://github.com/pytorch/tutorials/pull/2073/files . I can put that into a new commit, if that's the most convenient. (I'm not sure if you can directly make changes to this PR, since I believe only "maintainers" can do that) Just let me know what you think is best.

I think limiting the # of episodes to 600 is an acceptable change. I guess we'll wait for @malfet 's feedback, since he was going to benchmark it on colab.

vmoens:
What happens when gym moves to 0.27? Do we expect this not to run? 
Can't we use the `version.parse` you compare versions?

IMO this kind of version-dependent code in a tutorial is counter productive... might be confusing for our first-time users no?

I agree with @malfet, we could have 2 number of episodes depending on the presence of cuda.

pseudo-rnd-thoughts:
We (the current maintenance team) are not planning on making a Gym v27 release, there will be an announcement explaining why hopefully released on Monday

pseudo-rnd-thoughts:
> Yes, my proposed code would be
> 
> ```python
> if gym.__version__[:4] == '0.26':
>     env = gym.make('CartPole-v1')
> elif gym.__version__[:4] == '0.25':
>     env = gym.make('CartPole-v1', new_step_api=True)
> else:
>     raise ImportException(f"Requires gym v25 or v26, actual version: {gym.__version__}")
> ```

I realised it is more ugly than this, on `reset`, if we need to check the gym version as v25 will default to returning just the observation while v26 will return the observation and info. I don't know the elegant way of solving this for both other than adding the if statement above (we can simplify to know that it must be 0.26 or 0.25 now).

SiftingSands:
> > Yes, my proposed code would be
> > ```python
> > if gym.__version__[:4] == '0.26':
> >     env = gym.make('CartPole-v1')
> > elif gym.__version__[:4] == '0.25':
> >     env = gym.make('CartPole-v1', new_step_api=True)
> > else:
> >     raise ImportException(f"Requires gym v25 or v26, actual version: {gym.__version__}")
> > ```
> 
> I realised it is more ugly than this, on `reset`, if we need to check the gym version as v25 will default to returning just the observation while v26 will return the observation and info. I don't know the elegant way of solving this for both other than adding the if statement above (we can simplify to know that it must be 0.26 or 0.25 now).

@pseudo-rnd-thoughts I looked for the announcement for Gym's versioning roadmap that you mentioned last week, but I didn't see anything. No one else commented on limiting the tutorial to v.25 or v.25, so let me know if that code snippet is still what you think is the best way forward. If that's still the case, then I'll make those changes and add the logic for changing the number of episodes dependent on CUDA existence. Hopefully, we can then finally close this PR.

pseudo-rnd-thoughts:
Currently PyTorch tutorials uses Gym v25 as its version.
Therefore. I agree it would be easier to use a single version both for maintenance and ease of reading, we would need to update the other RL tutorials at the same time.

pseudo-rnd-thoughts:
@malfet @SiftingSands I don't think that it is possible for support v26 and v25. The CI is currently failing on `obs, _ = env.reset()` which to support for v25 needs to use `obs, _ = env.reset(return_info=True)`. 

As #2069 is already updating to v26, I would merge #2069 first and then merge this PR

Therefore, I wouldn't bother with supporting both and just support v26. 
It means that this PR will fail CI until the #2069 is merged 

Do you both agree or think there is a better plan?

SiftingSands:
> @malfet @SiftingSands I don't think that it is possible for support v26 and v25. The CI is currently failing on `obs, _ = env.reset()` which to support for v25 needs to use `obs, _ = env.reset(return_info=True)`.
> 
> As #2069 is already updating to v26, I would merge #2069 first and then merge this PR
> 
> Therefore, I wouldn't bother with supporting both and just support v26. It means that this PR will fail CI until the #2069 is merged
> 
> Do you both agree or think there is a better plan?

@pseudo-rnd-thoughts Sorry for the delayed response. I'm fine with waiting until #2069 is merged and only supporting v.26. However, I'm not a maintainer of the pytorch tutorials repo, so @malfet would probably have the final say.

I'm guessing there's a reason why this won't work, besides being ugly?
```
if gym.__version__[:4] == '0.26':
    obs, _ = env.reset() 
elif gym.__version__[:4] == '0.25':
    obs, _ = env.reset(return_info=True)
```

pseudo-rnd-thoughts:
There shouldn't be an issue with that code, I just dislike it for being ugly and that #2069 is already planning on updating to v26.
The problem is that on line 264 you don't include that if else statement for resetting with info.

SiftingSands:
> There shouldn't be an issue with that code, I just dislike it for being ugly and that #2069 is already planning on updating to v26. The problem is that on line 264 you don't include that if else statement for resetting with info.

I made the changes for v.25, because nobody has replied to your last comment #2069 yet. It looks like CI now passes (got tired of not seeing green). I'll leave this up to @malfet or @vmoens on if they want to merge this now or wait.

vmoens (approved):
LGTM! Thanks for the effort @SiftingSands and @pseudo-rnd-thoughts

SiftingSands:
@svekars I believe this will resolve https://github.com/pytorch/tutorials/issues/263

arr00:
Why isn't this being merged? I followed the current tutorial with no luck but this modified version works perfectly.

vmoens:
> What's the hold up on merging this? @vmoens has already approved this.

Can you solve the conflicts? I'll make sure that we merge it after that

SiftingSands:
> > What's the hold up on merging this? @vmoens has already approved this.
> 
> Can you solve the conflicts? I'll make sure that we merge it after that

Conflict is resolved, but CICD is failing on `intermediate_source/ax_multiobjective_nas_tutorial.py`?

vmoens:
Closed by #2145

SiftingSands:
> Closed by #2145

@vmoens Thanks for taking a look at those CI issues and getting this merged!

PR #2031, opened 2022-09-07
[ci] script to validate tutorials were built in > 0 sec

clee2000:
If the running time is 0 minutes 0.000 seconds this means that python code probably didn't run either due to error or to lacking python code, so we check that the output doesn't have this string.

I don't want to break everyone's builds so we have a list of known bad tutorials that currently fail and only fail if there is a new one that is not on this list. 

Example of failure: https://app.circleci.com/pipelines/github/pytorch/tutorials/6423/workflows/9f59712e-0d3b-428c-81cc-13d5eb22f259/jobs/125997

### DIFF
--- a/.jenkins/build.sh
+++ b/.jenkins/build.sh
@@ -105,6 +105,7 @@ if [[ "${JOB_BASE_NAME}" == *worker_* ]]; then
 
   # Step 5: Remove INVISIBLE_CODE_BLOCK from .html/.rst.txt/.ipynb/.py files
   bash $DIR/remove_invisible_code_block_batch.sh docs
+  python .jenkins/validate_tutorial_built_correct.py docs
 
   # Step 6: Copy generated files to S3, tag with commit ID
   7z a worker_${WORKER_ID}.7z docs
--- a/.jenkins/build.sh
+++ b/.jenkins/build.sh
@@ -105,7 +105,7 @@ if [[ "${JOB_BASE_NAME}" == *worker_* ]]; then
 
   # Step 5: Remove INVISIBLE_CODE_BLOCK from .html/.rst.txt/.ipynb/.py files
   bash $DIR/remove_invisible_code_block_batch.sh docs
-  python .jenkins/validate_tutorial_built_correct.py docs
+  python .jenkins/validate_tutorials_built.py docs
 
   # Step 6: Copy generated files to S3, tag with commit ID
   7z a worker_${WORKER_ID}.7z docs
--- a/.jenkins/build.sh
+++ b/.jenkins/build.sh
@@ -139,6 +139,7 @@ elif [[ "${JOB_BASE_NAME}" == *manager ]]; then
 
   # Step 5: Remove INVISIBLE_CODE_BLOCK from .html/.rst.txt/.ipynb/.py files
   bash $DIR/remove_invisible_code_block_batch.sh docs
+  python .jenkins/validate_tutorials_built.py docs
 
   # Step 6: Copy generated HTML files and static files to S3
   7z a manager.7z docs
--- a/.jenkins/build.sh
+++ b/.jenkins/build.sh
@@ -105,7 +105,7 @@ if [[ "${JOB_BASE_NAME}" == *worker_* ]]; then
 
   # Step 5: Remove INVISIBLE_CODE_BLOCK from .html/.rst.txt/.ipynb/.py files
   bash $DIR/remove_invisible_code_block_batch.sh docs
-  python .jenkins/validate_tutorials_built.py docs
+  python .jenkins/validate_tutorials_built.py
 
   # Step 6: Copy generated files to S3, tag with commit ID
   7z a worker_${WORKER_ID}.7z docs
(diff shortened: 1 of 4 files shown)


.jenkins/validate_tutorials_built.py
```diff
@@ -0,0 +1,102 @@
...
+            if (
+                "Total running time of the script: ( 0 minutes  0.000 seconds)"
+                in elem.text
+                and not any(html_file_path.endswith(file) for file in KNOWN_BAD + OK_TO_NOT_RUN)
+            ):
+                did_not_run.append(html_file_path)
+
+    if len(did_not_run) != 0:
+        raise RuntimeError(
+            f"File(s) {' '.join(did_not_run)} are not known bad but ran in 0.000 sec, meaning that any python code in this tutorial probably didn't run"
```
ZainRizvi, line 97: Small tweak for readability

```suggestion
            f"The following file(s) are not known bad but ran in 0.000 sec, meaning that any python code in this tutorial probably didn't run:\n{'\n'.join(did_not_run)}"
```

.jenkins/build.sh
```diff
@@ -105,6 +105,7 @@ if [[ "${JOB_BASE_NAME}" == *worker_* ]]; then
 
   # Step 5: Remove INVISIBLE_CODE_BLOCK from .html/.rst.txt/.ipynb/.py files
   bash $DIR/remove_invisible_code_block_batch.sh docs
+  python .jenkins/validate_tutorials_built.py docs
```
ZainRizvi, line 108: Should this also be added to line 142 below?

.jenkins/validate_tutorials_built.py
```diff
@@ -0,0 +1,102 @@
...
+        for elem in elems:
+            if (
+                "Total running time of the script: ( 0 minutes  0.000 seconds)"
+                in elem.text
+                and not any(html_file_path.endswith(file) for file in KNOWN_BAD + OK_TO_NOT_RUN)
+            ):
+                did_not_run.append(html_file_path)
+
+    if len(did_not_run) != 0:
+        raise RuntimeError(
```
ZainRizvi, line 96: Sanity check: Would this raise result in the build failing?

ZainRizvi (approved):
Looking good!

.jenkins/validate_tutorials_built.py
clee2000: yes, i put a link in the pr body to an example of how it fails

.jenkins/validate_tutorials_built.py
```diff
@@ -0,0 +1,107 @@
...
+                )
+            ):
+                did_not_run.append(html_file_path)
+
+    if len(did_not_run) != 0:
+        raise RuntimeError(
+            "The following file(s) are not known bad but ran in 0.000 sec, meaning that any "
+            + "python code in this tutorial probably didn't run:\n{}".format(
+                "\n".join(did_not_run)
+            )
```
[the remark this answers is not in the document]
clee2000: the black formatter was angry about the backslash inside the brackets for the f-string `f-string expression part cannot include a backslash`

.jenkins/validate_tutorials_built.py
```diff
@@ -0,0 +1,107 @@
...
+    "recipes/recipes/Captum_Recipe.html",
+]
+
+
+def main():
+    build_dir = sys.argv[1]
+
+    html_file_paths = []
+
+    for difficulty in ["beginner", "intermediate", "advanced", "prototype", "recipes"]:
```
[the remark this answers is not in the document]
ZainRizvi: Very cool command, though it took me a while to grok what it did, where that `-7` came from, etc. So please comment liberally if you take this route 😅.  

+1 to the at least renaming the `difficulty` var to `tutorial_source_dir`

.jenkins/validate_tutorials_built.py
```diff
@@ -0,0 +1,114 @@
...
+
+# from bs4 import BeautifulSoup
+
+REPO_ROOT = Path(__file__).parent.parent
+
+# files that are ok to have 0 min 0 sec time, probably because they don't have any python code
+OK_TO_NOT_RUN = [
+    "beginner/basics/intro.html",
+]
+
```
svekars, line 12: Maybe add a comment:
When the tutorial is fixed, remove it from this list.

.jenkins/validate_tutorials_built.py
```diff
@@ -0,0 +1,101 @@
...
+# when the tutorial is fixed, remove it from this list
+KNOWN_BAD = [
```
mthrok, line 14: Can we call it something other than `bad`?

Some tutorials are intentionally disabled. TTS tutorial is disabled because conflicting dependency issue but the original tutorial is executed on TorchAudio CI and we make sure it works.

mthrok (commented):
The white space changes and docker image changes don't seem essential to the validation of execution time. They should be in separate PRs.

.jenkins/validate_tutorials_built.py
clee2000: which tutorial is this? i can put it in the `ok_to_not_run` list, which isnt really a better name
the end goal is that we fix everything in the known bad list or decide that its ok to to run (b/c it has no code or some other reason)

PR #2042, opened 2022-09-20
[maskedtensor] Add overview tutorial

george-qi:
* #2047
* #2046
* #2045
* #2044
* #2043
* __->__ #2042

### DIFF
--- a/beginner_source/maskedtensor_overview.rst
+++ b/beginner_source/maskedtensor_overview.rst
@@ -1,0 +1,244 @@
+MaskedTensor Overview
+=====================
+
+This tutorial is designed to serve as a starting point for using MaskedTensors
+and discuss its masking semantics.
+
+Using MaskedTensor
+++++++++++++++++++
+
+Construction
+------------
+
+There are a few different ways to construct a MaskedTensor:
+
+* The first way is to directly invoke the MaskedTensor class
+* The second (and our recommended way) is to use :func:`masked.masked_tensor` and :func:`masked.as_masked_tensor` factory functions,
+  which are analogous to :func:`torch.tensor` and :func:`torch.as_tensor`
+
+  .. autosummary::
+    :toctree: generated
+    :nosignatures:
+
+    masked.masked_tensor
+    masked.as_masked_tensor
+
+Accessing the data and mask
+---------------------------
+
+The underlying fields in a MaskedTensor can be accessed through:
+
+* the :meth:`MaskedTensor.get_data` function
+* the :meth:`MaskedTensor.get_mask` function. Recall that ``True`` indicates "specified" or "valid" while ``False`` indicates
+  "unspecified" or "invalid".
+
+In general, the underlying data that is returned may not be valid in the unspecified entries, so we recommend that
+when users require a Tensor without any masked entries, that they use :meth:`MaskedTensor.to_tensor` (as shown above) to
+return a Tensor with filled values.
+
+Indexing and slicing
+--------------------
+
+:class:`MaskedTensor` is a Tensor subclass, which means that it inherits the same semantics for indexing and slicing
+as :class:`torch.Tensor`. Below are some examples of common indexing and slicing patterns:
+
+    >>> data = torch.arange(60).reshape(3, 4, 5)
+    >>> mask = data % 2 == 0
+    >>> mt = masked_tensor(data.float(), mask)
+    >>> mt[0]
+    MaskedTensor(
+      [
+        [  0.0000,       --,   2.0000,       --,   4.0000],
+        [      --,   6.0000,       --,   8.0000,       --],
+        [ 10.0000,       --,  12.0000,       --,  14.0000],
+        [      --,  16.0000,       --,  18.0000,       --]
+      ]
+    )
+    >>> mt[[0,2]]
+    MaskedTensor(
+      [
+        [
+          [  0.0000,       --,   2.0000,       --,   4.0000],
+          [      --,   6.0000,       --,   8.0000,       --],
+          [ 10.0000,       --,  12.0000,       --,  14.0000],
+          [      --,  16.0000,       --,  18.0000,       --]
+        ],
+        [
+          [ 40.0000,       --,  42.0000,       --,  44.0000],
+          [      --,  46.0000,       --,  48.0000,       --],
+          [ 50.0000,       --,  52.0000,       --,  54.0000],
+          [      --,  56.0000,       --,  58.0000,       --]
+        ]
+      ]
+    )
+    >>> mt[:, :2]
+    MaskedTensor(
+      [
+        [
+          [  0.0000,       --,   2.0000,       --,   4.0000],
+          [      --,   6.0000,       --,   8.0000,       --]
+        ],
+        [
+          [ 20.0000,       --,  22.0000,       --,  24.0000],
+          [      --,  26.0000,       --,  28.0000,       --]
+        ],
+        [
+          [ 40.0000,       --,  42.0000,       --,  44.0000],
+          [      --,  46.0000,       --,  48.0000,       --]
+        ]
+      ]
+    )
+
+Semantics
++++++++++
+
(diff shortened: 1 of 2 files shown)


malfet (changes requested):
Please do not use stacks, just submit a regular PR (which otherwise looks good to me)

beginner_source/maskedtensor_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+  "unspecified" or "invalid".
+
+In general, the underlying data that is returned may not be valid in the unspecified entries, so we recommend that
+when users require a Tensor without any masked entries, that they use :meth:`MaskedTensor.to_tensor` (as shown above) to
+return a Tensor with filled values.
+
+Indexing and slicing
+--------------------
+
+:class:`MaskedTensor` is a Tensor subclass, which means that it inherits the same semantics for indexing and slicing
```
jisaacso, line 42: Can you write a docstring on MaskedTensor? 
```
Init signature: MaskedTensor(data, mask, requires_grad=False)
Docstring:      <no docstring>
```

beginner_source/maskedtensor_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+Using MaskedTensor
+++++++++++++++++++
+
+Construction
+------------
+
+There are a few different ways to construct a MaskedTensor:
+
+* The first way is to directly invoke the MaskedTensor class
+* The second (and our recommended way) is to use :func:`masked.masked_tensor` and :func:`masked.as_masked_tensor` factory functions,
```
jisaacso, line 16: is this still true that this is preferred to MaskedTensor?

beginner_source/maskedtensor_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+when users require a Tensor without any masked entries, that they use :meth:`MaskedTensor.to_tensor` (as shown above) to
+return a Tensor with filled values.
+
+Indexing and slicing
+--------------------
+
+:class:`MaskedTensor` is a Tensor subclass, which means that it inherits the same semantics for indexing and slicing
+as :class:`torch.Tensor`. Below are some examples of common indexing and slicing patterns:
+
+    >>> data = torch.arange(60).reshape(3, 4, 5)
```
jisaacso, line 45: can you copy the output of this call and the mask call below as well? It helps the user see the original data before masking.

beginner_source/maskedtensor_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+          [ 40.0000,       --,  42.0000,       --,  44.0000],
+          [      --,  46.0000,       --,  48.0000,       --]
+        ]
+      ]
+    )
+
+Semantics
++++++++++
+
+MaskedTensor vs NumPy's MaskedArray
```
jisaacso, line 95: I feel like this can be moved to the end? The flow of the tutorial should be:

what is a MT
How do you access/slice
What operations can you do / reductions
How does this compare with alternatives to masked tensor

beginner_source/maskedtensor_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+    >>> mt2
+    MaskedTensor(
+      [      --,       --,       --,       --,       --]
+
+.. _reduction-semantics:
+
+Reduction semantics
+-------------------
+
+The basis for reduction semantics `has been documented and discussed at length <https://github.com/pytorch/rfcs/pull/27>`__,
```
jisaacso, line 168: I feel like this can be rephrased.

The purpose of a tutorial is to be a self contained, helpful, end to end showcase of a new feature. It is not assumed that the reader has seen RFCs. 

I would start off explaining what a reduction semantic is (we ignore masked values for your favorite functions!), then give the code examples before, then conclude by saying "for more details on the basis for reduction semantics, see this RFC".

beginner_source/maskedtensor_overview.rst
jisaacso, line 45: minor nit, but you can reduce the size of data tensor to make the outputs more readable. Maybe a 2x2x2 tensor instead of 3x4x5

beginner_source/maskedtensor_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+but again, by way of example:
+
+    >>> data = torch.arange(12, dtype=torch.float).reshape(3, 4)
```
jisaacso, line 171: similar to above comment, i think it's helpful to print out the outputs of these data and mask calls.

jisaacso (changes requested):
As a noob, I'm pretty confused by the focus of this on comparisons to Numpy.

IMO, the first overview tutorial should not speak of other libraries. There can be a whole separate tutorial covering "comparisons to numpy". It's just too confusing for a new user to think about alternative semantics that they don't need.

I would personally pull out all of the numpy conversation into a separate tutorial, and focus this one on building up knowledge: what is a MT, how are they instantiated, How are the visualized, how do you do reductions, how do you do combine MTs via ops like sum, and then you can touch on the requirement that masks have the same shape (and that you can get around this with an &). If you want, you can end the whole tutorial with a single sentence, "for those familiar with numpy, we recognize the semantics deviate from Numpy's. Details on why are addressed in <tutorial>"

beginner_source/maskedtensor_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+    )
+    >>> mt0 + mt1
+    ValueError: Input masks must match. If you need support for this, please open an issue on Github.
+
+However, if this behavior is desired, MaskedTensor does support these semantics by giving access to the data and masks
+and conveniently converting a MaskedTensor to a Tensor with masked values filled in using :func:`to_tensor`.
+
+    >>> t0 = mt0.to_tensor(0)
+    >>> t1 = mt1.to_tensor(0)
+    >>> mt2 = masked_tensor(t0 + t1, mt0.get_mask() & mt1.get_mask())
```
jisaacso, line 154: Why is the `&` instead of `|`?

It feels like to me, that combining two masks should be the union operator of those two masks, not the &?

jisaacso (commented):
adding one more comment

beginner_source/maskedtensor_overview.rst
jisaacso, line 154: is there a way to union two masks and return another mask, without returning a boolean? Right now

```python
>>> mt0.get_mask() & mt1.get_mask()
tensor([[False, False, False, False],
        [False, False, False, False],
        [False, False, False, False]])

# it would be nice if union ops on two raw MTs returned a MT
>>> mt0 | mt1
MaskedTensor(
  [
    [  --,       --,       --,       --],
    [      --,   --,       --,       --],
    [      --,   --,       --,       --]
  ]
)
```

beginner_source/maskedtensor_overview.rst
george-qi: Discussed offline, yes it's still true

beginner_source/maskedtensor_overview.rst
george-qi: Done!

george-qi: Agree this was a bit much. Reduced to 2x3x4 so I can still show some slightly advanced indexing.. honestly not quite sure this helped enough, so please let me know if this was enough

beginner_source/maskedtensor_overview.rst
george-qi: Discussed offline - this will include "what is a MT and why is an MT useful"

beginner_source/maskedtensor_overview.rst
george-qi: NumPy uses `logical_or` and our mask is the inverse of theirs, so ours is `&` if  you want the same behavior

george-qi: Yeah, we could add this in!

PR #2043, opened 2022-09-20
[maskedtensor] Add sparsity tutorial

george-qi:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #2047
* #2046
* #2045
* #2044
* __->__ #2043
* #2042

PR #2045, opened 2022-09-20
[maskedtensor] Add safe softmax tutorial

george-qi:
* #2047
* #2046
* __->__ #2045
* #2044
* #2043
* #2042

### DIFF
--- a/beginner_source/maskedtensor_safe_softmax.rst
+++ b/beginner_source/maskedtensor_safe_softmax.rst
@@ -1,0 +1,33 @@
+Safe Softmax
+------------
+
+One of the issues that frequently comes up is the necessity for a safe softmax -- that is, if there is an entire
+batch that is "masked out" or consists entirely of padding (which, in the softmax case, translates to being set `-inf`),
+then this will result in NaNs, which can leading to training divergence. For more detail on why this functionality
+is necessary, please find refer to 
+`Issue 55056 - Feature Request for Safe Softmax <https://github.com/pytorch/pytorch/issues/55056>`__.
+
+Luckily, :class:`MaskedTensor` has solved this issue:
+
+    >>> data = torch.randn(3, 3)
+    >>> mask = torch.tensor([[True, False, False], [True, False, True], [False, False, False]])
+    >>> x = data.masked_fill(~mask, float('-inf'))
+    >>> mt = masked_tensor(data, mask)
+
+PyTorch result:
+
+    >>> x.softmax(0)
+    tensor([[0.3548,    nan, 0.0000],
+            [0.6452,    nan, 1.0000],
+            [0.0000,    nan, 0.0000]])
+
+:class:`MaskedTensor` result:
+
+    >>> mt.softmax(0)
+    MaskedTensor(
+      [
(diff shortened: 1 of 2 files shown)


beginner_source/maskedtensor_safe_softmax.rst
```diff
@@ -0,0 +1,33 @@
...
+
+One of the issues that frequently comes up is the necessity for a safe softmax -- that is, if there is an entire
+batch that is "masked out" or consists entirely of padding (which, in the softmax case, translates to being set `-inf`),
+then this will result in NaNs, which can leading to training divergence. For more detail on why this functionality
+is necessary, please find refer to 
+`Issue 55056 - Feature Request for Safe Softmax <https://github.com/pytorch/pytorch/issues/55056>`__.
+
+Luckily, :class:`MaskedTensor` has solved this issue:
+
+    >>> data = torch.randn(3, 3)
```
jisaacso, line 12: I would explain a bit more here or inline what you are doing:

Let's create a Tensor with an "unsafe" column, that is a column of all -inf padding values, for which PyTorch's traditional softmax will result in nans.

<do it and show nan>

Now let's create a safeter MaskedTensor from this Tensor, and show that the resultant values are masked out

beginner_source/maskedtensor_safe_softmax.rst
```diff
@@ -0,0 +1,33 @@
...
+PyTorch result:
+
+    >>> x.softmax(0)
+    tensor([[0.3548,    nan, 0.0000],
+            [0.6452,    nan, 1.0000],
+            [0.0000,    nan, 0.0000]])
+
+:class:`MaskedTensor` result:
+
+    >>> mt.softmax(0)
```
jisaacso, line 26: I'm confused on this result. The softmax above returns two different values that are getting masked here: 0.0 and nan. Do we really want to mask out both outputs? I assume the idea is that we want to mask all input padding (aka mask all -infs). but in situations of unstable training, do 0's make a difference?

jisaacso:
Another high level comment similar to other tutorials -- I think this tutorial makes more sense as one component of a larger tutorial that showcases end to end training. At first glance, it's not intuitive to me why I would care to end with a MaskedTensor after softmax vs ending with a tensor. You do reference an issue on unstable training for all nan examples in a batch, but I feel like *showing* that first hand as part of e2e training would be more powerful.

beginner_source/maskedtensor_safe_softmax.rst
george-qi: I don't think the 0's make a difference in unstable training, but maybe I'm not quite understanding the question. In general for MaskedTensor, if the user has deemed a particular index as "unspecified", then it will remain unspecified or masked out in both the forward as well as the backward. And so in this case, while the underlying data that is being masked out would have been different (i.e. 0.0 vs nan), they're both unspecified and so we still want to mask them out. 

For training specifically, I agree that leaving the 0's would be identical, but I think it would contradict MaskedTensor semantics and become confusing (e.g. wouldn't know when to expect masked out values vs not.)

PR #2046, opened 2022-09-20
[maskedtensor] Add missing nan ops tutorial

george-qi:
* (to be filled)

### DIFF
--- a/beginner_source/maskedtensor_missing_nan_ops.rst
+++ b/beginner_source/maskedtensor_missing_nan_ops.rst
@@ -1,0 +1,32 @@
+Implementing missing torch.nan* operators
+-----------------------------------------
+
+In the above issue, there is a request to add additional operators to cover the various `torch.nan*` applications,
+such as ``torch.nanmax``, ``torch.nanmin``, etc.
+
+In general, these problems lend themselves more naturally to masked semantics, so instead of introducing additional
+operators, we propose using MaskedTensors instead. Since
+`nanmean has already landed <https://github.com/pytorch/pytorch/issues/21987>`__, we can use it as a comparison point:
+
+    >>> x = torch.arange(16).float()
+    >>> y = x * x.fmod(4)
+    >>> y = y.masked_fill(y ==0, float('nan'))
+    >>> y
+    tensor([nan,  1.,  4.,  9., nan,  5., 12., 21., nan,  9., 20., 33., nan, 13.,
+            28., 45.])
+    >>> y.nanmean()
+    tensor(16.6667)
+    >>> torch.mean(masked_tensor(y, ~torch.isnan(y)))
+    MaskedTensor( 16.6667, True)
+
+:class:`MaskedTensor` can also support reductions when the data is fully masked out, which is equivalent
+to the case above when the data Tensor is completely ``nan``. ``nanmean`` would return ``nan``
+(an ambiguous return value), while MaskedTensor would more accurately indicate a masked out result.
+
+    >>> x = torch.empty(16).fill_(float('nan'))
+    >>> x
+    tensor([nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan])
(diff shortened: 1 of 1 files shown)


beginner_source/maskedtensor_missing_nan_ops.rst
```diff
@@ -0,0 +1,32 @@
...
+
+    >>> x = torch.arange(16).float()
+    >>> y = x * x.fmod(4)
+    >>> y = y.masked_fill(y ==0, float('nan'))
+    >>> y
+    tensor([nan,  1.,  4.,  9., nan,  5., 12., 21., nan,  9., 20., 33., nan, 13.,
+            28., 45.])
+    >>> y.nanmean()
+    tensor(16.6667)
+    >>> torch.mean(masked_tensor(y, ~torch.isnan(y)))
```
jisaacso, line 19: is the goal to have sequential tutorials or keep each self contained? If the latter, can you add the relevant imports up top.

jisaacso, line 19: did you replace `masked_tensor` with `MaskedTensor` in https://github.com/pytorch/pytorch/commit/5e9c26c8e25e5d5be18ff98b7808b674b1e7a0a5 ? If so, can you update the tutorial here?

jisaacso, line 19: have we considered API sugar:

(1) instantiating a MT from a Tensor assuming na is the mask

```python
>>> MaskedTensor(y)
MaskedTensor(
  [      --,   1.0000,   4.0000,   9.0000,       --,   5.0000,  12.0000,  21.0000,       --,   9.0000,  20.0000,  33.0000,       --,  13.0000,  28.0000,  45.0000]
)
```

(2) instantiating a MT where user just states the mask value instead of passing the mask

```python
y = MaskedTensor(y, mask_value=float(1))
```

jisaacso (changes requested):
Please address comments on masked_tensor vs MaskedTensor. I'm also not sure that this tutorial is all that helpful as a standalone? Would it make more sense to just merge this into the overall tutorial? Or have a second tutorial that aggregate advanced features? This feels like a pretty minor demonstration of nanmean with Tensors vs mean with MT.

beginner_source/maskedtensor_missing_nan_ops.rst
george-qi: `masked_tensor` is the preferred function to use!

george-qi: This tutorial will be merged with overview!

beginner_source/maskedtensor_missing_nan_ops.rst
```diff
@@ -0,0 +1,32 @@
...
+operators, we propose using MaskedTensors instead. Since
+`nanmean has already landed <https://github.com/pytorch/pytorch/issues/21987>`__, we can use it as a comparison point:
+
+    >>> x = torch.arange(16).float()
+    >>> y = x * x.fmod(4)
+    >>> y = y.masked_fill(y ==0, float('nan'))
+    >>> y
+    tensor([nan,  1.,  4.,  9., nan,  5., 12., 21., nan,  9., 20., 33., nan, 13.,
+            28., 45.])
+    >>> y.nanmean()
```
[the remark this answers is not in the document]
george-qi: Not sure..

beginner_source/maskedtensor_missing_nan_ops.rst
george-qi: Not yet! I think an unspecified mask could also be an indication that they would like all `True` values for the mask, so that could be a third option as well.

Another one would be to allow for just `MaskedTensor(y)` if `y` is a sparse tensor because then the mask is "implied".

All been discussed and will take note to add in :)

jisaacso: where are you tracking feature requests?

PR #2058, opened 2022-09-27
Update cifar10_tutorial.py

junecv:
fixed running loss per iteration

### DIFF
--- a/beginner_source/blitz/cifar10_tutorial.py
+++ b/beginner_source/blitz/cifar10_tutorial.py
@@ -182,7 +182,7 @@ for epoch in range(2):  # loop over the dataset multiple times
         # print statistics
         running_loss += loss.item()
         if i % 2000 == 1999:    # print every 2000 mini-batches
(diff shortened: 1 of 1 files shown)


beginner_source/blitz/cifar10_tutorial.py
```diff
@@ -182,7 +182,7 @@ def forward(self, x):
         # print statistics
         running_loss += loss.item()
         if i % 2000 == 1999:    # print every 2000 mini-batches
-            print(f'[{epoch + 1}, {i + 1:5d}] loss: {running_loss / 2000:.3f}')
+            print(f'[{epoch + 1}, {i + 1:5d}] loss: {running_loss / (i + 1):.3f}')
```
malfet, line 185: This feels wrong unless you delete the next line that resets running_loss

malfet (changes requested):
`running_loss` is accumulated over 2000 iterations, should it should be divided by 2000, isn't it?

junecv:
You're right. I wasn't aware of the reset line in the original tutorial.

junecv:
Thank you for your time!

PR #2069, opened 2022-10-06
Update gym to v0.26.1

svekars:


arjun-kg (changes requested):
Thank you for the PR. I believe more changes are needed. I'm unable to comment inline for some of them so I'll add them here - 

1. The `nes-py` environment it depends on has not been updated to 0.26. We can use a compatibility wrapper in the meantime, and change the make statement like below -  
```python
env = gym_super_mario_bros.make("SuperMarioBros-1-1-v0", apply_api_compatibility=True)
```

2. `env.reset()` now returns two values, please change the reset statement to to 
`state, _ = env.reset()` 

3. Please replace any place that has 4 step returns (obs, reward, done, info), to 5 step returns (obs, reward, terminated, truncated, info)
4. The replay buffer should hold `terminated` instead of `done`. It would be more clear if we replaced `done` with `terminated` in all the references with respect to replay buffer, and bootstrapping.
5. In places where episode end is checked, please change `if done` to `if (terminated or truncated)`

pseudo-rnd-thoughts:
Hey, is there any progress on this PR? Im happy to finish it if necessary

svekars:
I'm closing this for now and this will be handled separately.

jkterry1:
@svekars is there a new PR for this?

Issue #2065, opened 2022-10-03
AUTOMATIC DIFFERENTIATION WITH TORCH.AUTOGRAD - Jacobian Product Tutorial Shape Confusion


nb8g13:
In the tutorial **AUTOMATIC DIFFERENTIATION WITH TORCH.AUTOGRAD** a working toy example is presented in **Optional Reading: Tensor Gradients and Jacobian Products**. 

However, the shape of the input tensor `inp` is used when specifying the vector $v$ in the the backwards call instead of the shape of the output tensor `out`, as highlighted in the image below. 

![image](https://user-images.githubusercontent.com/25404404/193595920-faa1adbe-1620-4d21-8170-80055090ff77.png)

This code runs correctly because the input and output tensors happen to be the same size in this example. However, if the output tensor is a different shape then using the input shape in the backwards() call leads to an exception. I believe it would be more instructive to use `out` in place of `inp` in the positions highlighted. This would also better match the text above the code snippet.

cc @suraj813

malfet:
Thank you for being a thorough reader. Indeed, it would be better to use `out` as argument to `tensor.ones_like`

PR #2071, opened 2022-10-07
Update `Tensor Gradients and Jacobian Products` example

malfet:
Shape of gradient argument of `x.backwards` method should be that of `x`.

Modify example, to make input tensor a 5x4 matrix and transpose it in the output

Fixes https://github.com/pytorch/tutorials/issues/2065


albanD (approved):
Yep.

PR #2073, opened 2022-10-07
update rl tutorial to last gym version

younik:
Update the DQN tutorial to match the last version of gym

### DIFF
--- a/intermediate_source/reinforcement_q_learning.py
+++ b/intermediate_source/reinforcement_q_learning.py
@@ -79,7 +79,7 @@ import torch.nn.functional as F
 import torchvision.transforms as T
 
 
-env = gym.make('CartPole-v0', new_step_api=True, render_mode='single_rgb_array').unwrapped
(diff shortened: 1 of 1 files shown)


intermediate_source/reinforcement_q_learning.py
```diff
@@ -79,7 +79,7 @@
 import torchvision.transforms as T
 
 
-env = gym.make('CartPole-v0', new_step_api=True, render_mode='single_rgb_array').unwrapped
+env = gym.make('CartPole-v0', render_mode='rgb_array').unwrapped
```
malfet, line 82: This will make it compatible with both v0.25(which is installed on Colab) and new gym version
```suggestion
if gym.__version__ < '0.26':
    env = gym.make('CartPole-v0', new_step_api=True, render_mode='single_rgb_array').unwrapped
else:
    env = gym.make('CartPole-v0', render_mode='rgb_array').unwrapped
```

malfet (approved):

Issue #2090, opened 2022-10-20
ax_multiobjective_nas_tutorial.py is failing against 1.13


svekars:
[ax_multiobjective_nas_tutorial.py](https://github.com/pytorch/tutorials/blob/master/intermediate_source/ax_multiobjective_nas_tutorial.py) is failing against 1.13 binaries. More details [here](https://app.circleci.com/pipelines/github/pytorch/tutorials/6772/workflows/e5ee8793-cec0-4a60-a9f5-924e2ad4cea0/jobs/132480?invite=true#step-104-2552). 

We will disable this tutorial until this issue is resolved.

cc: @Balandat @malfet

Balandat:
Thanks for the heads up.

This seems to be an issue with TorchX which appears to fail on import? @d4l3k are there known incompatibilities between torchx and torch 1.13 (looks like the torchx version installed in CI is `0.2.0`).

Balandat:
cc @dme65, @esantorella

d4l3k:
The issue isn't TorchX or torch 1.13 -- importlib_metadata broke backwards compatibility with Python 3.7

https://github.com/python/importlib_metadata/issues/411

We've had to pin <5.0 in trunk https://github.com/pytorch/torchx/blob/main/dev-requirements.txt#L10 to resolve the issue

Balandat:
Great, thanks for the quick response. Let me put up a PR to add that pin also to the reqs for the tutorials.

Balandat:
@svekars I merged the fix, I assume this can be closed? Or is there anything that needs to be done to re-enable the tutorial?

PR #2091, opened 2022-10-20
Pin importlib_metadata<5.0 in requirements.txt

Balandat:
v5.0.0 causes issues with python 3.7: https://github.com/python/importlib_metadata/issues/411

This was raised in the context of https://github.com/pytorch/tutorials/issues/2090

This PR pins the version to <5.0 to circumvent this. Maybe a better fix would be to bump the python version in CI to 3.8, but that's something to discuss more broadly.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,8 +30,8 @@ nbformat>=4.2.0
 deep_phonemizer==0.0.17
 
 # the following is necessary due to https://github.com/python/importlib_metadata/issues/411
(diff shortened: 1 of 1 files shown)


svekars:
@malfet any objections?

malfet (approved):
Looks good to me, but I wonder if we can do something like (see [PEP-508](https://peps.python.org/pep-0508/) ):
```
importlib-metadata<5.0; python-version<='3.7'
importlib-metadata; python-version>'3.7'
```

PR #2110, opened 2022-11-02
update target network weight during episode after x timesteps

Mdhvince:
Currently the target network remain fixed during the entire episode.
In this PR I have moved the update of the weigths inside the inner loop of the training. So the update can occure during the episode.

Without this change, if the episode is too long, the target network will be too different from the behavior network.

cc @vmoens @nairbv


vmoens (approved):
LGTM thanks for spotting this!

Mdhvince:
You welcome, thanks to you 🔥

PR #2124, opened 2022-11-21
[Quant] Add tutorial for BackendConfig

andrewor14:
Summary: This commit adds the tutorial for the BackendConfig,
the integration point for backend developers to specify the
quantization behavior on a given target backend.

Note to reviewers: To see the rendered version, just go to
"Files changed" > "..." > "View file".

Reviewers: jerryzh168, vkuzo

Subscribers: jerryzh168, vkuzo


prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,410 @@
...
+- `DTypeConfig <https://pytorch.org/docs/stable/generated/torch.ao.quantization.backend_config.DTypeConfig.html>`__:
+  Config object that specifies the supported data types and constraints
+  (if any) for input and output activations, weights, and biases. Each
+  BackendPatternConfig consists of one or more of these.
+- `DTypeWithConstraints <https://pytorch.org/docs/stable/generated/torch.ao.quantization.backend_config.DTypeWithConstraints.html>`__:
+  Constraints imposed by the backend on the quantization parameters
+  (scale and zero point) and ranges when quantizing to a given data
+  type. Each DTypeConfig consists of many of these.
+
+Pattern Specification
```
vkuzo, line 35: I remember that we decided a few months ago we cannot ship the current pattern format to users as it's confusing.  Do we have any thoughts on when we plan to switch to a more intuitive format?

prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,410 @@
...
+    
+    # For quantizing Linear
+    linear_config = BackendPatternConfig(torch.nn.Linear) \
+        .set_observation_type(ObservationType.OUTPUT_USE_DIFFERENT_OBSERVER_AS_INPUT) \
+        .add_dtype_config(weighted_int8_dtype_config) \
+        .set_root_module(torch.nn.Linear) \
+        .set_qat_module(torch.nn.qat.Linear) \
+        .set_reference_quantized_module(torch.ao.nn.quantized.reference.Linear)
+    
+    # For fusing Conv2d + ReLU into ConvReLU2d
```
vkuzo, line 208: I remember we decided that we shouldn't expose patterns in untuitive formats.  If the pattern is "conv-relu", the API should expect "conv-relu", not "relu-conv".

prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,410 @@
...
+        QConfigMapping,
+    )
+    from torch.ao.quantization.backend_config import (
+        BackendConfig,
+        BackendPatternConfig,
+        DTypeConfig,
+        DTypeWithConstraints,
+        ObservationType,
+    )
+    from torch.ao.quantization.fuser_method_mappings import _reverse_sequential_wrapper2
```
vkuzo, line 178: I remember we talked about not exposing things like `_reverse_sequential_wrapper2` to the user, can we remove this before adding tutorials?

prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,410 @@
...
+    # ======================
+    
+    quint8_with_constraints = DTypeWithConstraints(
+        dtype=torch.quint8,
+        quant_min_lower_bound=0,
+        quant_max_upper_bound=255,
+        scale_min_lower_bound=2 ** -12,
+    )
+    
+    weighted_int8_dtype_config = DTypeConfig(
```
vkuzo, line 194: can we clarify what does `dtype` mean here?

For `QConfig`, we aligned that `dtype` means "the dtype argument to a quantize function in reference model spec".  Does it have exactly the same meaning here?

prototype_source/backend_config_tutorial.rst
andrewor14: Ok, I can inline this

prototype_source/backend_config_tutorial.rst
andrewor14: Yes, we had a design review on this but the implementation was deprioritized due to other projects. We should revisit the priority of this next half since changing the frontend should be relatively low effort

prototype_source/backend_config_tutorial.rst
andrewor14: I agree. This is part of the broader effort I mentioned [in this comment](https://github.com/pytorch/tutorials/pull/2124#discussion_r1028507042).

prototype_source/backend_config_tutorial.rst
andrewor14: Yes, it means these are the dtypes that are supported by the backend for this pattern. "Supported" here means the QConfig must specify the same `dtype` for the pattern, otherwise it will be ignored

prototype_source/backend_config_tutorial.rst
vkuzo: I think it makes sense for fixes like these to block user facing documents.  We really shouldn't be shipping stuff like this to OSS users.

prototype_source/backend_config_tutorial.rst
vkuzo: > Ok, I can inline this

I would imagine that we actualy remove this from the codebase and expose a more user friendly API, not just in this tutorial

prototype_source/backend_config_tutorial.rst
vkuzo: I think that effort should be completed first, and then we do user marketing.  We shouldn't just forget/deprioritize user friendliness improvements like this one.

vkuzo (changes requested):
the tutorial looks good, but requesting changes for cleaning up the user API.  I think we need to make sure this is user friendly before going further.

andrewor14:
Ok, I updated the tutorial to use the simple format now that #90351 is merged. This should address the comments from the last round of review. Please have a look @vkuzo and @jerryzh168.

prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,398 @@
...
+each supported operator pattern (e.g. linear, conv-bn-relu). The API is
+broken down into the following class hierarchy:
+
+- `BackendConfig <https://pytorch.org/docs/stable/generated/torch.ao.quantization.backend_config.BackendConfig.html>`__:
+  The main class to be passed to prepare and convert functions
+- `BackendPatternConfig <https://pytorch.org/docs/stable/generated/torch.ao.quantization.backend_config.BackendPatternConfig.html>`__:
+  Config object that specifies quantization behavior for a given
+  operator pattern. Each BackendConfig consists of many of these.
+- `DTypeConfig <https://pytorch.org/docs/stable/generated/torch.ao.quantization.backend_config.DTypeConfig.html>`__:
+  Config object that specifies the supported data types and constraints
```
vkuzo, line 27: This looks like it is describing interface_dtype.  I thought the `dtype` in `BackendConfig` represented "dtype argument to quantize op in reference model spec", which may or may not equal to interface_type.  Thoughts?  cc @jerryzh168

prototype_source/backend_config_tutorial.rst
vkuzo: I think this should be clarified.  When a user reads `input_dtype`, I'd bet some money they will think "the actual dtype of the input tensor".  I don't think this is what this `dtype` argument means, it is defined in terms of the reference model spec.

prototype_source/backend_config_tutorial.rst
jerryzh168: yeah I think we should be talking about the dtype for the output Tensor of quantize op in the reference model spec for now

prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,400 @@
...
+for a slightly more complicated usage.
+
+set_fused_module
+^^^^^^^^^^^^^^^^
+
+This is used to identify fused weighted modules (e.g.
+``torch.ao.nn.intrinsic.LinearReLU``) that need to be converted to
+reference quantized modules.
+
+Data Type Restrictions
```
vkuzo, line 101: This still reads like "interface_dtype".  I feel like this section should have an example with an expected reference model, and an explicit mention that the dtype here is the dtype argument to the quantize function inside the reference model.

Check out `observer.py` which is more explicit in defining what dtype means inside of it.

vkuzo (approved):
lg, but I think we should clarify what DType means here

prototype_source/backend_config_tutorial.rst
andrewor14: Ok, I'll fix the wording here too to be more explicit

prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,400 @@
...
+        output_dtype=quint8_with_constraints,
+        weight_dtype=torch.qint8,
+        bias_dtype=torch.float)
+
+   def fuse_conv2d_relu(is_qat, conv, relu):
+       """Return a fused ConvReLU2d from individual conv and relu modules."""
+       return torch.ao.nn.intrinsic.ConvReLU2d(conv, relu)
+    
+    # For quantizing Linear
+    linear_config = BackendPatternConfig(torch.nn.Linear) \
```
jerryzh168, line 188: nit: maybe use`set_pattern` to keep consistent, and add a follow up PR to remove the constructor argument?

prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,400 @@
...
+        input_dtype=quint8_with_constraints,
+        output_dtype=quint8_with_constraints,
+        weight_dtype=torch.qint8,
+        bias_dtype=torch.float)
+
+   def fuse_conv2d_relu(is_qat, conv, relu):
+       """Return a fused ConvReLU2d from individual conv and relu modules."""
+       return torch.ao.nn.intrinsic.ConvReLU2d(conv, relu)
+    
+    # For quantizing Linear
```
jerryzh168, line 187: I think we can add a bit more explanation here about what this config means exactly, we have docs before for each field, but explaining them in a real example might be more helpful to users I think, same for other configs we have

jerryzh168: especially for observation_type and dtype_config, since these two things we might reuse in the new flow as well.

please let me know if you need some clarification for dtype_config, I think that one is a bit tricky to understand

prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,400 @@
...
+broken down into the following class hierarchy:
+
+- `BackendConfig <https://pytorch.org/docs/stable/generated/torch.ao.quantization.backend_config.BackendConfig.html>`__:
+  The main class to be passed to prepare and convert functions
+- `BackendPatternConfig <https://pytorch.org/docs/stable/generated/torch.ao.quantization.backend_config.BackendPatternConfig.html>`__:
+  Config object that specifies quantization behavior for a given
+  operator pattern. Each BackendConfig consists of many of these.
+- `DTypeConfig <https://pytorch.org/docs/stable/generated/torch.ao.quantization.backend_config.DTypeConfig.html>`__:
+  Config object that specifies the supported data types passed as
+  arguments to quantize ops, for input and output activations,
```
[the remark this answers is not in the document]
andrewor14: Ok, this section is meant to be a quick summary of each class but I can add the example in the "Data Type Restrictions" section below

prototype_source/backend_config_tutorial.rst
andrewor14: Ok, added a comment here and 2 paragraphs in the DTypeConfig section to explain this further

prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,447 @@
...
+  (scale and zero point) and ranges when quantizing to a given data
+  type. Each DTypeConfig consists of many of these.
+
+The pattern specified in BackendPatternConfig follows the format
+described `here <https://github.com/pytorch/pytorch/blob/master/torch/ao/quantization/backend_config/README.md#pattern-specification>`__.
+
+BackendPatternConfig Specification
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+set_observation_type
```
jerryzh168, line 43: wondering is there overlap with the docstring of BackendConfig?

prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,447 @@
...
+        weight_dtype=torch.qint8,
+        bias_dtype=torch.float)
+
+   def fuse_conv2d_relu(is_qat, conv, relu):
+       """Return a fused ConvReLU2d from individual conv and relu modules."""
+       return torch.ao.nn.intrinsic.ConvReLU2d(conv, relu)
+    
+    # The following BackendPatternConfig refers to static quantization for the
+    # linear op. This inserts quant-dequant ops before and after the float linear
+    # op, producing the following reference model spec:
```
jerryzh168, line 213: nit: `model spec` -> `model`

prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,447 @@
...
+    # and output for the linear op, so the quantization params passed to the two
+    # quantize ops will be different. This is commonly the case for weighted ops
+    # like linear and conv.
+    #
+    # The input dtype specified in the DTypeConfig will be passed as the dtype
+    # argument to quant1, while the output dtype will be passed as the dtype
+    # argument to quant2. If the output dtype is fp32, as in the case of dynamic
+    # quantization, then the output quant-dequant pair will not be inserted.
+    #
+    # Ultimately, how this reference model spec will be fused will depend on
```
jerryzh168, line 227: nit: spec -> model

prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,447 @@
...
+    # how the backend lowers the model. For static quantization, for example,
```
jerryzh168, line 228: maybe: how the backend lowers the model --> what are the supported quantized operators in the backend

prototype_source/backend_config_tutorial.rst
```diff
@@ -0,0 +1,447 @@
...
+    # the backend may fuse the reference pattern in the square brackets into
+    # a single quantized linear op.
+
+    linear_config = BackendPatternConfig() \
```
jerryzh168, line 232: another thing I want to mention is that we can structure the 3 configs to linear and conv_relu, something like:

let's say the target backend has two operators: quantized::linear and quantized::conv2d_relu, 
1. here is the reference pattern for quantized::linear, 
2. here are the steps we'll get this pattern
3. here is the BackendPatternConfig

and do the same for quantized::conv2d_relu op

This would help backend developers to get the right mental model of how they can come up with the backend pattern config for the target quantized operator I think

jerryzh168 (approved):
lg, one comment about adding more docs to talk about the process of how people come up with BackendPatternConfig for each op

andrewor14:
Ok, I think I addressed all the comments. Thank you both for your feedback! I'm merging this.

andrewor14:
Hi @svekars, could you help me merge this? Thanks.

PR #2132, opened 2022-11-30
fix on cifar10 tutorial for windows platform

iremyux:
A fix to the cifar10_tutorial.py file is proposed resolving [issue #12645](https://github.com/pytorch/pytorch/issues/12645) on pytorch repo. This fix will allow the tutorial to run correctly without needing any modification on Windows. For other platforms, there should not be any difference. To keep the structure and flow of the tutorial the same, the necessary if statement was added on top.

cc @suraj813


beginner_source/blitz/cifar10_tutorial.py
```diff
@@ -55,313 +55,321 @@
...
+# https://docs.python.org/2/library/multiprocessing.html#windows 
 
+if __name__ == '__main__': 
 
-def imshow(img):
-    img = img / 2 + 0.5     # unnormalize
-    npimg = img.numpy()
-    plt.imshow(np.transpose(npimg, (1, 2, 0)))
-    plt.show()
+    import torchg
```
svekars, line 66:
```suggestion
    import torch
```

suraj813:
Hi @iremyux welcome and thanks for making your first PR. I don't think this change is needed because a) the tutorial contains a note on what to do (i.e. change num_workers=0) for this exact issue, and b) it is not conventional practice to wrap an entire script within the if block. Closing the issue, feel free to reopen if you think otherwise.

PR #2144, opened 2022-12-08
Fix a display bug which cannot update figure correctly

Yichi-Lionel-Cheung:
When using notebook such as colab and local jupyter where `is_ipython == True`, the training procedure is not shown correctly. The figure is closed too fast to observe since `clear_output(wait=True)` is always activated.

Excanging `display.display(plt.gcf())` and `display.clear_output(wait=True)` could fix the bug.

In addition, after training the model, the resulting figure contains all episodes shall be displaied.

cc @vmoens @nairbv


vmoens:
LGTM 
@svekars: can we wait to land this until we merge #2030 or #2145?
This refactoring has been waiting for a long time and it's a recurrent concern on the pytorch forum...
@Yichi-Lionel-Cheung it should not take long :) Hope that's ok

Yichi-Lionel-Cheung:
> 

Sure!

svekars:
@Yichi-Lionel-Cheung can you please resolve the conflict? Thanks!

PR #2146, opened 2022-12-08
updated documentation

shyamgupta196:
this will look better on the site


svekars:
We have disabled this tutorial build as of now because of build issues. We can review this PR when other issues with the tutorial are fixed and it's building again.

PR #2152, opened 2022-12-12
Update translation_transformer.ipynb  env packages

iamdanialkamali:
Add some required packages and file download scripts to make the notebook runnable without problems. Before this change, it raised an error for the lack of data ` [E050] Can't find model 'de_core_news_sm.' It doesn't seem to be a Python package or a valid path to a data directory.` And raised the `torchdata` requirement error. Should I have to change anywhere else for this change as well, please tell me in this PR.


svekars:
This is a wrong branch to propose changes to. Please submit your PR against the master branch.

PR #2153, opened 2022-12-13
Enable key navigation

mthrok:
This commits allows navigating the tutorials with arrow keys ⇦ ⇨.


svekars (commented):
Thanks, Moto! Great idea! But it does not look like it's working in the preview. Is this expected?

mthrok:
> Thanks, Moto! Great idea! But it does not look like it's working in the preview. Is this expected?

Not sure how Netlify works, but once it's deployed it should work. https://pytorch.org/audio/stable/index.html

PR #2166, opened 2023-01-07
fixes bug when calculating len of dataloaders in translation tutorial

GabrielDornelles:
I was following this [tutorial](https://pytorch.org/tutorials/beginner/translation_transformer.html) on pytorch.org and faced this small error when getting ```len``` of the train and val ```Dataloader``` instances.

Error:
```
TypeError: _IterDataPipeSerializationWrapper instance doesn't have valid length
```

Since this is an ```IterDataPipe``` length cannot be directly inferred. The proposed solution fixes this error although it's doesn't calculate accurately the dataset sizes.

As informed in the ```Multi30k``` dataset method, the split is:

  - train: 29000
  - valid: 1014
  - test: 1000

Of course one batch is smaller than ```BATCH_SIZE```, so calculating it directly gives sizes:
- train: 29056
- valid: 1024

It doesn't affect anything other than logging, so I think it's ok. But if someone has a better approach I'll add it.


svekars:
Thank you for contributing to PyTorch tutorials. Changes need to be merged to the main branch rather than master branch. I updated the base branch to main but there is a conflict. Can you please either resolve the conflict to align with the latest main branch or close this PR and create a new one against the main branch?

GabrielDornelles:
Solved the conflicts

Nayef211:
Hey @GabrielDornelles thanks for creating this PR! I think the best way to add this fix would be to add a `LengthSetterIterDataPipe` to the [Multi30k dataset](https://github.com/pytorch/text/blob/main/torchtext/datasets/multi30k.py#L130) within torchtext. This way we can directly call `len(dp)` on the datapipe without having to materialize all of the data first, which is what we do currently when calling `len(list(dp))`.

This is being tracked in https://github.com/pytorch/text/issues/1943. You're welcome to add this change to torchtext if you'd like!

PR #2170, opened 2023-01-11
Replaces `gym` with `gymnasium` in reinforcement learning q-learning tutorial

pseudo-rnd-thoughts:
Gymnasium is a fork of OpenAI gym that is developed by the same team that maintained Gym since v0.19-v0.26

You can read more about the reason for the fork in https://farama.org/Announcing-The-Farama-Foundation

In this PR, I have replaced the instances of Gym with gymnasium and removed old code that provided support for Gym v0.25

@vmoens Let me know if there is anything else you would like me to change


vmoens (approved):
LGTM thanks for this change

PR #2171, opened 2023-01-13
Add torch.func tutorials for PyTorch 2.0

zou3519:
As the final step of integrating functorch into PyTorch, we have move the functorch APIs from under functorch.* to torch.func.* and made some adjustments to them.

This PR moves the relevant functorch tutorials from the functorch docs (https://pytorch.org/functorch/stable/) to pytorch/tutorials. We moved four tutorials:
- Jacobians, Hessians, hvp, vhp, and more
- Model ensembling
- per-sample-gradients
- Neural Tangent Kernels.

We also rewrite the tutorials to use the torch.func.* APIs instead of the functorch APIs, and excised mentions of functorch where appropriate.

Test Plan:
- view preview (is that possible for tutorials?)


svekars:
rebased to 2.0-RC-TEST branch

malfet (approved):

PR #2180, opened 2023-01-24
Fix YouTube beginner tutorial menu

kiersten-stokes:
While going through the beginner YT tutorial, I noticed that the "Training with PyTorch" and "Model Understanding with Captum" topics did not appear on the navigation menu. This PR changes the names of these two files to append `_tutorial` to the end of the `captumyt.py` and `trainingyt.py` files, as well as changes their reference links in the relevant places.

If there is a historical reason that this was not done earlier, please feel free to close

### Before:
Screenshot 2023-01-24 at 12 43 10 PM

### After:
Screenshot 2023-01-24 at 12 43 04 PM

cc @svekars @carljparker

### DIFF
--- a/.jenkins/validate_tutorials_built.py
+++ b/.jenkins/validate_tutorials_built.py
@@ -13,6 +13,8 @@ NOT_RUN = [
     "translation_transformer",
     "profiler",
     "saving_loading_models",
+    "introyt/captumyt_tutorial",
+    "introyt/trainingyt_tutorial",
     "examples_nn/polynomial_module",
     "examples_nn/dynamic_net",
     "examples_nn/polynomial_optim",
--- a/.jenkins/validate_tutorials_built.py
+++ b/.jenkins/validate_tutorials_built.py
@@ -13,8 +13,6 @@ NOT_RUN = [
     "translation_transformer",
     "profiler",
     "saving_loading_models",
-    "introyt/captumyt",
-    "introyt/trainingyt",
     "examples_nn/polynomial_module",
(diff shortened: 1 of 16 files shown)


kiersten-stokes:
Working on CLA 🙂

svekars:
Thank you for contributing to PyTorch. We can enable the tutorial (in the .jenkins/validate_tutorials_built.py) but we don't want to rename them.

kiersten-stokes:
> we don't want to rename them

Fair enough! @svekars Would you be willing to share why? Just wanting to get some insight into what kinds of changes will or won't be accepted 🙂 

> We can enable the tutorial (in the .jenkins/validate_tutorials_built.py)

I believe when I enabled them in a previous commit that I had a handful of related test failures. I'll revert the name changes and attempt enabling the 2 tests in a new commit. If I'm still getting failures, I'll just go ahead and close the PR

kiersten-stokes:
Looks like validation still fails for these 2 tests. Since renaming is not an option and I haven't figured out another way around this, I'll go ahead and close this PR

svekars:
> > we don't want to rename them
> 
> Fair enough! @svekars Would you be willing to share why? Just wanting to get some insight into what kinds of changes will or won't be accepted 🙂 .
> 
Sure, it's for practical reasons—to avoid creating 404 pages or redirects when it can be avoided since we don't have an automatic way to create them. I hope this helps =)

kiersten-stokes:
Makes total sense! I figured it would be something like that. Thanks for taking the time to answer!

PR #2191, opened 2023-02-02
Use the correct batch values in the output

spzala:
The current logic uses enumerate counter (i.e. variable `batch`). It displays the loss for trained data in an increment of 100 using `if batch % 100 == 0`. i.e. 1st batch, 101 stbatch, and so on for the given dataset. This maps to batch =0, 100, and so on). So for the first batch, the loss displayed should be `[ 64/60000]` instead of `[ 0/60000]`. For the second it should be `[ 6464/60000]` instead of `[ 6400/60000]` , and so on. The values displayed in the `Out:` text box on the tutorial page, e.g. `loss: 2.306185  [ 0/60000]`, read as a loss observed for zero input data which seems incorrect. This loss mentioned here was for the first batch, which was 64 input data. The second was for 101 batch which has 6464 input data, and so on.

cc @suraj813

### DIFF
--- a/beginner_source/basics/optimization_tutorial.py
+++ b/beginner_source/basics/optimization_tutorial.py
@@ -160,7 +160,7 @@ def train_loop(dataloader, model, loss_fn, optimizer):
         optimizer.step()
 
         if batch % 100 == 0:
-            loss, current = loss.item(), batch * len(X)
+            loss, current = loss.item(), (batch + 1) * len(X)
             print(f"loss: {loss:>7f}  [{current:>5d}/{size:>5d}]")
 
 
--- a/beginner_source/basics/quickstart_tutorial.py
+++ b/beginner_source/basics/quickstart_tutorial.py
@@ -151,7 +151,7 @@ def train(dataloader, model, loss_fn, optimizer):
         optimizer.step()
 
         if batch % 100 == 0:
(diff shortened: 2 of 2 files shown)


spzala:
@svekars thanks! The other thing I want to mention is that I couldn’t find a way to update the  values in the `Out:` box on the tutorial page, I hope you can help there :) If the PR pass the review, here are the values that should be used in all the epoch. We can leave the `loss` value as such considering this is an example and the values will be different in different user environment. 
```
[   64/60000]
[ 6464/60000]
[12864/60000]
[19264/60000]
[32064/60000]
[25664/60000]
[38464/60000]
[44864/60000]
[51264/60000]
[57664/60000]

spzala:
Fixed merge conflict.

spzala:
cc @HamidShojanazeri

usteiner9:
thanks

spzala:
@svekars hello, considering the changes, I am wondering if I need to request review from any specific reviewer? No rush, I just thought good idea I ask in case it's needed. Thanks!

suraj813 (approved):
@spzala thank you for improving the tutorial!

spzala:
@suraj813 yrw, and thank you so much for the review!

PR #2192, opened 2023-02-06
Fix kernel dimensions for LeNet model code example

### LINKED ISSUE #2104: Incorrect dimension in Figure: LeNet-5
In Figure LeNet-5 on this [page](https://pytorch.org/tutorials/beginner/introyt/introyt1_tutorial.html), there seems to be typo in the dimensionality of the outputs of each layer.

With an input of 32 x 32 x 1 and 6 filter of size 3 x 3, the resulting dimension after the first convolution should be 6@30 x 30. After the first subsampling(maxpool of size 2), the dimension should become 6@15 x 15. 

The second convolution with 16 filters of size 3x3 should result in 16@ 13x13, after the subsampling, the dimension should be 16@6x6

cc @suraj813

kiersten-stokes:
Fixes #2104 

In going through the [YouTube Intro tutorial](https://pytorch.org/tutorials/beginner/introyt/introyt1_tutorial.html?highlight=lenet), I noticed that the dimensions in the LeNet diagram don't match up with code that immediately follows it. Namely, the kernel size should be 5x5 rather than the 3x3 that's given. (And I believe the same should be true for the second convolutional layer?) While not really necessary for the point the tutorial is making, it may reduce confusion for those with no experience with this model (myself included 😄). Especially when comparing with the same network in [this blitz tutorial](https://pytorch.org/tutorials/beginner/blitz/neural_networks_tutorial.html?highlight=lenet) that has the correct kernel size.

cc @suraj813

### DIFF
--- a/beginner_source/introyt/introyt1_tutorial.py
+++ b/beginner_source/introyt/introyt1_tutorial.py
@@ -179,7 +179,7 @@ class LeNet(nn.Module):
         # 1 input image channel (black & white), 6 output channels, 5x5 square convolution
         # kernel
         self.conv1 = nn.Conv2d(1, 6, 5)
-        self.conv2 = nn.Conv2d(6, 16, 3)
+        self.conv2 = nn.Conv2d(6, 16, 5)
         # an affine operation: y = Wx + b
         self.fc1 = nn.Linear(16 * 6 * 6, 120)  # 6*6 from image dimension
         self.fc2 = nn.Linear(120, 84)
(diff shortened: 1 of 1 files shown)


kiersten-stokes:
I'm just realizing that there's already an open issue for this: #2104. Apologies if I didn't give enough time for the issue's author to open a PR for it

svekars:
@suraj813 can you please take a look? Thanks!

suraj813 (approved):
@kiersten-stokes thanks for fixing this!

PR #2194, opened 2023-02-06
Changing syntax for nested tensor

alex-rakowski:
Currently, this notebook doesn't run in google Colab by default environment `torch.__version__` -> 1.13.1+cu116

replacing `torch.nested_tensor(...` with `torch.nested.nested_tensor(...`


svekars:
Thank you for contributing to PyTorch! Please submit this request against the main branch rather than master.

PR #2204, opened 2023-02-10
Fix assert statements to compare the expected and actual values

bjhargrave:
The original assert statements incorrectly used the second value as the assertion failure message. This means the assert statements always succeeded as long the the object id was not zero.

cc @suraj813


suraj813 (approved):
@bjhargrave nice catch, thanks for fixing!

svekars:
@bjhargrave please sign the CLA so we can accept your change. Thanks!

bjhargrave:
> please sign the CLA so we can accept your change. Thanks!

Trying to get the right person in IBM to add me to the CCLA...

bjhargrave:
> Thank you for signing our Contributor License Agreement.

@svekars CLA is finally signed.

PR #2210, opened 2023-02-20
Use a more readable and consistent Batch variable name

spzala:
There is no bug here, but this PR provides a small improvement. In `N, C, H, W` variables, C=Channel, H=Height, and W=Width, which is fine, but N=Batch (or batch size or images in a batch) is not intuitive specially for new users and contributors. The use of `B` can be straightforwardly read as a Batch (or batch size) in the context of tutorial.

I also noticed that in another doc
https://pytorch.org/vision/stable/transforms.html#torchvision.transforms.ToTensor these variable are referenced as `B, C, H, W`, which is easy to understand. So, these PR changes also make the usage of the variables consistent.


spzala:
cc @svekars @suraj813

spzala:
@svekars @suraj813 - hello, I was out of office and traveling but back now. I wanted to follow-up with a review request :-). I will be happy to rebase after your feedback. Thanks!

suraj813:
Hi @spzala, NCHW is a standard acronym to describe a minibatch of 2D images (N = Number of samples).

spzala:
> Hi @spzala, NCHW is a standard acronym to describe a minibatch of 2D images (N = Number of samples), so I don't think this merits a change.

Thanks @subramen I agree, that's how I read it (e.g. numbers of samples in a batch) but I did notice the use of BCHW here,  https://pytorch.org/vision/stable/transforms.html#torchvision.transforms.ToTensor (`A batch of Tensor Images is a tensor of (B, C, H, W) shape, where B is a number of images in the batch...`) so thought, that's better for readability. Let me close this PR. Thanks!

PR #2243, opened 2023-03-07
Change variable name from `net` to `model` in checkpointing recipe

### LINKED ISSUE #1430: Problem of "SAVING AND LOADING A GENERAL CHECKPOINT IN PYTORCH"
In  chapter5-Load the general checkpoint of "SAVING AND LOADING A GENERAL CHECKPOINT IN PYTORCH":
model = Net()
optimizer = optim.SGD(net.parameters(), lr=0.001, momentum=0.9)
checkpoint = torch.load(PATH)
model.load_state_dict(checkpoint['model_state_dict'])
optimizer.load_state_dict(checkpoint['optimizer_state_dict'])

Now, we have loaded parameters of checkpoint for model, but optimizer loaded parameters of net which is saved into checkpoint.
I think optimizer should be loaded parameters of model rather than net, am i right?

cc @suraj813

kiersten-stokes:
Super small PR that closes #1430 

This isn't high priority, but can help with clarity for those learning the concepts. It also prevents an error from being thrown unnecessarily if not all cells of the notebook have been run in the same session. [This tutorial](https://pytorch.org/tutorials/beginner/saving_loading_models.html) has very similar code snippets, and this change makes the code consistent between the two.

[Link](https://pytorch.org/tutorials/recipes/recipes/saving_and_loading_a_general_checkpoint.html) to tutorial in question.

cc @suraj813 @pytorch/team-text-core


Nayef211:
@svekars I don't think this should be assigned to torchtext as it is a general tutorial change.

Issue #2218, opened 2023-02-24
FORWARD-MODE AUTOMATIC DIFFERENTIATION tutorial fails on build 


svekars:
The FORWARD-MODE AUTOMATIC DIFFERENTIATION tutorial is failing when ran against 2.0 binaries:
```
ModuleNotFoundError: No module named 'torch.nn.utils._stateless'`
```
The issue seems to be related to `functorch`. 

You can view the log [here](https://app.circleci.com/pipelines/github/pytorch/tutorials/7389/workflows/933408d0-cffe-4253-8ae5-913242931ee7/jobs/144308).

cc: @Chillee 

cc @zou3519

PR #2244, opened 2023-03-08
Fix forward mode ad tutorial by updating stateless import

zou3519:
Updated the import from the now deleted torch.nn.utils._stateless to torch.func

Fixes https://github.com/pytorch/tutorials/issues/2218

Test Plan:
- tested locally


svekars:
Thanks, @zou3519! Will merge when 2.0 is released.

soulitzer (approved):
Thanks!

svekars (approved):

PR #2249, opened 2023-03-13
Fix #1983 by moving input and model to device

### LINKED ISSUE #1983: Runtime error @ quickstart_tutorial.py
https://github.com/pytorch/tutorials/blob/master/beginner_source/basics/quickstart_tutorial.py

I get a runtime error when using *model.eval()* on cuda

* model.eval()
x, y = test_data[0][0], test_data[0][1]
with torch.no_grad():
    pred = model(x)
    predicted, actual = classes[pred[0].argmax(0)], classes[y]
    print(f'Predicted: "{predicted}", Actual: "{actual}"')  *

RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument mat1 in method wrapper_addmm)

I could solve this by changing: 
        pred = model(x.to(device))

heloa-net:
Fixes #1983

Moved input to device on line `234` instead of `232` for readability. `y` on line `232` is an int and doesn't have a `.to` method. Not moving it to device doesn't generate the runtime error.

I've found out about this issue while following the tutorial from Quickstart and reaching "Optimization model parameters". I was iterating on the code started on Quickstart and the model had been sent to the device. [Optimization tutorial](https://github.com/pytorch/tutorials/blob/main/beginner_source/basics/optimization_tutorial.py) doesn't move the model or input to the device and this is when I got the error described in pytorch#1983.
In light of this, **should optimization_tutorial.py also be updated?**

Test plan:
* Tested in Colab using GPU, [link](https://colab.research.google.com/drive/1X7BJHd42FN9N_8DiXd7h4ScplN2-OmLo?usp=sharing)

cc @suraj813


albanD (approved):
SGTM

PR #2251, opened 2023-03-15
Temporarily disable failing tutorials

svekars:


Balandat:
@svekars Would it be possible to notify the tutorial owners (or recent committers) about this disabling? Otherwise there is really no way to know that things are broken and need fixing. Thanks!

PR #2253, opened 2023-03-15
Use extra index to download pytorch binaries

seemethere:
To allow us to pick up changes from our own index before having to publish to PyPI

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+--extra-index-url https://download.pytorch.org/whl/cu117/index.html
 # Refer to ./jenkins/build.sh for tutorial build instructions
 
 sphinx==5.0.0


malfet (approved):
Sure, though we probably should remove it after release is done

requirements.txt
```diff
@@ -1,3 +1,4 @@
+--extra-index-url https://download.pytorch.org/whl/cu117/index.html
```
svekars, line 1: Add a comment: "Use this to run/publish tutorials against the latest binaries during the RC stage. Comment out after the release. Each release verify the correct cuda version."

svekars, line 1:
```suggestion
--extra-index-url https://download.pytorch.org/whl/cu117/index.html # Use this to run/publish tutorials against the latest binaries during the RC stage. Comment out after the release. Each release verify the correct cuda version.
```

PR #2258, opened 2023-03-17
Fix colab note

svekars:
Append a note with the link to the Colab doc to the first cell in the notebook:

Screenshot 2023-03-17 at 11 45 23 AM


intermediate_source/torch_compile_tutorial.py
```diff
@@ -36,8 +36,6 @@
 #
 # Note: a modern NVIDIA GPU (Volta or Ampere) is recommended for this tutorial.
 #
-# .. include:: ../beginner_source/new-release-colab.rst
```
carljparker, line 39: This file, `torch_compile_tutorial.py`, and the one below, `changing_default_device.py`, don't follow the same pattern as those above. In `torch_compile_tutorial.py`, we're removing `new-release-colab.rst`, but not adding the note. And in `changing_default_device.py`, we add the note, but I don't see us removing `new-release-colab.rst`. Perhaps this is intentional, but thought I should call it out.

svekars: The note will be appended to the first cell in the Jupyter notebook and will be visible only in the notebook, see changes in conf.py

carljparker: Discussed on a different channel with Svetlana; we are all good hear.

carljparker (approved):
LGTM

malfet (approved):
Sure, but please make sure to revert it when update happens

svekars:
> Sure, but please make sure to revert it when update happens

@malfet This actually just adds a link to a doc with general tips and tricks about what can be done in colab, like how to mount google drive, how to install pytorch version. We can add how to enable GPU runtime and so on and so forth. So I don't think we need to remove it.

PR #2259, opened 2023-03-19
cpp_frontend : Fix invalid beta1 call.

p-groarke:
Match tutorial source
https://github.com/pytorch/examples/blob/54f4572509891883a947411fd7239237dd2a39c3/cpp/dcgan/dcgan.cpp#L124 https://github.com/pytorch/examples/blob/54f4572509891883a947411fd7239237dd2a39c3/cpp/dcgan/dcgan.cpp#L126

### DIFF
--- a/advanced_source/cpp_frontend.rst
+++ b/advanced_source/cpp_frontend.rst
@@ -946,9 +946,9 @@ we use implement the `Adam <https://arxiv.org/pdf/1412.6980.pdf>`_ algorithm:
 .. code-block:: cpp
 
   torch::optim::Adam generator_optimizer(
-      generator->parameters(), torch::optim::AdamOptions(2e-4).beta1(0.5));
+      generator->parameters(), torch::optim::AdamOptions(2e-4).betas(std::make_tuple(0.5, 0.5)));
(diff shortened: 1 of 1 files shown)


advanced_source/cpp_frontend.rst
```diff
@@ -946,9 +946,9 @@ we use implement the `Adam <https://arxiv.org/pdf/1412.6980.pdf>`_ algorithm:
 .. code-block:: cpp
 
   torch::optim::Adam generator_optimizer(
-      generator->parameters(), torch::optim::AdamOptions(2e-4).beta1(0.5));
+      generator->parameters(), torch::optim::AdamOptions(2e-4).betas(std::make_tuple(0.5, 0.5)));
   torch::optim::Adam discriminator_optimizer(
-      discriminator->parameters(), torch::optim::AdamOptions(5e-4).beta1(0.5));
+      discriminator->parameters(), torch::optim::AdamOptions(5e-4).betas(std::make_tuple(0.5, 0.5)));
```
p-groarke, line 951: Note, the source uses 2e-4 for the `discriminator_optimizer`, I'm unsure which is right. I'll try running some learning with the source value.

PR #2264, opened 2023-03-21
Add a Python tutorial template

svekars:
Add a tutorial template for interactive tutorials stored in Python files.


beginner_source/template_tutorial.py
```diff
@@ -0,0 +1,73 @@
...
+If you have a video, add it here like this:
+
+.. raw:: html
+
+   <div style="margin-top:10px; margin-bottom:10px;">
+     <iframe width="560" height="315" src="https://www.youtube.com/embed/IC0_FRiX-sw" frameborder="0" allow="accelerometer; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+   </div>
+
+To test your tutorial locally, make a copy of this repo and add only your
+tutorial removing everything. Then run ``make html``. Verify that all outputs were generated correctly
```
carljparker, line 32: Consider adding "else": "removing everything" -> "removing everything else"

beginner_source/template_tutorial.py
```diff
@@ -0,0 +1,73 @@
...
+# Describe Why is this topic important? Add Links to relevant research papers.
+#
+# This tutorial walks you through the process of....
+#
+# Steps
+# -----
+#
+# Example code (the output below is generated automatically):
+# 
+import torch
```
carljparker, line 50: Will we know which libraries they import just by inspecting their code? I think that is probably the most efficient way. If we ask them to list their imports in a second location--such as above--then we have the possibility that list will get out of sync with what they actually end up importing.

beginner_source/template_tutorial.py
```diff
@@ -0,0 +1,73 @@
...
+
+.. raw:: html
+
+   <div style="margin-top:10px; margin-bottom:10px;">
+     <iframe width="560" height="315" src="https://www.youtube.com/embed/IC0_FRiX-sw" frameborder="0" allow="accelerometer; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+   </div>
+
+To test your tutorial locally, make a copy of this repo and add only your
+tutorial removing everything. Then run ``make html``. Verify that all outputs were generated correctly
+in the created HTML.
```
carljparker, line 33: Do we want to start requiring that people run **pylint** or **flake8** or **black**? Let's discuss--but not a reason to wait on merging.

carljparker (changes requested):
Suggest one minor text edit, otherwise LGTM.

carljparker (approved):
I'm so happy.

PR #2265, opened 2023-03-23
Update mnist_train_nas script for PyTorch Lightning 2.0 API

Balandat:
PyTorch lightning 2.0 has a different `Trainer` API which breaks the current training script used in the `ax_multiobjective_nas_tutorial`. Looks like the `pytorch-lightning` dep is not pinned, and so the setup should use the 2.0 version. 

This PR updates the callsite to avoid this breakage.

cc @dme65


Balandat:
Hmm looks like I missed the move to `main` - will update.

svekars:
Thanks @Balandat, can you please remove the ax tutorial from this list as well so it's building: https://github.com/pytorch/tutorials/blob/main/.jenkins/validate_tutorials_built.py#L52

svekars (approved):

PR #2268, opened 2023-03-24
Store build artifacts

svekars:


malfet (changes requested):
You only need to add it to the `pytorch_tutorial_build_manager_defaults`, don't you? (Last node in the fanout)

PR #2274, opened 2023-03-29
Add `pyspelling` config and workflow

malfet:
Pipeline is:
 - Filter python comments
 - Remove encoding comment and one `:: figure:` rst 
 - Convert markup to html
 - Remove code sections from converted rst (important: for markup to properly recognize something as code-section it must have correct indentation)

Add non-standard words to `tutorials-worldlist.txt`
So far only covers `data_loading_tutorial.py`


.pyspelling.yml
```diff
@@ -0,0 +1,25 @@
...
+      context_visible_first: true
+      delimiters:
+        # Exclude figure rST tags
+        - open: '\.\.\s+(figure|literalinclude|)::'
+          close: '\n'
+        # Exclude Python coding directives
+        - open: '-\*- coding:'
+          close: '\n'
+  - pyspelling.filters.markdown:
+  - pyspelling.filters.html:
```
svekars, line 22: Do we still need this filter?

malfet: Yes, as it takes care of ignoring urls, codeblocks etc defined in markup

svekars (approved):
Thank you!

PR #2300, opened 2023-04-26
add a sample code and correct some error statements

washing1127:
Add a simple sample code for torch.enable_grad() in autogradyt_tutorial.py

cc @suraj813


washing1127:
In `modelsyt_tutorial.py`, at the part of `Activation Functions`, there are 'would reduce' and 'could be reduced' repeatedly used to cause incorrect statements, I removed the 'would reduce'.

PR #2303, opened 2023-04-28
Add note and check for GPU capability

williamwen42:
Make it more clear in the torch.compile tutorial that speedup numbers may be lower than advertised if the GPU is not V100 or A100.

Follow-up to https://github.com/pytorch/pytorch/issues/99939

### DIFF
--- a/intermediate_source/torch_compile_tutorial.py
+++ b/intermediate_source/torch_compile_tutorial.py
@@ -35,7 +35,7 @@ torch.compile Tutorial
 # - ``tabulate``
 
 ######################################################################
-# NOTE: a modern NVIDIA GPU (A100 or V100) is recommended for this tutorial in
+# NOTE: a modern NVIDIA GPU (H100, A100, or V100) is recommended for this tutorial in
 # order to reproduce the speedup numbers shown below and documented elsewhere.
 
 import torch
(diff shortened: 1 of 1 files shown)


intermediate_source/torch_compile_tutorial.py
```diff
@@ -33,9 +33,25 @@
 # - ``numpy``
 # - ``scipy``
 # - ``tabulate``
-#
-# Note: a modern NVIDIA GPU (Volta or Ampere) is recommended for this tutorial.
-#
+
+######################################################################
+# NOTE: a modern NVIDIA GPU (A100 or V100) is recommended for this tutorial in
```
malfet, line 38: H100?

intermediate_source/torch_compile_tutorial.py
```diff
@@ -33,9 +33,25 @@
...
+# order to reproduce the speedup numbers shown below and documented elsewhere.
+
+import torch
+import warnings
+
+gpu_ok = False
+if torch.cuda.is_available():
+    device = torch.cuda.get_device_name()
+    if "V100" in device or "A100" in device:
```
malfet, line 47: Better use device capabilities rather than vendor name
```suggestion
    device_cap = torch.cuda.get_device_capability()
    if device_cap in [(7, 0), (8,0)]:
```

malfet (approved):
LGTM, but how about `torch.cuda.get_

PR #2320, opened 2023-05-15
Add the assign workflow

svekars:
Adding a workflow that allows users with no permissions assign issues to themselves by commenting on an issue: "/assigntome".

### DIFF
--- a/.github/workflows/docathon-assign.yml
+++ b/.github/workflows/docathon-assign.yml
@@ -31,10 +31,35 @@ jobs:
               const octokit = new Octokit({
                 auth: process.env.GITHUB_TOKEN,
               });
-              octokit.issues.addAssignees({
+              const { data: issue } = await octokit.issues.get({
                 owner: context.repo.owner,
                 repo: context.repo.repo,
-                issue_number: issueNumber,
-                assignees: [assignee]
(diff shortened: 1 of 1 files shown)


malfet (approved):

.github/workflows/docathon-assign.yml
```diff
@@ -0,0 +1,40 @@
...
+            const issueComment = context.payload.comment.body;
+            const assignRegex = /\/assigntome/i;
+            if (assignRegex.test(issueComment)) {
+              const assignee = context.payload.comment.user.login;
+              const issueNumber = context.payload.issue.number;
+              const { Octokit } = require("@octokit/rest");
+              const octokit = new Octokit({
+                auth: process.env.GITHUB_TOKEN,
+              });
+              octokit.issues.addAssignees({
```
malfet, line 34: Should we check that issue has not been assigned to others. Also, perhaps it'll be great to only allow that on issue with specific labels

svekars: Added a check that verifies that the issue is unassigned and if it assigned adds a comment. Also added a check to verify that the label is "docathon-h1-2023" and if not post a comment to pick another issue.

PR #2321, opened 2023-05-16
Update process_group_cpp_extension tutorial for dispatching to backends

H-Huang:
Update tutorial to use `Backend` instead of `ProcessGroup`. The extension of ProcessGroup is not recommended anymore and users should use `Backend` as a more flexible alternative.

Fixes a task in https://github.com/pytorch/pytorch/issues/90544

Update tutorial to point to a correct custom C++ extension example: https://github.com/H-Huang/torch_collective_extension


kwen2501 (approved):
LGTM

PR #2357, opened 2023-05-30
Propagate labels from the issue to the PR

svekars:
Add a GH action that propagates labels from the issue to PR. The issue must be tagged with the docathon label.

### DIFF
--- a/.github/scripts/docathon-label-sync.py
+++ b/.github/scripts/docathon-label-sync.py
@@ -1,0 +1,37 @@
+import os
+from github import Github
+import sys
+import re
+
+token = os.environ.get('GITHUB_TOKEN')
+
+repo_owner = "pytorch"
+repo_name = "tutorials"
+pull_request_number = int(sys.argv[1])
+
+g = Github(token)
+repo = g.get_repo(f'{repo_owner}/{repo_name}')
+pull_request = repo.get_pull(pull_request_number)
+pull_request_body = pull_request.body
+
+# get issue number from the PR body
+if re.search(r'#\d{1,5}', pull_request_body):
+    issue_number = int(re.findall(r'#(\d{1,5})', pull_request_body)[0])
+    issue = repo.get_issue(issue_number)
+    issue_labels = issue.labels
+    docathon_label_present = any(label.name == 'docathon-h1-2023' for label in issue_labels)
+
+    # if the issue has a docathon label, add all labels from the issue to the PR.
+    if docathon_label_present:
+        pull_request_labels = pull_request.get_labels()
+        issue_label_names = [label.name for label in issue_labels]
+        labels_to_add = [label for label in issue_label_names if label not in pull_request_labels]
+        if labels_to_add:
+            pull_request.set_labels(*labels_to_add)
+            print("Labels added to the pull request!")
+        else:
+            print("The pull request already has the same labels.")
+    else:
+        print("The 'docathon-h1-2023' label is not present in the issue.")
+else:
+    print("The pull request does not mention an issue.")
--- a/.github/scripts/docathon-label-sync.py
+++ b/.github/scripts/docathon-label-sync.py
@@ -3,35 +3,41 @@ from github import Github
 import sys
 import re
 
-token = os.environ.get('GITHUB_TOKEN')
+def main():
+    token = os.environ.get('GITHUB_TOKEN')
 
-repo_owner = "pytorch"
-repo_name = "tutorials"
-pull_request_number = int(sys.argv[1])
+    repo_owner = "pytorch"
+    repo_name = "tutorials"
+    pull_request_number = int(sys.argv[1])
 
-g = Github(token)
-repo = g.get_repo(f'{repo_owner}/{repo_name}')
-pull_request = repo.get_pull(pull_request_number)
-pull_request_body = pull_request.body
+    g = Github(token)
+    repo = g.get_repo(f'{repo_owner}/{repo_name}')
+    pull_request = repo.get_pull(pull_request_number)
+    pull_request_body = pull_request.body
 
-# get issue number from the PR body
-if re.search(r'#\d{1,5}', pull_request_body):
+    # get issue number from the PR body
+    if not re.search(r'#\d{1,5}', pull_request_body):
+        print("The pull request does not mention an issue.")
+        return
     issue_number = int(re.findall(r'#(\d{1,5})', pull_request_body)[0])
     issue = repo.get_issue(issue_number)
     issue_labels = issue.labels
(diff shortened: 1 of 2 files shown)


.github/workflows/docathon-label-sync.yml
```diff
@@ -0,0 +1,33 @@
...
+          fetch-depth: 0
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.x
+      - name: Install dependencies
+        run: |
+          pip install requests
+          pip install PyGithub
+      - name: Run Python script
```
kit1980, line 24: It's the same the previous step, remove?

.github/workflows/docathon-label-sync.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: |
+          pip install requests
+          pip install PyGithub
```
malfet, line 27: Duplicate?
```suggestion
```

.github/workflows/docathon-label-sync.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Set up env var
+        run: echo "PR_NUMBER=$((${{ github.event.pull_request.number }}))" >> $GITHUB_ENV
+      - name: Run Python script
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: python ./.github/scripts/docathon-label-sync.py $PR_NUMBER
```
malfet, line 33: What's wrong with this?
```suggestion
      - name: Run Python script
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: python ./.github/scripts/docathon-label-sync.py ${{ github.event.pull_request.number }}
```

.github/workflows/docathon-label-sync.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Install dependencies
+        run: |
+          pip install requests
+          pip install PyGithub
+      - name: Run Python script
+        run: |
+          pip install requests
+          pip install PyGithub
+      - name: Set up env var
+        run: echo "PR_NUMBER=$((${{ github.event.pull_request.number }}))" >> $GITHUB_ENV
```
kit1980, line 29: Instead of a separate step for "Set up env var", you can pass the PR number like this:

https://github.com/pytorch/pytorch/blob/main/.github/workflows/check-labels.yml#L52

.github/scripts/docathon-label-sync.py
```diff
@@ -0,0 +1,37 @@
...
+    issue_labels = issue.labels
+    docathon_label_present = any(label.name == 'docathon-h1-2023' for label in issue_labels)
+
+    # if the issue has a docathon label, add all labels from the issue to the PR.
+    if docathon_label_present:
+        pull_request_labels = pull_request.get_labels()
+        issue_label_names = [label.name for label in issue_labels]
+        labels_to_add = [label for label in issue_label_names if label not in pull_request_labels]
+        if labels_to_add:
+            pull_request.set_labels(*labels_to_add)
```
malfet, line 30: Name `set_labels` implies it will delete existing labels, would't it? Can you please add a ref to the doc showing that it would not remove existing labels?

.github/scripts/docathon-label-sync.py
```diff
@@ -0,0 +1,37 @@
...
+pull_request_number = int(sys.argv[1])
+
+g = Github(token)
+repo = g.get_repo(f'{repo_owner}/{repo_name}')
+pull_request = repo.get_pull(pull_request_number)
+pull_request_body = pull_request.body
+
+# get issue number from the PR body
+if re.search(r'#\d{1,5}', pull_request_body):
+    issue_number = int(re.findall(r'#(\d{1,5})', pull_request_body)[0])
```
kit1980, line 19: This is too brittle.
What if there is a text like "This is a follow-up of for my PR #12345", then `repo.get_issue` will fail.
Also I'm not sure `pull_request_body` returns "#12345" and not "https://github.com/pytorch/tutorials/issue/12345"

.github/scripts/docathon-label-sync.py
```diff
@@ -0,0 +1,37 @@
...
+        labels_to_add = [label for label in issue_label_names if label not in pull_request_labels]
+        if labels_to_add:
+            pull_request.set_labels(*labels_to_add)
+            print("Labels added to the pull request!")
+        else:
+            print("The pull request already has the same labels.")
+    else:
+        print("The 'docathon-h1-2023' label is not present in the issue.")
+else:
+    print("The pull request does not mention an issue.")
```
malfet, line 37: Please consider dismounting the pyramid of doom by exiting from `main()` routine earlier
```suggestion
def main():
    token = os.environ.get('GITHUB_TOKEN')

    repo_owner = "pytorch"
    repo_name = "tutorials"
    pull_request_number = int(sys.argv[1])

    g = Github(token)
    repo = g.get_repo(f'{repo_owner}/{repo_name}')
    pull_request = repo.get_pull(pull_request_number)
    pull_request_body = pull_request.body

    # get issue number from the PR body
    if not re.search(r'#\d{1,5}', pull_request_body):
        print("The pull request does not mention an issue.")
        return
    issue_number = int(re.findall(r'#(\d{1,5})', pull_request_body)[0])
    issue = repo.get_issue(issue_number)
    issue_labels = issue.labels
    docathon_label_present = any(label.name == 'docathon-h1-2023' for label in issue_labels)

    # if the issue has a docathon label, add all labels from the issue to the PR.
    if not docathon_label_present:
        print("The 'docathon-h1-2023' label is not present in the issue.")
        return    
    pull_request_labels = pull_request.get_labels()
    issue_label_names = [label.name for label in issue_labels]
    labels_to_add = [label for label in issue_label_names if label not in pull_request_labels]
    if not labels_to_add:
        print("The pull request already has the same labels.")
        return
    pull_request.set_labels(*labels_to_add)
    print("Labels added to the pull request!")
        

           
if __name__  == "__main__":
    main()
```

malfet (approved):

.github/workflows/docathon-label-sync.yml
svekars: resolved

.github/scripts/docathon-label-sync.py
kit1980: Probably need to use `add_to_labels` instead.

.github/scripts/docathon-label-sync.py
svekars: As discussed separately, for this purpose, this solution will suffice.

PR #2367, opened 2023-05-31
Fix run_demo(demo_model_parallel, world_size) issue

### LINKED ISSUE #1750: Fix Model Parallel demo world_size parameter in DDP Tutorial
In [this](https://pytorch.org/tutorials/intermediate/ddp_tutorial.html) tutorial, there are three demos on Distributed Training. The last one is the Model Parallel use case. In the last code block, under the `if` statement, the `world_size` parameter of the function call `run_demo(demo_model_parallel, world_size)` has to be `world_size//2` because in the Model Parallel demo, two exclusive GPU s are assigned to every process so there must be half as many processes as GPU s.

In the demo though, we see `world_size = n_gpus` in the last code block. This assignment is correct for the function calls `run_demo(demo_basic, world_size)` and `run_demo(demo_checkpoint, world_size)` but not for the `run_demo(demo_model_parallel, world_size)`

I propose to edit `if` statement in the last block to be:
```
if __name__ == "__main__":
    n_gpus = torch.cuda.device_count()
    assert n_gpus >= 2, f"Requires at least 2 GPUs to run, but got {n_gpus}"
    world_size = n_gpus
    run_demo(demo_basic, world_size)
    run_demo(demo_checkpoint, world_size)
    world_size = n_gpus//2
    run_demo(demo_model_parallel, world_size)
```

cc @mrshenli @osalpekar @H-Huang @kwen2501

TheMemoryDealer:
Fixes #1750

## Description
Fixes run_demo(demo_model_parallel, world_size) issue as described in #1750

## Checklist

cc @mrshenli @osalpekar @H-Huang @kwen2501

### DIFF
--- a/intermediate_source/ddp_tutorial.rst
+++ b/intermediate_source/ddp_tutorial.rst
@@ -269,8 +269,8 @@ either the application or the model ``forward()`` method.
         setup(rank, world_size)
 
         # setup mp_model and devices for this process
-        dev0 = (rank * 2) % (world_size * 2)
-        dev1 = (rank * 2 + 1) % (world_size * 2)
+        dev0 = rank * 2
+        dev1 = rank * 2 + 1
         mp_model = ToyMpModel(dev0, dev1)
         ddp_mp_model = DDP(mp_model)
 
--- a/intermediate_source/ddp_tutorial.rst
+++ b/intermediate_source/ddp_tutorial.rst
@@ -269,8 +269,8 @@ either the application or the model ``forward()`` method.
         setup(rank, world_size)
 
         # setup mp_model and devices for this process
-        dev0 = (rank * 2) % world_size
-        dev1 = (rank * 2 + 1) % world_size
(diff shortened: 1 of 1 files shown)


TheMemoryDealer:
@subramen  I've updated the calculation in a simple way to take into account the prior division. So now dev0 and dev1 are calculated as 
```python
    dev0 = (rank * 2) % (world_size * 2)
    dev1 = (rank * 2 + 1) % (world_size * 2)
```

TheMemoryDealer:
@subramen going back on it, does `% (world_size * 2)` not simply eliminate the `world_size = n_gpus//2` ? Would it not make more sense to just ignore `world_size = n_gpus//2` and keep 
```
dev0 = rank * 2
dev1 = rank * 2 + 1
``` 
?

subramen:
Yes, you don't actually need `world_size` anymore :) that is the update I was looking for.

PR #2375, opened 2023-06-01
Optimize DataLoader iteration in WrappedDataLoader

### LINKED ISSUE #1757: Misleading call of 'iter' in the 'WrappedDataLoader'
https://github.com/pytorch/tutorials/blob/796013c10cde2cc8afb2e6f55acc26f12ca7ff2f/beginner_source/nn_tutorial.py#L776
There is no need to call `iter` in this line and that is misleading. Internally, by calling `iter` an `_BaseDataLoaderIter` instance is returned. Then in the `for` loop, again `iter` is called by the interpreter and the [same instance](https://github.com/pytorch/pytorch/blob/3202028ed1ca24c91dc7192ef69b305690db7abc/torch/utils/data/dataloader.py#L504) is returned. So there is no need to that call.

cc @suraj813

HemanthSai7:
Fixes #1757 

## Description

The commit removes redundant iter() calls in the WrappedDataLoader class, optimizing the iteration process. Previously, iter() was called when initializing the iterator and again within the for loop, which was unnecessary. The code is now clearer and more efficient by eliminating the redundant call.

Previously
```python
def __iter__(self):
        batches = iter(self.dl)
        for b in batches:
            yield (self.func(*b))
```

Modified
```python
def __iter__(self):
        for b in self.dl:
            yield (self.func(*b))
```

## Checklist

- [x] The issue that is being fixed is referred to in the description (see above "Fixes #1757")
- [x] No unnecessary issues are included in this pull request.

cc @suraj813 @svekars

### DIFF
--- a/beginner_source/nn_tutorial.py
+++ b/beginner_source/nn_tutorial.py
@@ -790,8 +790,7 @@ class WrappedDataLoader:
         return len(self.dl)
 
     def __iter__(self):
-        batches = iter(self.dl)
-        for b in batches:
+        for b in self.dl:
             yield (self.func(*b))
 
 train_dl, valid_dl = get_data(train_ds, valid_ds, bs)


svekars (approved):
LGTM

PR #2380, opened 2023-06-01
Fix the loss initialization in intermediate_source/char_rnn_generation_tutorial.py

### LINKED ISSUE #844: Typo instantiating loss at zero
https://github.com/pytorch/tutorials/blob/48e5cccecbd385e00b95cb60ecdfecd06a43282b/intermediate_source/char_rnn_generation_tutorial.py#L281

Hello folks at Pytorch,

When instantiating loss at a real zero, i think you rather meant something like (i don't know if there is a better way but I usually do it this way) : 

`loss = torch.autograd.Variable(torch.Tensor([0]))`

akjalok:
Fixes #844

## Description
Changed the loss initialization from 0 to torch.Tensor([0]) in order to avoid confusion about the dtypes.

## Checklist

- [x]  Only one issue is addressed in this pull request
cc @svekars @carljparker @kit1980

### DIFF
--- a/intermediate_source/char_rnn_generation_tutorial.py
+++ b/intermediate_source/char_rnn_generation_tutorial.py
@@ -278,7 +278,7 @@ def train(category_tensor, input_line_tensor, target_line_tensor):
 
     rnn.zero_grad()
 
-    loss = 0
+    loss = torch.Tensor([0])
 
     for i in range(input_line_tensor.size(0)):
         output, hidden = rnn(category_tensor, input_line_tensor[i], hidden)
--- a/intermediate_source/char_rnn_generation_tutorial.py
+++ b/intermediate_source/char_rnn_generation_tutorial.py
@@ -278,7 +278,7 @@ def train(category_tensor, input_line_tensor, target_line_tensor):
 
     rnn.zero_grad()
 
-    loss = torch.Tensor([0])
+    loss = torch.tensor([0])
 
     for i in range(input_line_tensor.size(0)):
         output, hidden = rnn(category_tensor, input_line_tensor[i], hidden)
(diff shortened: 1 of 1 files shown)


intermediate_source/char_rnn_generation_tutorial.py
```diff
@@ -278,7 +278,7 @@ def train(category_tensor, input_line_tensor, target_line_tensor):
 
     rnn.zero_grad()
 
-    loss = 0
+    loss = torch.Tensor([0])
```
NicolasHug, line 281: This works but it should be

```suggestion
    loss = torch.tensor([0])
```

NicolasHug (approved):
I'll approve, but as mentioned in the original issue https://github.com/pytorch/tutorials/issues/844#issuecomment-1571271807 the current code is just fine. We don't need to initialize `loss` as a tensor.

intermediate_source/char_rnn_generation_tutorial.py
```diff
@@ -278,7 +278,7 @@ def train(category_tensor, input_line_tensor, target_line_tensor):
 
     rnn.zero_grad()
 
-    loss = 0
+    loss = torch.tensor([0]) # in PyTorch 2.0 and later can be don't need to initialize as tensor, can be ``loss = 0``
```
NicolasHug, line 281: I believe this is true for any pytorch version, it's not related to 2.0

svekars, line 281:
```suggestion
    loss = torch.tensor([0]) # you can also just simply use ``loss = 0``
```

intermediate_source/char_rnn_generation_tutorial.py
```diff
@@ -278,7 +278,7 @@ def train(category_tensor, input_line_tensor, target_line_tensor):
 
     rnn.zero_grad()
 
-    loss = 0
+    loss = torch.tensor([0]) # you can also just simply use ``loss = 0``
```
NicolasHug, line 281:
```suggestion
    loss = torch.Tensor([0]) # you can also just simply use ``loss = 0``
```

PR #2381, opened 2023-06-01
Update mario_rl_tutorial.py

### LINKED ISSUE #1620: RuntimeError: CUDA error: out of memory while trying mario_rl_tutorial.py
When I tried the mario_rl_tutorial.py using CUDA, I received an out of memory error:
```
RuntimeError: CUDA error: out of memory
```
Monitoring the GPU memory I saw a constant incremental of the use of its memory, specially when the Agent starts the training. As a workaround I changed the cache method of the Agent, as it seems that all the experiencies are sent to the GPU:
```python
        if self.use_cuda:
            state = torch.tensor(state).cuda()
            next_state = torch.tensor(next_state).cuda()
            action = torch.tensor([action]).cuda()
            reward = torch.tensor([reward]).cuda()
            done = torch.tensor([done]).cuda()
 (...)

        self.memory.append((state, next_state, action, reward, done,))
```
Instead of that, I modified this piece for using CPU and changed the recall method, to send at this moment the batch of experiencies to GPU:
```python

    def recall(self):
        """
        Retrieve a batch of experiences from memory
        """
        batch = random.sample(self.memory, self.batch_size)
        state, next_state, action, reward, done = map(torch.stack, zip(*batch))
        return state.cuda(), next_state.cuda(), action.squeeze().cuda(), reward.squeeze().cuda(), done.squeeze().cuda()
```

Not sure if this is the right approach, but it works. However, it might impact on the learning performance.

Could anyone help here?

cc @vmoens @nairbv

neuralninja27:
Fixes #1620 

## Description
Modified code to remove the CUDA out of memory error

## Checklist

- [x] The issue that is being fixed is referred in the description (see above "Fixes #1620 ")

cc @vmoens @nairbv


intermediate_source/mario_rl_tutorial.py
```diff
@@ -348,7 +348,7 @@ def act(self, state):
 class Mario(Mario):  # subclassing for continuity
     def __init__(self, state_dim, action_dim, save_dir):
         super().__init__(state_dim, action_dim, save_dir)
-        self.memory = deque(maxlen=100000)
+        self.memory = deque(maxlen=60000) # Please change this value according to your GPU memory (current value is suitable for a GPU with 16GB RAM)
```
vmoens, line 351: @svekars 
TorchRL replay buffer won't sit on CUDA and be more efficient than a deque.
Not sure if we want to integrate it here?

neuralninja27:
@svekars sure I'll look into it

intermediate_source/mario_rl_tutorial.py
```diff
@@ -1,3 +1,4 @@
+, storage
```
vmoens, line 1:
```suggestion
```

vmoens (approved):
LGTM thanks for this!

PR #2383, opened 2023-06-01
Change batchify desc to remove ambiguity

### LINKED ISSUE #1037: Ambiguity in transformer_tutorial.py
The text in the tutorial indicates that the batch size is the outermost dimension of the data: "For instance, with the alphabet as the sequence (total length of 26) and a batch size of 4, we would divide the alphabet into 4 sequences of length 6:" But in the code there is a call to `.t()` so that the sequence length is the outermost dimension.

QasimKhan5x:
Fixes #1037 

## Description
Changed the text,

> we would divide the alphabet into 4 sequences of length 6

to 

> we would divide the alphabet into sequences of length 6, resulting in 4 such sequences

With this, the reader knows that a tensor of shape (seq_length, batch_size) is being created, which is consistent with the shape of the tensor returned by the `batchify` function


carljparker (approved):
This looks good, Qasim. Thanks for your help!

PR #2384, opened 2023-06-01
Fix cropping to include last column and last row

### LINKED ISSUE #876: Missing last row and column of image in result of RandomCrop in data_loading_tutorial
Result of cropping never contains last column and last row of image pixels.

Method np.random.randint(low, high) returns value from low (inclusive) and to high(exclusive) (See [documentation](https://numpy.org/doc/1.18/reference/random/generated/numpy.random.randint.html?highlight=randint#numpy.random.randint))
Therefore variable `top` takes value from `0` to `h - new_h - 1` inclusive. 
Maximal bottom equals `(h - new_h - 1) + new_h - 1 = h - 2` but not `h - 1`. 
Row indexed `h - 1` never gets into result.
Similarly, column indexed `w - 1`.

arunppsg:
Fixes #876

## Description

I have updated image cropping lines to include last row and column. Earlier the top index was from `[0, h - new_h)` where `h` is the height of the image and `new_h` is the height of the cropped image. 

Consider a case where `new_h = 3` and `h = 4`. In this case, top index can only be 0 and hence the last row will never get started. To fix this, I have updated `np.random.randint` to sample from `[0, h - new_h + 1)` for choosing last row and for chosing last column, I have update the func to sample from `[0, w - new_w + 1)`
 
## Checklist

### DIFF
--- a/beginner_source/data_loading_tutorial.py
+++ b/beginner_source/data_loading_tutorial.py
@@ -268,8 +268,8 @@ class RandomCrop(object):
         h, w = image.shape[:2]
         new_h, new_w = self.output_size
 
-        top = np.random.randint(0, h - new_h)
-        left = np.random.randint(0, w - new_w)
+        top = np.random.randint(0, h - new_h + 1)
+        left = np.random.randint(0, w - new_w + 1)
 
         image = image[top: top + new_h,
                       left: left + new_w]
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
LGTM, thanks @arunppsg

PR #2401, opened 2023-06-01
Render a non-empty plot in Mario-RL tutorial

### LINKED ISSUE #1607: Blank image in Mario-RL tutorial
There is a large whitespace due to a blank image in the mario-rl tutorial

image

This image should be removed or replaced with the intended image.

cc @vmoens @nairbv @svekars @carljparker

kiersten-stokes:
Fixes #1607

## Description
This PR makes a few changes in order to render a non-empty plot in the doc HTML.
- Ups the number of episodes run so that there are enough data points to plot (w/ 10 episodes, only one data point would be available due to a save every 20)
- Clears the plot at the beginning of the loop that plots all 4 metrics; this ensure the plot is not cleared on the final iteration and results in something being displayed
- Adds a legend to the plot so that the rendered image contains some context and isn't just a random straight line

New rendering:
Screenshot 2023-06-01 at 1 52 23 PM

## Checklist

cc @vmoens @nairbv @svekars @carljparker

### DIFF
--- a/intermediate_source/mario_rl_tutorial.py
+++ b/intermediate_source/mario_rl_tutorial.py
@@ -711,17 +711,18 @@ class MetricLogger:
                 f"{datetime.datetime.now().strftime('%Y-%m-%dT%H:%M:%S'):>20}\n"
             )
 
-        for metric in ["ep_rewards", "ep_lengths", "ep_avg_losses", "ep_avg_qs"]:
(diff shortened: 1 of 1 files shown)


svekars:
Will need to wait for the manager worker to finish and check in the build artifacts. Netlify only runs the simple make html-noplot without actually running the code and generating outputs. It's a quick preview for .rst/HTML only fixes.

PR #2402, opened 2023-06-01
Update transformer_tutorial.py | Resolving issue #1778

### LINKED ISSUE #1778: [Help Wanted] Why take the log function and then apply exp?
In [line of code](https://github.com/pytorch/tutorials/blob/master/beginner_source/transformer_tutorial.py#L113), you calculate positional encoding for Transformers by taking the log first and then apply the exponential function.

Would you please elaborate on why you do this instead of directly doing the calculation?

I'm aware that log transformation can make multiplication become addition, but it seems that this is not the case here.

cc @suraj813

YoushaaMurhij:
Add description for positional encoding calculation for Transformers

Fixes #1778

## Description
Add explanation for calculating positional encoding for Transformers by taking the log first and then applying the exponential function.

## Checklist

- [x] The issue that is being fixed is referred in the description (see above "Fixes #1778")

cc @suraj813

### DIFF
--- a/beginner_source/transformer_tutorial.py
+++ b/beginner_source/transformer_tutorial.py
@@ -103,22 +103,22 @@ def generate_square_subsequent_mask(sz: int) -> Tensor:
 # positional encodings have the same dimension as the embeddings so that
 # the two can be summed. Here, we use ``sine`` and ``cosine`` functions of
 # different frequencies.
-# The div_term in the code is calculated as 
-# torch.exp(torch.arange(0, d_model, 2) * (-math.log(10000.0) / d_model)). 
+# The ``div_term`` in the code is calculated as 
+# ``torch.exp(torch.arange(0, d_model, 2) * (-math.log(10000.0) / d_model))``. 
 # This calculation is based on the original Transformer paper’s formulation 
 # for positional encoding. The purpose of this calculation is to create 
 # a range of values that decrease exponentially. 
 # This allows the model to learn to attend to positions based on their relative distances.
-# The math.log(10000.0) term in the exponent represents the maximum effective 
-# input length (in this case, 10000). Dividing this term by d_model scales 
+# The ``math.log(10000.0)`` term in the exponent represents the maximum effective 
+# input length (in this case, ``10000``). Dividing this term by ``d_model`` scales 
 # the values to be within a reasonable range for the exponential function. 
 # The negative sign in front of the logarithm ensures that the values decrease exponentially.
-# The reason for writing math.log(10000.0) instead of 4 in the code is to make it clear
+# The reason for writing ``math.log(10000.0)`` instead of ``4`` in the code is to make it clear
 # that this value represents the logarithm of the maximum effective input length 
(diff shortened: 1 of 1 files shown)


beginner_source/transformer_tutorial.py
```diff
@@ -103,7 +103,23 @@ def generate_square_subsequent_mask(sz: int) -> Tensor:
...
+# ``torch.exp(torch.arange(0, d_model, 2) * (-math.log(10000.0) / d_model))``. 
+# This calculation is based on the original Transformer paper’s formulation 
+# for positional encoding. The purpose of this calculation is to create 
+# a range of values that decrease exponentially. 
+# This allows the model to learn to attend to positions based on their relative distances.
+# The ``math.log(10000.0)`` term in the exponent represents the maximum effective 
+# input length (in this case, ``10000``). Dividing this term by ``d_model`` scales 
+# the values to be within a reasonable range for the exponential function. 
+# The negative sign in front of the logarithm ensures that the values decrease exponentially.
+# The reason for writing ``math.log(10000.0)`` instead of ``4`` in the code is to make it clear
```
NicolasHug, line 116: I don't understand this comment. math.log(10000.0) is 9.2, not 4.

carljparker (approved):
Wow. Nice explanation. Thanks.

beginner_source/transformer_tutorial.py
```diff
@@ -103,6 +103,15 @@ def generate_square_subsequent_mask(sz: int) -> Tensor:
 # positional encodings have the same dimension as the embeddings so that
 # the two can be summed. Here, we use ``sine`` and ``cosine`` functions of
 # different frequencies.
+# The ``div_term`` in the code is calculated as 
+# ``torch.exp(torch.arange(0, d_model, 2) * (-math.log(10000.0) / d_model))``. 
+# This calculation is based on the original Transformer paper’s formulation 
+# for positional encoding. The purpose of this calculation is to create 
+# a range of values that decrease exponentially. 
+# This allows the model to learn to attend to positions based on their relative distances.
+# The ``math.log(10000.0)`` term in the exponent represents the maximum effective
```
NicolasHug, line 112: I'm not sure this is correct, the maximum input length is `max_len`, not 10000. Am I missing something?

YoushaaMurhij: The purpose of this value is to make the frequencies of the sine and cosine functions very large. This is important because it helps to ensure that the positional encodings are unique for each position in the sequence. Right?

YoushaaMurhij: I think, I need to update this, too.

svekars:
@YoushaaMurhij please submit a PR updating the description as suggested. Tag @NicolasHug and @kit1980 to review your update.

beginner_source/transformer_tutorial.py
svekars: Also, can you please make the description less lengthy and in a simpler language. Thank you!

PR #2403, opened 2023-06-01
Implement  function for BERT quantization tutorial, resolves issue #1971

### LINKED ISSUE #1971: ids_tensor function not defined in bert quantization tutorial

JoseLuisC99:
Fixes #1971 

Fix missing implementation of `ids_tensor`.

### DIFF
--- a/intermediate_source/dynamic_quantization_bert_tutorial.rst
+++ b/intermediate_source/dynamic_quantization_bert_tutorial.rst
@@ -255,6 +255,9 @@ model before and after the dynamic quantization.
         torch.manual_seed(seed)
     set_seed(42)
 
+    # Initialize a global random number generator
+    global_rng = random.Random()
+
 
 2.2 Load the fine-tuned BERT model
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


carljparker (approved):
I continue to wonder about the `name=None` parameter, but it does exist in the version of this function at the following URL, so I guess we can keep it.

Quantizing BERT Model
https://pytorch.org/tutorials/prototype/graph_mode_dynamic_bert_tutorial.html?highlight=transformer#setup

intermediate_source/dynamic_quantization_bert_tutorial.rst
```diff
@@ -525,6 +528,21 @@ We can serialize and save the quantized model for the future use using
...
+
+        total_dims = 1
+        for dim in shape:
+            total_dims *= dim
+
+        values = []
+        for _ in range(total_dims):
+            values.append(rng.randint(0, vocab_size - 1))
+
+        return torch.tensor(data=values, dtype=torch.long, device='cpu').view(shape).contiguous()
```
malfet, line 544: This would create `int64` tensor, not `int32` one, wouldn't it?

malfet, line 544: Also, isn't it all just equivalent of `return torch.randint(0, vocab_size, shape=shape, dtype=torch.int, device='cpu')

malfet: https://github.com/pytorch/tutorials/pull/2431

PR #2407, opened 2023-06-02
Improve data_loading_tutorial.py with use of enumerate

### LINKED ISSUE #995: data_loading_tutorial.py iterators
I like this tutorial but I think it would be better if it included an example of how to define __next__() and __iter__() methods so that the dataset can be used with `enumerate`.

noqqaqq:
Fixes #995

## Description
Replace iteration over range of indexes with enumerate over dataset in data_loading_tutorial.py

## Checklist


NicolasHug (approved):
LGTM, thanks @noqqaqq

PR #2410, opened 2023-06-02
Update Positional Encoding description | Resolving issue #1778

### LINKED ISSUE #1778: [Help Wanted] Why take the log function and then apply exp?

YoushaaMurhij:
Fixes #1778 

## Description
Update Positional Encoding description

## Checklist

- [x] The issue that is being fixed is referred in the description (see above "Fixes #1778 ")

cc @suraj813


svekars:
Thanks. Unfortunately, this does not really quite work either. We recommend that you pick another issue to work on.

PR #2411, opened 2023-06-02
Set global device back to cpu at the end of tutorial

malfet:
We are using sphinx to render those tutorials, which does not start a new process to render, so one needs to restore global state to default value, by calling `torch.set_default_device('cpu')`


clee2000 (approved):

recipes_source/recipes/changing_default_device.py
```diff
@@ -43,6 +43,9 @@
 print(mod.weight.device)
 print(mod(torch.randn(128, 20)).device)
 
+# And then gloablly return it back to CPU
```
svekars, line 46:
```suggestion
# And then globally return it back to CPU
```

Issue #2177, opened 2023-01-21
Do not use deepcopy to copy nn.modules (including entire models)


mikekgfb:
copy.deepcopy is not defined for nn.module() and does not reliably copy an nn.module hierrchy, such as a model or partial model.  Our tutorials should using `copy.deepcopy()` as this will induce our users to make incorrect use of the primitive.  

Please update https://pytorch.org/tutorials/beginner/transformer_tutorial.html and other tutorials to avoid the use.  Recommended way for snapshotting is to use torch.save() for trained models. to create multiple clones of an untrained model, create multiple copies from first principles from the model architectural parameters.

cc @pytorch/team-text-core @Nayef211

mikekgfb:
cc: @mthrok

Nayef211:
Thanks for creating this issue @mikekgfb. Will submit a PR with the fixes soon.

Nayef211:
@svekars I don't think this issue should be closed. The PR I sent out fixed the issue for https://pytorch.org/tutorials/beginner/transformer_tutorial.html. But there are several [other places](https://github.com/search?q=repo%3Apytorch%2Ftutorials%20copy.deepcopy(&type=code) where it's still being used.

svekars:
Can the owners of the tutorials please take a look and submit corrections? If some of these tutorials describe an obsolete feature, please submit a PR to remove them:

* [fx_numeric_suite_tutorial](https://github.com/pytorch/tutorials/blob/6ae56a1a2639308550a69c5006e879d331f05b6d/prototype_source/fx_numeric_suite_tutorial.py#L89) - @vkuzo 
* [fx_graph_mode_ptq_static](https://github.com/pytorch/tutorials/blob/6ae56a1a2639308550a69c5006e879d331f05b6d/prototype_source/fx_graph_mode_ptq_static.rst#L218) - @jerryzh168 
* [fx_conv_bn_fuser](https://github.com/pytorch/tutorials/blob/6ae56a1a2639308550a69c5006e879d331f05b6d/intermediate_source/fx_conv_bn_fuser.py#L107) - @Chillee 
* [transfer_learning_tutorial](https://github.com/pytorch/tutorials/blob/6ae56a1a2639308550a69c5006e879d331f05b6d/beginner_source/transfer_learning_tutorial.py#L149) - @z-a-f 
[...]

PR #2417, opened 2023-06-03
Update Quantized Transfer Learning for Computer Vision tutorial to avoid use of copy.deepcopy()

### LINKED ISSUE #2333: Update tutorial to avoid use of copy.deepcopy()- Quantized Transfer Learning for Computer Vision
Update the  [Quantized Learning for Computer Vision Tutorial](https://github.com/pytorch/tutorials/blob/6ae56a1a2639308550a69c5006e879d331f05b6d/intermediate_source/quantized_transfer_learning_tutorial.rst#L193) to avoid the use of copy.deepcopy().

Tutorial link: https://github.com/pytorch/tutorials/blob/6ae56a1a2639308550a69c5006e879d331f05b6d/intermediate_source/quantized_transfer_learning_tutorial.rst#L193

See issue #2177  for more information

cc @svekars @carljparker @NicolasHug @kit1980 @subramen

mikgor:
Fixes #2333 

## Description
Replace `copy.deepcopy()` with `model.load_state_dict` in Quantized Transfer Learning for Computer Vision tutorial.


NicolasHug:
Thank you for the PR @mikgor . I'm afraid the original code `copy.deepcopy(model.state_dict())` is actually correct and doesn't need to be updated. I see https://github.com/pytorch/tutorials/issues/2333 was opened but I don't think it fits the original issue: here we're creating deepcopies of `model.state_dict()` which is a dict and that's a fine thing to do. The original problem stated in https://github.com/pytorch/tutorials/issues/2177 is about copying `nn.Module` but that's not the case here.

I don't think we should be merging this PR but let me see if we can still get you credits for it somehow

svekars:
Thank you for looking into this @mikgor. We have recorded your contribution. Closing.

PR #2419, opened 2023-06-03
 Fix inconsistencies in fgsm_tutorial

### LINKED ISSUE #1032: Adversarial example generation by FGSM: different normalization of training vs test images?
In the Adversarial example generation tutorial the classifier from https://github.com/pytorch/examples/tree/master/mnist is used. However, this classifier is trained with input normalization  `transforms.Normalize((0.1307,), (0.3081,))` while in the FGSM tutorial no normalization is used and the perturbed images are clamped to [0,1] - is this not a contradiction?

QasimKhan5x:
Fixes #1032 

## Description

1. The architecture of the model is changed. The tutorial says that it uses the same MNIST model from [pytorch/examples/mnist](https://github.com/pytorch/examples/tree/master/mnist). However, the dimensions of the convolutional layers were different. Therefore, the exact model is copy-pasted as is.
2. The model in the MNIST example was trained using a transform in the data loader that normalized the data using `mean=0.1307` and `std=0.3081`. However, no normalization was being applied in this example. Thus, the same normalization is applied to the `test_loader` in this example.
3. The `state_dict` of the MNIST was being loaded on the CPU even though `use_cuda=True` and `device` is being dynamically defined. Therefore, `map_location=device` instead of `cpu`
4. A denormalization function is added to unnormalize the data before applying FGSM on it, since FGSM takes unnormalized input
5. The `test` function is modified. After computing the gradient of the normalized data, the data is unnormalized. FGSM is then applied on unnormalized data. Finally, the perturbed data is again normalized before applying the model on it. 

I also checked the code without step 5, i.e., not performing normalization and denormalization inside the `test` function. The observation was that the accuracy stayed the same between 98-99%, although it was supposed to fall. Applying step 5 rectified this issue.

## Checklist


QasimKhan5x:
the CI is failing because the tutorial relies on `"data/lenet_mnist_model.pth"`, which contained the weights of the old model. I don't have access to this file, otherwise I would change it.

malfet:
@QasimKhan5x I can update the file, but weights [referenced](https://drive.google.com/drive/folders/1fn83DF14tWmit0RTKWRhPq5uVXt73e0h?usp=sharing) from the tutorial are indeed for small model:
```
>>> import torch
>>> for k, v in torch.load("lenet_mnist_model.pth", weights_only=True, map_location='cpu').items():
...   print(k, v.shape)
conv1.weight torch.Size([10, 1, 5, 5])
conv1.bias torch.Size([10])
conv2.weight torch.Size([20, 10, 5, 5])
conv2.bias torch.Size([20])
fc1.weight torch.Size([50, 320])
fc1.bias torch.Size([50])
fc2.weight torch.Size([10, 50])
fc2.bias torch.Size([10])
```

I.e. your PR should include new location to download the model in both tutorial itself as well as here: https://github.com/pytorch/tutorials/blob/47b9ea4628764a19e867d8a0227a81fc2138c0bc/Makefile#L85

malfet:
@QasimKhan5x any public location will do, though google drive is preferable (as tutorials are most frequently run on colab, and one can connect collab to g-drive, can't they?)

QasimKhan5x:
@malfet completed the changes

QasimKhan5x:
current build failed due to an error in intermediate_source/mario_rl_tutorial.py which is unrelated to this PR

svekars (approved):

PR #2423, opened 2023-06-05
Remove improper src_mask from encoder tutorial

### LINKED ISSUE #1877: Transformer Language Model Tutorial - Incorrect Attention Mask Description
The tutorial [Language Modeling With nn.Transformer and Torchtext](https://pytorch.org/tutorials/beginner/transformer_tutorial.html) describes an attention mask that will prevent `nn.TransformerEncoder` from attending to not-yet-seen tokens:

> Along with the input sequence, a square attention mask is required because the self-attention layers in nn.TransformerEncoder are only allowed to attend the earlier positions in the sequence.

I think the mask is actually upper triangular: a square mask would not prevent a token attending to future token. 

I may have misunderstood the mask description; if not, happy to write the PR that fixes this.

cc @pytorch/team-text-core @Nayef211

bjhargrave:
Fixes #1877 
The tutorial is using a transformer encoder and the mask used was for masking a decoder which is not part of the tutorial.

The mask is removed. Some variable names are changed to better reflect the purpose of the variable. Also, some unused imports are removed.

## Checklist

cc @pytorch/team-text-core @Nayef211

### DIFF
--- a/beginner_source/transformer_tutorial.py
+++ b/beginner_source/transformer_tutorial.py
@@ -36,12 +36,8 @@ can be easily adapted/composed.
 # of the word (see the next paragraph for more details). The
 # ``nn.TransformerEncoder`` consists of multiple layers of
 # `nn.TransformerEncoderLayer <https://pytorch.org/docs/stable/generated/torch.nn.TransformerEncoderLayer.html>`__.
-# Along with the input sequence, a square attention mask is required because the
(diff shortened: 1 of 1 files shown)


Nayef211 (approved):
LGTM. Thanks for helping improve PyTorch tutorials @bjhargrave

PR #2427, opened 2023-06-05
Sequence to sequence encoder-decoder attention

### LINKED ISSUE #2121: Sequence to sequence encoder-decoder attention not looking at encoder hidden states
Hi,

I believe that the [code/notebook example](https://github.com/pytorch/tutorials/blob/master/intermediate_source/seq2seq_translation_tutorial.py) of a sequence to sequence model that uses attention should be changed.

Attention (as described in [Bahdanau 2014](https://arxiv.org/abs/1409.0473)) serves as a mechanism for looking at encoder hidden states (now often referred to as cross-attention when talking about Transformers). Theoretically, its justification is that it allows a variable size representation of information depending on input sequence length (which is variable), so that there is not a representation bottleneck at the output of the final hidden layer for the encoder.

The tutorial/code currently computes attention by using the decoder's hidden state after the first input and the embedding of the decoder's first output. I believe that this is incorrect. 

I think it should instead be using the decoder's prior hidden state and the encoder's hidden states so that the model knows what to focus on from the input sequence for the current timestep of the decoder's output.

I can submit a PR to change this in the tutorial if correct; thanks.

cc @pytorch/team-text-core @Nayef211

JoseLuisC99:
Fixes #2121 

## Description
Redefines the attention mechanism and decoder network from the tutorial [NLP From Scratch: Translation with a Sequence to Sequence Network and Attention](https://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html).

cc @pytorch/team-text-core @Nayef211

### DIFF
--- a/prototype_source/quantization_in_pytorch_2_0_export_tutorial.rst
+++ b/prototype_source/quantization_in_pytorch_2_0_export_tutorial.rst
@@ -1,7 +1,9 @@
-(Work in Progress) Quantization in PyTorch 2.0 Export Tutorial
+Quantization in PyTorch 2.0 Export Tutorial
 ==============================================================
 
 **Author**: `Leslie Fang <https://github.com/leslie-fang-intel>`_, `Weiwen Xia <https://github.com/Xia-Weiwen>`__, `Jiong Gong <https://github.com/jgong5>`__, `Kimish Patel <https://github.com/kimishpatel>`__, `Jerry Zhang <https://github.com/jerryzh168>`__
+
+.. note:: Quantization in PyTorch 2.0 export is still a work in progress.
 
 Today we have `FX Graph Mode
 Quantization <https://pytorch.org/docs/stable/quantization.html#prototype-fx-graph-mode-quantization>`__


intermediate_source/seq2seq_translation_tutorial.py
```diff
@@ -619,42 +642,48 @@ def timeSince(since, percent):
...
-        if iter % plot_every == 0:
-            plot_loss_avg = plot_loss_total / plot_every
-            plot_losses.append(plot_loss_avg)
-            plot_loss_total = 0
-
+    tot_iter = 0
+    stop = False
+    while True:
+        for data in train_dataloader:
+            tot_iter += 1
```
Nayef211, line 660: Do we need this `while True` statement? Once the dataloader is exhausted I assume the iterator doesn't reset to the beginning?

svekars:
@JoseLuisC99 just a friendly reminder that the docathon is ending on Thursday and we'd appreciate if you could address  feedback ASAP. Thank you!

Issue #1449, opened 2021-03-27
The "traced_cell.code" output is wrong and misleading!!


git-hub-tig:
In this code:
https://github.com/pytorch/tutorials/blob/70f82fe9537c6ceb21bf7f2a633206c022fa0031/beginner_source/Intro_to_TorchScript_tutorial.py#L224

The `print(traced_cell.code)` output in docs [https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html](url) is:
```python
def forward(self,
    input: Tensor,
    h: Tensor) -> Tuple[Tensor, Tensor]:
  _0 = torch.add((self.linear).forward(input, ), h, alpha=1)
  _1 = torch.tanh(_0)
  return (_1, _1)
```
But I think the right `print(traced_cell.code)` output should be like:
```python
def forward(self,  
    input: Tensor,  
    h: Tensor) -> Tuple[Tensor, Tensor]:  
  _0 = self.dg  
  _1 = (self.linear).forward(input, )  
  _2 = (_0).forward(_1, )  
  _3 = torch.tanh(torch.add(_1, h, alpha=1))  
  return (_3, _3)  
```
The problem of generation of misleading output maybe related to the same variable name in both trace and script codes:
```python
traced_cell = torch.jit.trace(my_cell, (x, h))  
traced_cell = torch.jit.trace(my_cell, (x, h))   
```
What I know is that:
> torch.jit.script has a non-negligible start up time when it is first called due to lazy-initializations of many compiler builtins.  

If any guy know the real answer, please help this!!

arunppsg:
If I am not wrong, the result is random because of the `if-else` condition in the `forward` method. Quoting from the tutorial, 

```
Tracing does exactly what we said it would: run the code, record the operations that happen and construct a ScriptModule that does exactly that. Unfortunately, things like control flow are erased.
```

When the control flows through the `else` part, it prints one trace and the when the control flows through the `if` part, it prints another trace. Since `x` is random, we can't know beforehand how the control will flow in that particular trace.

PR #2428, opened 2023-06-05
Set the random seed for reproducibility of the output

bjhargrave:
We also fix the code to use the scripted_cell just created.

Fixes https://github.com/pytorch/tutorials/issues/1449

## Checklist


NicolasHug (approved):
Thanks @bjhargrave

PR #2431, opened 2023-06-06
[BE] Simplify `ids_tensor`

malfet:
Remove `global_rng` and use `torch.randint` to feel the tensor of shape `shape` with values in range `[0, vocab_size)`


svekars (approved):
Thanks, Nikita!

PR #2433, opened 2023-06-06
Optimizer state is not synchronized across replicas like model state is

timgianitsos:
From `DistributedDataParallel` docs:
"The module... assumes that [gradients] will be modified by the optimizer in all processes in the same way." Note that this is "assumed", not enforced. From https://pytorch.org/tutorials/recipes/zero_redundancy_optimizer.html : "each process keeps a dedicated replica of the optimizer. Since DDP has already synchronized gradients in the backward pass, all optimizer replicas will operate on the same parameter and gradient values in every iteration, and this is how DDP keeps model replicas in the same state"

## Checklist


svekars:
Can you please reference an issue you are fixing in the PR description?

subramen (changes requested):
The states are indeed not replicated, but I believe the author intended to say that the optimizer's initial hyperparameters (lr, seed, momentum etc.) are the same across all devices. I suggest adding that clarity here instead of deleting the point about replicating optimizers

timgianitsos:
Thanks! I interpret you as making a distinction between optimizer state synchronization at (a) initialization versus (b) after each step, and you argue "yes" for the former and "no" for the latter. But I am arguing that there is no optimizer synchronization in either case.

If the optimizers are incidentally initialized to be the same across processes as you say (which I concede will often be the case) that's only because of a decision the user made (or because of the user's oversight as this is default behavior) and NOT because DDP has any effect on the synchronization. That is, if you just create an optimizer like normal in a function that is sent to `multiprocess.spawn`, then they will be identical (e.g. `opt = RMSprop(ddp.parameters(), **opt_kwargs)`). The user could make them different if they said `if rank == x: <do something different>`.

The way the docs were worded before my edit, it seems to imply that the optimizers are initialized to be identical _because_ DDP enforces this. But if so, then I don't understand the mechanism - the only dependency between DDP and the optimizers is the passing of `ddp.parameters()`. This is a generator which simply `yield`s each layer as a `nn.Parameter`. This seems absent of any information that the optimizer could use to tell if the model is being distributed. From this, I conclude that the optimizers on different processes are NOT syncing with each other.

subramen (approved):

rohan-varma (approved):
I guess this is technically right, we don't explicitly sync / replicate optimizer states, but they are usually the same since the optimizer acts on the same parameters and gradients.

timgianitsos:
Please advise

timgianitsos:
Please advise on what is causing the delay, @rohan-varma @subramen. It's a very simple change.

PR #2436, opened 2023-06-06
Update torch_compile_tutorial.py to use unused parameter

laithhas:
I noticed when reading through these docs that the two examples did not use the parameter 'y'. I assume it was meant to be used so I updated the code in the examples. Another possibility is that we don't need param 'y' and only need 'x'. Let me know if that is the case and I will fix this :)


williamwen42 (approved):
Looks to be a mistake on my end - thanks for the catch!

malfet (approved):

PR #2438, opened 2023-06-06
Set random seed

clee2000:
To make tutorial builds predictable, but still keep randomness when one rans it on Collab.
Also, reset default_device after every tutorial run

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -88,7 +88,7 @@ intersphinx_mapping = {
 
 def reset_seeds(gallery_conf, fname):
     torch.manual_seed(42)
-    torch.set_default_device('cpu')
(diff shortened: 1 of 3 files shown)


conf.py
```diff
@@ -85,6 +86,10 @@
 
 # -- Sphinx-gallery configuration --------------------------------------------
 
+def reset_seeds(gallery_conf, fname):
+    torch.manual_seed(42)
+    random.seed(10)
```
malfet, line 91: Nit
```suggestion
    torch.set_default_device('cpu')
    random.seed(10)
```

malfet (approved):

svekars (approved):

PR #2443, opened 2023-06-07
Increase the number of workers

svekars:
Changed the number of workers to 15 in attempt to speed up the build.


malfet (approved):
Looks good to me, can you please add before/after numbers to the PR description

Issue #810, opened 2020-01-08
Code in Adversarial examples tutorial needs to be vectorized and it is too slow. 


AmrKhalifa:
The code can be vectorized to work on a batch of examples instead of 1 example at a time. 
I modified it in my implementation, and I also added code for different attacks, 
please check the following code to see my implementation. 

https://github.com/AmrKhalifa/Utilizing-Variational-Auto-Encoder-Representations-to-Enhance-Neural-Networks-Adversarial-Robustness/blob/master/Code/vae_classifier_attacks.py

check the function :
test_attack()

 @inkawhich

mikebrow:
/assigntome

finished the other assigned ready for the next one :)

PR #2483, opened 2023-06-23
add vectorization options as a goal target

mikebrow:
Consider closing #810 in concert with fix #2419

The original issue mentioned a performance problem and a suggestion for going beyond the simple test loop to vectorizing the processing in some form, batch / parallel etc.

However, fix 2419 linked above, addressed the performance problem by switching from always cpu to device and by correcting other inconsistencies. Thanks!

Further, this is targeted as a beginner example, and with the above corrections I believe this example now stands on it's own as is.

## Description
This change merely proposes to the reader, in the where to go section, that:
```
# A further direction to go, depending on available resources, is to modify
# the code to support processing work in batch, in parallel, and or distributed
# vs. working on one attack at a time in the above for each epsilon test() loop.
```

## Checklist


beginner_source/fgsm_tutorial.py
```diff
@@ -447,4 +447,8 @@ def test( model, device, test_loader, epsilon ):
 # implement a different attack from the NIPS 2017 competition, and see how
 # it differs from FGSM. Then, try to defend the model from your own
 # attacks.
-# 
+#
+# A further direction to go, depending on available resources, is to modify
+# the code to support processing work in batch, in parallel, and or distributed
+# vs. working on one attack at a time in the above for each epsilon test() loop.
```
svekars, line 453: Do you want to add examples of those?

mikebrow: thought about it and was thinking this is more a beginner example and appropriate for this section.. Perhaps a second example addressing one or more of the above concerns is warranted and we could link from here?

svekars, line 453:
```suggestion
# vs working on one attack at a time in the above for each ``epsilon test()`` loop.
```

svekars (approved):

Issue #208, opened 2018-02-06
No .eval() in tutorials


Kaixhin:
See https://github.com/pytorch/vision/pull/405 for a doc update and link to a forum thread. It seems that actually people following any of the basic tutorials will miss `.eval()`, which admittedly is not as easy to come across as it should be. We should update some of the basic tutorials to add this note.

Kaixhin:
[This keeps on getting missed](https://twitter.com/quasimondo/status/1021669507184644097).

kwasnydam:
Actually I was just about opening an issue on this. In the _TRANSLATION WITH A SEQUENCE TO SEQUENCE NETWORK AND ATTENTION_, [(https://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html)](url), 
under the Evaluation section the author is missing the _.eval()_ call on the models when performing the evaluation, and he uses a dropout in his model. Is it possible to push a fix for that? Never have contributed on github yet, so do not really know the procedures.

Kaixhin:
@kwasnydam helping fix an error you've found is very helpful, so thanks. Rough steps are as follows, :

1. Fork this repository
2. Add the missing calls in the [file](https://github.com/pytorch/tutorials/blob/master/intermediate_source/seq2seq_translation_tutorial.py)
3. In your forked repository, start a pull request (button should be near the clone button once you've added a commit)
4. Fill in the details and submit

PR #2484, opened 2023-06-23
set eval mode before running evaluation

mikebrow:
Partial address of #208
Specifically address comment https://github.com/pytorch/tutorials/issues/208#issuecomment-484789093

## Description
set eval mode before running evaluation

## Checklist

cc @albanD


albanD (approved):
Thanks!

PR #2494, opened 2023-06-28
Update pyspelling config to skip code-blocks with 3+ indents. 

svekars:
Update pyspelling config to skip code-blocks with 3+ indents.


malfet (changes requested):
Why not simplify condition above?

PR #2500, opened 2023-07-01
gramatically updated sentence

Subhamsarkar121:
Fixes #ISSUE_NUMBER

## Description
<! The sentence -Tensors are similar to [NumPy’s](https://numpy.org/) ndarrays, except that tensors can run on GPUs or other hardware accelerators. 
Has been updated to -Tensors are similar to [NumPy’s](https://numpy.org/) ndarrays, except those tensors can run on GPUs or other hardware accelerators.>


svekars (commented):
I don't think this is an improvement and it changes the meaning. This sentence should remain as is.

PR #2501, opened 2023-07-03
updating a typo :-)

abhi-11nav:
Fixes #ISSUE_NUMBER

## Checklist


albanD (changes requested):
This is already in the loss function: https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html 
So we don't need it I think.

PR #2503, opened 2023-07-05
[export] Added a tutorial for export

angelayi:
This requires torch._export which is not available in the pytorch 2.0 release but will be in the 2.1 release. Not sure why this is a problem, but Omkar said it should be ok!

### DIFF
--- a/prototype_source/export_tutorial.py
+++ b/prototype_source/export_tutorial.py
@@ -1,17 +1,17 @@
 """
-(prototype) torch.export
+Introducing ``torch.export``
 ========================
 
 This tutorial introduces steps to export a PyTorch program using the PyTorch 2.0 features.
 
-To begin, certain prerequisites need to be met in order to get an exported
+The following prerequisites must be met to get an exported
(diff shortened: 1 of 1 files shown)


svekars (changes requested):
Thanks so much @angelayi! I added some editorial suggestions. Let me know if you have any questions!

zhxchen17 (changes requested):
I did a quick scan of this and I feel we might need some rework on this tutorial. For example, the first thing I saw is this function:

```
def check_soundness(f: Callable, exported_program: ExportedProgram, *inputs):
    try:
        print(
            "Sound!"
            if [torch.equal](https://pytorch.org/docs/stable/generated/torch.equal.html#torch.equal)(f(*inputs), exported_program(*inputs))
            else "Not Sound!"
        )
    except Exception as e:
        print("Raise runtime error!")
        raise e
```

why this is even relevant to users? Ideally we should start with a simple example demonstrating the most basic and common usage, then progressively go to a set of important features.

If you wanna talk about this we can follow up offline :)

PR #2507, opened 2023-07-25
Update README.md

atalman:
Update Feature classification, to reflect current reality. The prototype features are included in the release and are part of released binaries. See: https://github.com/pytorch/pytorch.github.io/pull/1414

### DIFF
--- a/prototype_source/README.md
+++ b/prototype_source/README.md
@@ -2,7 +2,7 @@
 
 This directory contains tutorials and recipes demonstrating prototype features in PyTorch. 
 
-**Prototype features** are not available as part of binary distributions like PyPI or Conda (except maybe behind run-time flags). To test these features we would, depending on the feature, recommend building from master or using the nightly wheels that are made available on pytorch.org. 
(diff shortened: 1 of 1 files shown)


svekars (approved):

prototype_source/README.md
```diff
@@ -2,7 +2,7 @@
 
 This directory contains tutorials and recipes demonstrating prototype features in PyTorch. 
 
-**Prototype features** are not available as part of binary distributions like PyPI or Conda (except maybe behind run-time flags). To test these features we would, depending on the feature, recommend building from master or using the nightly wheels that are made available on pytorch.org. 
+**Prototype features** are part of the release and are available as part of binary distributions like PyPI or Conda. To test these features we would, depending on the feature, recommend building from master or using the nightly wheels that are made available on pytorch.org or using release wheels available from PyPI or Conda.
```
malfet, line 5: Sorry, but who are `we` in this case? I would've rephrased it to something like `to have a full experience of those features it is recommended to use nightly builds or build from source, if feature is gated behind a compiler flag`

Also, I would clarify, that prototype feature are a preview of what will soon become available, but might undergone some changes, so it's not recommended to build ones production pipelines around those.

PR #2508, opened 2023-07-26
Fix tutorial dependencies install

malfet:
`/opt/conda/bin` folder contains default Python (3.11 after the last update) rather than the one specified promised by the docker container (which is installed in the environment)

Remove explicit `PATH` to stick to Python-3.10 installed as part of the container (as `torchx` is yet not available for Python-3.11)

Also, uninstall `ghstack` to prevent downgrade `importlib-metadata` from `6.8.0` to `3.X`.
Pin `importlib-metadata`, `scipy`, `numba` versions. 

Minor: Add `--progress-bar off` option, to keep logs readable.

### DIFF
--- a/.jenkins/build.sh
+++ b/.jenkins/build.sh
@@ -18,6 +18,8 @@ sudo apt-get install -y --no-install-recommends unzip p7zip-full sox libsox-dev 
 # NS: Path to python runtime should already be part of docker container
 # export PATH=/opt/conda/bin:$PATH
 rm -rf src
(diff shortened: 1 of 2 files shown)


malfet:
Ok, somehow it needs `pip install importlib-metadata==6.8.0`

malfet:
t5_tutorial currently fails as http://www.quest.dcs.shef.ac.uk/wmt16_files_mmt/training.tar.gz is unaccessible at the moment

PR #2516, opened 2023-08-02
Update ddp_tutorial.rst

HUGHNew:
## Description
1. add `dist.destroy_process_group()` in example code block to make sure release the resources
2. modify the link syntax error about `torchrun`

## Checklist

cc @osalpekar @H-Huang @kwen2501

### DIFF
--- a/intermediate_source/ddp_tutorial.rst
+++ b/intermediate_source/ddp_tutorial.rst
@@ -340,11 +340,12 @@ Let's still use the Toymodel example and create a file named ``elastic_ddp.py``.
         labels = torch.randn(20, 5).to(device_id)
         loss_fn(outputs, labels).backward()
         optimizer.step()
+        dist.destroy_process_group()
         
     if __name__ == "__main__":
         demo_basic()
 
-One can then run a `torch elastic/torchrun<https://pytorch.org/docs/stable/elastic/quickstart.html>`__ command 
+One can then run a `torch elastic/torchrun <https://pytorch.org/docs/stable/elastic/quickstart.html>`__ command 
 on all nodes to initialize the DDP job created above:
 
 .. code:: bash


svekars:
@H-Huang can you take a look?

intermediate_source/ddp_tutorial.rst
```diff
@@ -340,11 +340,12 @@ Let's still use the Toymodel example and create a file named ``elastic_ddp.py``.
         labels = torch.randn(20, 5).to(device_id)
         loss_fn(outputs, labels).backward()
         optimizer.step()
+        dist.destroy_process_group()
```
H-Huang, line 343: This is actually not necessary. After the script is finished the process_group will be garbage collected by Python.

`destroy_process_group` clears up the process_group state so that any references to it are freed and it can be garbage collected. Since this is the last line of the script, all references will be cleaned up anyways.

HUGHNew: Since the other examples in the same tutorial page call `destroy_process_group`, it's natural for me to complement this function call in this exmaple code to make they consistent.
On the other hand, after I send a SIGINT signal to terminate the DDP training, the used GPUs don't release the memory immediately. Therefore I think it's also necessary to call `destroy_process_group` to notify the nvidia driver to collect the graphic memory.

H-Huang: https://github.com/pytorch/pytorch/blob/0dc7f6df9d00428cd175018e2bf9b45a8ec39b9c/torch/distributed/distributed_c10d.py#L1359-L1425

If you look at the implementation for `destroy_process_group` you can see its pretty simple and we are just clearing some maps or deleting dictionaries that reference the process group which allow it to be GCed.

I checked and you are right that the rest of the examples use a `cleanup()` function to do this. I'm okay with adding this in to be consistent then

> On the other hand, after I send a SIGINT signal to terminate the DDP training, the used GPUs don't release the memory immediately. Therefore I think it's also necessary to call destroy_process_group to notify the nvidia driver to collect the graphic memory.

For this, can you clarify? If there is memory leakage after the script is finished executing then this is definitely an issue and we should file a bug and fix this.

H-Huang (approved):
Thanks, new doc snippet looks like this: 

image

LGTM

intermediate_source/ddp_tutorial.rst
HUGHNew: I have suffered the GPU memory leakage recently. I think I should do some experiments for reproduction to verify my statement before a issue.

PR #2517, opened 2023-08-02
Pyspelling: add config for beginner/.rst files

svekars:
* Add the reST job
* Fixed typos in the beginner_source/.rst files to pass the check


malfet (approved):
In theory, looks good to me, but those context exclusions create a possibility, that unbalanced context will take too long to parse or just ignore a much larger chunk of the file than expected..
Can we print total number of words validated and say it should be no less than 50% of total number of worlds in the files in question

svekars:
Will handle the check in a separate PR.

PR #2519, opened 2023-08-03
[wip] Loading nn.Module from checkpoint tutorial

mikaylagawarecki:
* __->__ #2519

### DIFF
--- a/recipes_source/recipes/module_load_state_dict_tips.py
+++ b/recipes_source/recipes/module_load_state_dict_tips.py
@@ -1,16 +1,15 @@
-# -*- coding: utf-8 -*-
 """
 
-Tips for Loading an nn.Module from a Checkpoint
-===============================================
+Tips for Loading an `nn.Module` from a Checkpoint
+=================================================
 
 In this tutorial, we will share some tips for loading a model from a checkpoint.
 In particular, we will discuss
 
 
-1.   The `mmap` kwarg on `torch.load`
-2.   The `torch.device()` context manager
-3.  The `assign` kwarg on `nn.Module.load_state_dict()`
+1.  The `mmap` keyword argument on `torch.load`
+2.  The `torch.device()` context manager
+3.  The `assign` keyword argument on `nn.Module.load_state_dict()`
 
 The following snippet of code illustrates the use of the above three utilities.
 """
@@ -33,7 +32,7 @@ m.load_state_dict(state_dict)
 ###############################################################################
 # At `torch.save` time, tensor storages are tagged with the device they are
 # saved on. At `torch.load` time, tensor storages will be loaded to the device
-# they were tagged with (unless this behavior is overriden using the
+# they were tagged with (unless this behavior is overridden using the
 # `map_location` flag). For ease of explanation, let us assume that the tensors
 # were saved on CPU. This means that on the first line
 
@@ -45,12 +44,12 @@ state_dict = torch.load('checkpoint.pth')
 #     2. The user does not want to wait for the entire checkpoint to be loaded
 #        into RAM before doing for example some per-tensor processing
 #
-# The `mmap` kwarg to `torch.load` attempts to solve the above two problems by
-# using an [mmap call](https://man7.org/linux/man-pages/man2/mmap.2.html) on
-# the checkpoint, so that tensor storages are memory-mapped and when they are
+# The `mmap` keyword argument to `torch.load` attempts to solve the above two
+# problems by using an [`mmap` call](https://man7.org/linux/man-pages/man2/mmap.2.html)
+# on the checkpoint, so that tensor storages are memory-mapped and when they are
 # fetched from disk to memory is managed by the OS.
 #
-# Next, we consider the instantiation of the module.
+# Next, we consider the creation of the module.
 
 m = SomeModule()
 
@@ -75,7 +74,6 @@ with torch.device('meta'):
 # context manager makes sure that factory calls will be performed as if they
 # were passed device as an argument. However, it does not affect factory
 # function calls which are called with an explicit device argument.
-# TODO: maybe include recent PR to override part 2 with TorchFunctionMode
 #
 # Tensors on the `meta` device do not carry data. However, they possess all
 # other metadata a tensor carries such as `.size()` and `.stride()`,
@@ -92,7 +90,7 @@ m.load_state_dict(state_dict)
 # the parameters/buffers in the `nn.Module`).
 #
 # However, an in-place copy into a tensor on the `meta` device is a no-op.
-# In order to avoid this, we can pass the `assign=True` kwarg to
+# In order to avoid this, we can pass the `assign=True` keyword argument to
 # `load_state_dict()`.
 
 meta_m.load_state_dict(state_dict, assign=True)
@@ -103,7 +101,7 @@ meta_m.load_state_dict(state_dict, assign=True)
 # is loaded from state dict if `assign=True` is passed.
 
 ###############################################################################
-# To recap, in this tutorial, we learnt about `torch.load(mmap=True)`, the
+# To recap, in this tutorial, we learned about `torch.load(mmap=True)`, the
 # `torch.device()` context manager with `device=meta` and the
 # `nn.Module.load_state_dict(assign=True)` and how these tools could be used
 # to aid when loading a model from a checkpoint.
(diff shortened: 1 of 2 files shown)


svekars:
Thanks, @mikaylagawarecki! Can you please resubmit as a regular PR? We don't support ghtstack in this repo.

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,107 @@
...
+
+state_dict = torch.load('checkpoint.pth')
+
+###############################################################################
+# all tensor storages will be loaded into CPU RAM, which can be problematic when
+#     1. CPU RAM is smaller than the size of the checkpoint
+#     2. The user does not want to wait for the entire checkpoint to be loaded
+#        into RAM before doing for example some per-tensor processing
+#
+# The `mmap` keyword argument to `torch.load` attempts to solve the above two
```
mikaylagawarecki, line 47: how to better explain this, 
1) should we go into internals of how the zip file is structured
2) should we talk about don't expect it to be mmap on CUDA

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,107 @@
...
+# In order to solve these two problems, we can use the `torch.device()`
+# context manager with `device='meta'` when we instantiate the `nn.Module()`.
+
+with torch.device('meta'):
+  meta_m = SomeModule()
+
+###############################################################################
+# The [`torch.device()`](https://pytorch.org/docs/main/tensor_attributes.html#torch-device)
+# context manager makes sure that factory calls will be performed as if they
+# were passed device as an argument. However, it does not affect factory
```
mikaylagawarecki, line 75: There is a [new PR (not landed yet)](https://github.com/pytorch/pytorch/pull/106514) that lets this "not affecting factory calls be overridden by function calls with a explicit device argument" be overriden with TorchFunctionMode, should that be mentioned?

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,107 @@
...
+# they were tagged with (unless this behavior is overridden using the
+# `map_location` flag). For ease of explanation, let us assume that the tensors
+# were saved on CPU. This means that on the first line
+
+state_dict = torch.load('checkpoint.pth')
+
+###############################################################################
+# all tensor storages will be loaded into CPU RAM, which can be problematic when
+#     1. CPU RAM is smaller than the size of the checkpoint
+#     2. The user does not want to wait for the entire checkpoint to be loaded
```
mikaylagawarecki, line 44: Should I include an example of 1 by 1 processing with actual timings of this step? 

Or, as a broader question, do we want to include timing code for all of these?

mikaylagawarecki (commented):
@svekars I will open a proper non-ghstack PR for landing, just using this one as a scratch for review for now if that's alright!

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,107 @@
...
+
+In this tutorial, we will share some tips for loading a model from a checkpoint.
+In particular, we will discuss
+
+
+1.  The `mmap` keyword argument on `torch.load`
+2.  The `torch.device()` context manager
+3.  The `assign` keyword argument on `nn.Module.load_state_dict()`
+
+The following snippet of code illustrates the use of the above three utilities.
```
albanD, line 14: I like the idea of having a "tl;dr:" at the top of the tutorial, we can make it more explicit: tl;dr: if you're loading a checkpoint and want to reduce compute and memory as much as posisble, do the following:

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,107 @@
...
+"""
+
+import torch
+
+state_dict = torch.load('checkpoint.pth', mmap=True)
```
albanD, line 19: This code is run when rendering, so this should be valid code!

recipes_source/recipes/module_load_state_dict_tips.py
albanD: nit: this is the only place where you use "The user". most likely should be rephrased to match the rest.

Adding timing code should be simple since this code is ran to generate the tutorial. So just use time() and print :)

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,107 @@
...
+
+###############################################################################
+# all tensor storages will be loaded into CPU RAM, which can be problematic when
+#     1. CPU RAM is smaller than the size of the checkpoint
+#     2. The user does not want to wait for the entire checkpoint to be loaded
+#        into RAM before doing for example some per-tensor processing
+#
+# The `mmap` keyword argument to `torch.load` attempts to solve the above two
+# problems by using an [`mmap` call](https://man7.org/linux/man-pages/man2/mmap.2.html)
+# on the checkpoint, so that tensor storages are memory-mapped and when they are
```
albanD, line 49: nit: I think this is a bit too much "in the middle": if you assume that the user doesn't know mmap, you should have one sentence that explains what it does (map the file on disk into virtual memory and let the OS handle loading/unloading into physical memory automatically). Or if you assume they already know, you can simplify this paragraph.

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,107 @@
...
+# fetched from disk to memory is managed by the OS.
+#
+# Next, we consider the creation of the module.
```
albanD, line 52: Have titles? Not sure how this renders

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,107 @@
...
+
+m = SomeModule()
+
+###############################################################################
+# This allocates memory for all parameters/buffers and initializes them per
+# the default initialization schemes defined in `SomeModule.__init__()`, which
+# is wasteful when we want to load a checkpoint as
+#     1. We are running the initialization kernels where the results are
+#        immediately overwritten by `load_state_dict()`
```
albanD, line 61: nit: the sentence the other way sounds simpler: the result will be overriden by load_state_dict without ever being read. So initializing it is not useful

recipes_source/recipes/module_load_state_dict_tips.py
albanD: I think this tutorial should consider the user is using the latest version of main (so after the PR is landed).

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,107 @@
...
+# The [`torch.device()`](https://pytorch.org/docs/main/tensor_attributes.html#torch-device)
+# context manager makes sure that factory calls will be performed as if they
+# were passed device as an argument. However, it does not affect factory
+# function calls which are called with an explicit device argument.
+#
+# Tensors on the `meta` device do not carry data. However, they possess all
+# other metadata a tensor carries such as `.size()` and `.stride()`,
+# `.requires_grad` etc.
+#
+# Next, we consider the loading of the state dictionary.
```
albanD, line 82: paragraph

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,107 @@
...
+# context manager with `device='meta'` when we instantiate the `nn.Module()`.
+
+with torch.device('meta'):
+  meta_m = SomeModule()
+
+###############################################################################
+# The [`torch.device()`](https://pytorch.org/docs/main/tensor_attributes.html#torch-device)
+# context manager makes sure that factory calls will be performed as if they
+# were passed device as an argument. However, it does not affect factory
+# function calls which are called with an explicit device argument.
```
albanD, line 76: This point is subtle and should most likely be covered with examples in the torch.device doc. (no need to have example for this here).

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,107 @@
...
+# the parameters/buffers in the `nn.Module`).
+#
+# However, an in-place copy into a tensor on the `meta` device is a no-op.
+# In order to avoid this, we can pass the `assign=True` keyword argument to
+# `load_state_dict()`.
+
+meta_m.load_state_dict(state_dict, assign=True)
+
+###############################################################################
+# Another caveat here is that since optimizers hold a reference to
```
albanD, line 99: Which one is the first for this one to be another? :D

Also I would mention that as a warning and give a code sample:
```
model.load_state_dict(state_dict, assign=True)
# This MUST be done AFTER the assigning load_state_dict.
opt = SGD(model.parameters(), ...)
```

PR #2522, opened 2023-08-08
Loading nn.Module from checkpoint tutorial (non ghstack)

mikaylagawarecki:
Added tutorial for some utilities for loading checkpoint 

1.  The ``mmap`` keyword argument on ``torch.load``
2.  The ``torch.device()`` context manager
3.  The ``assign`` keyword argument on ``nn.Module.load_state_dict()``

### DIFF
--- a/recipes_source/recipes/module_load_state_dict_tips.py
+++ b/recipes_source/recipes/module_load_state_dict_tips.py
@@ -50,6 +50,7 @@ state_dict = torch.load('checkpoint.pth')
 m = SomeModule(1000)
 m.load_state_dict(state_dict)
 
+#############################################################################
 # The second example does not use any of the features listed above and will be
 # less compute and memory efficient for loading a checkpoint. In the following
 # sections, we will discuss each of the features in further detail.
@@ -57,7 +58,7 @@ m.load_state_dict(state_dict)
 #####################################################################################
 # Using ``torch.load(mmap=True)``
 # -------------------------------
-# First, let us consider what happens when we load the checkpoint with``torch.load``.
+# First, let us consider what happens when we load the checkpoint with ``torch.load``.
 # When we save a checkpoint with ``torch.save``, tensor storages are tagged with the device they are
 # saved on. With ``torch.load``, tensor storages will be loaded to the device
 # they were tagged with (unless this behavior is overridden using the
@@ -66,8 +67,7 @@ m.load_state_dict(state_dict)
 # loaded into CPU RAM, which can be undesirable when:
 #
 # * CPU RAM is smaller than the size of the checkpoint.
-# * Waiting for the entire checkpoint to be loaded into RAM before
-#    performing, for example, some per-tensor processing.
+# * Waiting for the entire checkpoint to be loaded into RAM before performing, for example, some per-tensor processing.
 
 start_time = time.time()
 state_dict = torch.load('checkpoint.pth')
@@ -114,10 +114,11 @@ m = SomeModule(1000)
 # This allocates memory for all parameters/buffers and initializes them per
 # the default initialization schemes defined in ``SomeModule.__init__()``, which
 # is wasteful when we want to load a checkpoint for the following reasons:
-# * The result of the initialization kernels will be overwritten by ``load_state_dict()``
-#    without ever being used, so initialization is wasteful.
-# * We are allocating memory for these parameters/buffers in RAM while ``torch.load`` of
-#    the saved state dictionary also allocates memory in RAM for the parameters/buffers in the checkpoint.
+#
+# * The result of the initialization kernels will be overwritten by ``load_state_dict()`` without ever being used, so
+#   initialization is wasteful.
+# * We are allocating memory for these parameters/buffers in RAM while ``torch.load`` of the saved state dictionary also
+#   allocates memory in RAM for the parameters/buffers in the checkpoint.
 #
 # In order to solve these two problems, we can use the ``torch.device()``
 # context manager with ``device='meta'`` when we instantiate the ``nn.Module()``.
--- a/recipes_source/recipes/module_load_state_dict_tips.py
+++ b/recipes_source/recipes/module_load_state_dict_tips.py
@@ -1,0 +1,107 @@
+"""
+
+Tips for Loading an `nn.Module` from a Checkpoint
+=================================================
+
+In this tutorial, we will share some tips for loading a model from a checkpoint.
+In particular, we will discuss
+
+
+1.  The `mmap` keyword argument on `torch.load`
+2.  The `torch.device()` context manager
+3.  The `assign` keyword argument on `nn.Module.load_state_dict()`
+
+The following snippet of code illustrates the use of the above three utilities.
+"""
+
+import torch
+
+state_dict = torch.load('checkpoint.pth', mmap=True)
+with torch.device('meta'):
+  meta_m = SomeModule()
+meta_m.load_state_dict(state_dict, assign=True)
+
+###############################################################################
+# Taking a step back, let us inspect the following more vanilla code snippet
+# that does not use any of the features listed above
+
+state_dict = torch.load('checkpoint.pth')
+m = SomeModule()
+m.load_state_dict(state_dict)
+
+###############################################################################
+# At `torch.save` time, tensor storages are tagged with the device they are
+# saved on. At `torch.load` time, tensor storages will be loaded to the device
+# they were tagged with (unless this behavior is overridden using the
+# `map_location` flag). For ease of explanation, let us assume that the tensors
+# were saved on CPU. This means that on the first line
+
+state_dict = torch.load('checkpoint.pth')
+
+###############################################################################
+# all tensor storages will be loaded into CPU RAM, which can be problematic when
+#     1. CPU RAM is smaller than the size of the checkpoint
+#     2. waiting for the entire checkpoint to be loaded into RAM before
+#        doing for example some per-tensor processing
+#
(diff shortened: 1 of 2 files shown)


recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,156 @@
+"""
+
+Tips for Loading an ``nn.Module`` from a Checkpoint
+===================================================
+
+If you're loading a checkpoint and want to reduce compute and memory as much as possible,
+this tutorial shares some recommended practices. In particular, we will discuss
```
svekars, line 7:
```suggestion
this tutorial shares some recommended practices. In particular, we will discuss the following:
```

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,156 @@
...
+2.  The ``torch.device()`` context manager
+3.  The ``assign`` keyword argument on ``nn.Module.load_state_dict()``
+
+.. note::
+   This recipe requires PyTorch 2.1.0 or later.
+"""
+
+
+########################################
+# Let us consider a simple ``nn.Module``
```
svekars, line 19: Would be good add a bit more description here about what the code below does.

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def forward(self, x):
+        return self.linears(x)
+
+
+m = SomeModule(1000)
+torch.save(m.state_dict(), 'checkpoint.pth')
+
+#################################################################
+# The follow snippet demonstrates the use of the three utilities.
```
svekars, line 37: can you list the utilities.

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,156 @@
...
+
+state_dict = torch.load('checkpoint.pth', mmap=True)
+with torch.device('meta'):
+  meta_m = SomeModule(1000)
+meta_m.load_state_dict(state_dict, assign=True)
+
+#############################################################################
+# Taking a step back, let us inspect the following more vanilla code snippet
+# that does not use any of the features listed above:
```
svekars, line 46:
```suggestion
# Compare the snippet below to the one above:
```

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,156 @@
...
+
+state_dict = torch.load('checkpoint.pth')
+m = SomeModule(1000)
+m.load_state_dict(state_dict)
+
```
svekars, line 51:
```suggestion
# The second example does not use any of the features listed above.
```
Can you add a conclusion about this comparison?

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,156 @@
...
+# First let us consider what happens when we ``torch.load`` the checkpoint.
+# At ``torch.save`` time, tensor storages are tagged with the device they are
+# saved on. At ``torch.load`` time, tensor storages will be loaded to the device
+# they were tagged with (unless this behavior is overridden using the
+# ``map_location`` flag). For ease of explanation, let us assume that the tensors
+# were saved on CPU. This means that on the first line all tensor storages will be
+# loaded into CPU RAM, which can be undesirable when
+#     1. CPU RAM is smaller than the size of the checkpoint
+#     2. Waiting for the entire checkpoint to be loaded into RAM before
+#        doing for example some per-tensor processing
```
svekars, line 64:
```suggestion
# First, let us consider what happens when we load the checkpoint with``torch.load``.
# When we save a checkpoint with ``torch.save``, tensor storages are tagged with the device they are
# saved on. With ``torch.load``, tensor storages will be loaded to the device
# they were tagged with (unless this behavior is overridden using the
# ``map_location`` flag). For ease of explanation, let us assume that the tensors
# were saved on CPU. This means that on the first line all tensor storages will be
# loaded into CPU RAM, which can be undesirable when:
#
# * CPU RAM is smaller than the size of the checkpoint.
# * Waiting for the entire checkpoint to be loaded into RAM before
#    performing, for example, some per-tensor processing.
```

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,156 @@
...
+# storages will be memory-mapped.
+
+start_time = time.time()
+state_dict = torch.load('checkpoint.pth', mmap=True)
+end_time = time.time()
+print(f"loading time with mmap={end_time - start_time}")
+
+######################################################################################
+# As mentioned above, one can use this argument to do per-tensor processing on a
+# checkpoint without loading all tensor storages into CPU memory upfront. For example,
```
svekars, line 86:
```suggestion
# As mentioned above, one can use this argument to do per-tensor processing on a
# checkpoint without loading all tensor storages into CPU memory upfront. For example:
```

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,156 @@
...
+    return processed_t
+
+for key in state_dict.keys():
+    device = torch.device('cuda:' + str(int(key.lstrip("linears.")[0]) % 8))
+    state_dict[key] = my_processing_function(key, device)
+
+##############################################
+# Using ``torch.device('meta')``
+# ------------------------------
+# Next, we consider the creation of the module.
```
svekars, line 104:
```suggestion
# Next, let's consider the creation of the module.
```

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,156 @@
...
+#        the saved state dictionary also allocates memory for the parameters/buffers in the checkpoint.
+#
+# In order to solve these two problems, we can use the ``torch.device()``
+# context manager with ``device='meta'`` when we instantiate the ``nn.Module()``.
+#
+# The `torch.device() <https://pytorch.org/docs/main/tensor_attributes.html#torch-device>`_
+# context manager makes sure that factory calls will be performed as if they
+# were passed the specified ``device`` as an argument. Tensors on ``torch.device('meta')`` do not
+# carry data. However, they possess all other metadata a tensor carries such as ``.size()``, ``.stride()``
+# and ``.requires_grad`` etc.
```
svekars, line 123:
```suggestion
# is wasteful when we want to load a checkpoint as one of the following:
# * The result of the initialization kernels will be overwritten by ``load_state_dict()``
#    without ever being used, so initialization is wasteful.
# * We are allocating memory for these parameters/buffers in RAM while ``torch.load`` of
#    the saved state dictionary also allocates memory for the parameters/buffers in the checkpoint.
#
# In order to solve these two problems, we can use the ``torch.device()``
# context manager with ``device='meta'`` when we instantiate the ``nn.Module()``.
#
# The `torch.device() <https://pytorch.org/docs/main/tensor_attributes.html#torch-device>`_
# context manager makes sure that factory calls will be performed as if they
# were passed the specified ``device`` as an argument. Tensors on ``torch.device('meta')`` do not
# carry data. However, they possess all other metadata a tensor carries such as ``.size()``, ``.stride()``,
# ``.requires_grad``, and other.
```

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,156 @@
...
+# --------------------------------------
+# Next, we consider the loading of the state dictionary.
+
+m.load_state_dict(state_dict)
+
+###############################################################################
+# ``nn.Module.load_state_dict()`` is usually implemented via an in-place
+# ``param_in_model.copy_(param_in_state_dict)`` (i.e. a copy from the
+# parameter/buffer with the corresponding key in the state dictionary into
+# the parameter/buffer in the ``nn.Module``).
```
svekars, line 138:
```suggestion
# ``param_in_model.copy_(param_in_state_dict)``. For example, a copy from the
# parameter/buffer with the corresponding key in the state dictionary into
# the parameter/buffer in the ``nn.Module``.
```

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,156 @@
...
+# A caveat here is that since optimizers hold a reference to
+# ``nn.Module.parameters()``, the optimizer must be initialized after the module
+# is loaded from state dict if ``assign=True`` is passed.
+
+new_m.load_state_dict(state_dict, assign=True)
+# This MUST be done AFTER the load_state_dict with assign.
+opt = torch.optim.SGD(new_m.parameters(), lr=1e-3)
+
+###############################################################################
+# To recap, in this tutorial we learned about ``torch.load(mmap=True)``, the
```
svekars, line 153:
```suggestion
# Conclusion
# ------------
# 
# To recap, in this tutorial we learned about ``torch.load(mmap=True)``, the
```

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -0,0 +1,156 @@
...
+# ``torch.device()`` context manager with ``device=meta`` and
```
svekars, line 154:
```suggestion
# ``torch.device()`` context manager with ``device=meta``, and
```

svekars (commented):
thanks, @mikaylagawarecki. Looks good, just a couple of editorial suggestions.

albanD (approved):
SGTM

svekars (approved):

PR #2524, opened 2023-08-12
Move build manager to pytorchbot-env workflow environment if run on main

clee2000:
As in title
The pytorchbot-env is restricted to the main branch

The if statement is there because if you have the env there but don't qualify for the restriction, the workflow doesn't run at all.


atalman:
The environment needs to be added aftere line 35 as well, since line 43 uses secrets

atalman (approved):

PR #2535, opened 2023-08-25
conv-bn folding should be used for CNN-based Vision Models when AMP is used with oneDNN Graph

sanchitintel:
By default, conv-bn folding isn't done with CNN based models when AMP is used with oneDNN Graph. `torch.fx.experimental.optimize.fuse` should be used for such models

### DIFF
--- a/recipes_source/recipes/tuning_guide.py
+++ b/recipes_source/recipes/tuning_guide.py
@@ -295,7 +295,7 @@ with torch.no_grad():
 torch._C._jit_set_autocast_mode(False)
 
 with torch.no_grad(), torch.cpu.amp.autocast(cache_enabled=False, dtype=torch.bfloat16):
-    # Conv-BatchNorm folding for CNN-based Vision Models should be done with torch.fx.experimental.optimization.fuse when AMP is used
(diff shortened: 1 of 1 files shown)


msaroufim:
This seems fine, I'm a bit confused by `torch.fx.experimental` is a public API though, like why call it experimental then?

msaroufim (approved):

sanchitintel:
Thanks for reviewing, @msaroufim! :)
I'm also not sure why it wasn't renamed.

recipes_source/recipes/tuning_guide.py
```diff
@@ -295,6 +295,10 @@ def fused_gelu(x):
 torch._C._jit_set_autocast_mode(False)
 
 with torch.no_grad(), torch.cpu.amp.autocast(cache_enabled=False, dtype=torch.bfloat16):
+    # Conv-BatchNorm folding for CNN-based Vision Models should be done with torch.fx.experimental.optimization.fuse when AMP is used
```
svekars, line 298:
```suggestion
    # Conv-BatchNorm folding for CNN-based Vision Models should be done with ``torch.fx.experimental.optimization.fuse`` when AMP is used
```
To fix spellcheck.

sanchitintel: Thanks for your help, @svekars! :)

svekars (approved):

PR #2541, opened 2023-09-01
[ONNX 2] Add ONNX tutorial using torch.onnx.dynamo_export API

thiagocrepaldi:
This PR adds a tutorial that demonstrates how to use `torch.onnx.dynamo_export` to export a PyTorch model to ONNX using the latest TorchDynamo backend

It also refactors an existing ONNX Exporter tutorial to let users know that it uses the now to-be-deprecated ONNX exporter API

cc @BowenBao @williamwen42 @msaroufim

FIXES: https://github.com/pytorch/tutorials/issues/2543

PR #2550, opened 2023-09-06
[ONNX 1] Add Backend section to the left page of Tutorials layout page

thiagocrepaldi:
This PR adds a `Backends` section to the left menu which initially will have a placeholder for the series of ONNX export tutorials based on Torch Dynamo backend
    
A table of content tree page will have static URL `https://pytorch.org/tutorials/beginner/onnx/intro_onnx.html` that will be referenced on PyTorch user document for `torch.onnx module`

Following is a screenshot of the resulting page

Preview page: https://docs-preview.pytorch.org/pytorch/tutorials/2550/beginner/onnx/intro_onnx.html

Fixes: https://github.com/pytorch/tutorials/issues/2552

### DIFF
--- a/.jenkins/validate_tutorials_built.py
+++ b/.jenkins/validate_tutorials_built.py
@@ -10,6 +10,7 @@ REPO_ROOT = Path(__file__).parent.parent
 
 NOT_RUN = [
     "beginner_source/basics/intro",  # no code
+    "beginner_source/onnx/intro_onnx",
     "beginner_source/translation_transformer",
     "beginner_source/profiler",
     "beginner_source/saving_loading_models",
--- a/beginner_source/onnx/intro_onnx.py
+++ b/beginner_source/onnx/intro_onnx.py
@@ -1,0 +1,50 @@
+"""
+**Introduction to ONNX**
+
+Introduction to ONNX
+====================
+
+Authors:
+`Thiago Crepaldi <https://github.com/thiagocrepaldi>`_,
+
+`Open Neural Network eXchange (ONNX) <https://onnx.ai/>`_ is an open standard
+format for representing machine learning models. The ``torch.onnx`` module provides APIs to
+capture the computation graph from a native PyTorch :class:`torch.nn.Module` model and convert
+it into an `ONNX graph <https://github.com/onnx/onnx/blob/main/docs/IR.md>`_.
+
+The exported model can be consumed by any of the many
+`runtimes that support ONNX <https://onnx.ai/supported-tools.html#deployModel>`_,
+including Microsoft's `ONNX Runtime <https://www.onnxruntime.ai>`_.
+
+.. note::
+    Currently, there are two flavors of ONNX exporter APIs,
+    but this tutorial will focus on the ``torch.onnx.dynamo_export``.
+
+The TorchDynamo engine is leveraged to hook into Python's frame evaluation API and dynamically rewrite its
+bytecode into an  `FX graph <https://pytorch.org/docs/stable/fx.html>`_.
+The resulting FX Graph is polished before it is finally translated into an
+`ONNX graph <https://github.com/onnx/onnx/blob/main/docs/IR.md>`_.
+
+The main advantage of this approach is that the `FX graph <https://pytorch.org/docs/stable/fx.html>`_ is captured using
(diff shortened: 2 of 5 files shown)


thiagocrepaldi:
https://github.com/pytorch/tutorials/pull/2541/ will be added to this page as soon as it can be merged

thiagocrepaldi:
@msaroufim FYI

msaroufim (changes requested):
I'm generally OK with having a `tutorials/onnx` but having onnx be a in the left hand navigation is not something I'm as excited about since I'd then have to also be OK with other compiler vendors doing the same

Is there any way we can roll this content in to the deployment tab?

thiagocrepaldi:
> I'm generally OK with having a `tutorials/onnx` index page but having onnx be a in the left hand navigation is not something I'm as excited about since I'd then have to also be OK with other compiler vendors doing the same
> 
> Is there any way we can roll this content in to the deployment tab?

If we could have submenus under `Deploying PyTorch Models in Production` menu, that probably would help already, but I think this is not a trivial task by quickly going through the code. That is why we proposed a ONNX menu on its own.

We do understand the duality of ONNX, especially after pytorch 2.x series. On one hand, `torch.onnx` is within pytorch project under the github.com/pytorch umbrella, like torchvision, torchtext, and torchaudio. Each of these other pytorch components have a menu on their own. From this point of view, ONNX module as pytorch's component could have a menu on its own. On the other hand, after pt 2.0 the concept of dynamo backends got generalized and vendors started implementing theirs and adding all of them might be tricky. 

Also, because we introduced a new ONNX exporter based on Dynamo (`torch.onnx.dynamo_export`) but we still support the older one based on TorchScript (`torch.onnx.export`), we plan in having tutorials helping on the decision-making process of which one to use, which wouldn't be a "deployment" tutorial per se, but very related. Other related but not exactly deployment might be 1) migration process from the TorchScript to the Dynamo based exporter and 2) working around limitations of the exporter, helping users to fix their model for successful export. There is a whole bunch of ONNX-related tutorials to cover lack of documentation of the previous exporter that we plan that may not fit directly into the "deployment" umbrella and having them sprinkled in different menus would be a bit messy

I hope that helps making our case :)

thiagocrepaldi:
Changed the left menu to `Backends` instead of `ONNX` as per offline discussion with @svekars

msaroufim:
Ping me when the doc renders I still don't see the most recent changes

msaroufim (approved):

msaroufim:
ok just stamped thanks!

svekars:
Thanks! I will do an editorial pass. We will merge all 2.1 tutorials on the day of the release.

beginner_source/onnx/intro_onnx.py
```diff
@@ -0,0 +1,50 @@
...
+format for representing machine learning models. The ``torch.onnx`` module provides APIs to
+capture the computation graph from a native PyTorch :class:`torch.nn.Module` model and convert
+it into an `ONNX graph <https://github.com/onnx/onnx/blob/main/docs/IR.md>`_.
+
+The exported model can be consumed by any of the many
+`runtimes that support ONNX <https://onnx.ai/supported-tools.html#deployModel>`_,
+including Microsoft's `ONNX Runtime <https://www.onnxruntime.ai>`_.
+
+.. Note::
+    Currently there are two flavors of ONNX exporter APIs,
```
svekars, line 20:
```suggestion
    Currently, there are two flavors of ONNX exporter APIs,
```

svekars (approved):
LGTM, just a couple of nits.

thiagocrepaldi:
> LGTM, just a couple of nits.

nits fixed

.jenkins/validate_tutorials_built.py
```diff
@@ -10,6 +10,7 @@
 
 NOT_RUN = [
     "beginner_source/basics/intro",  # no code
+    "beginner_source/onnx/intro_onnx",
```
svekars, line 13:
```suggestion
    "beginner_source/onnx/intro_onnx", # has no usable code and serves as a landing for ONNX tutorials
```

svekars (approved):

PR #2571, opened 2023-09-25
SDPA tutorial requires newer CUDA, so host it on an A10G

janeyx99:
## Description
Should fix the fact that I run into tutorial build issues when the sharding algorithm punts the tutorial to a different worker that has too old of a GPU. 
```
RuntimeError: Found Tesla M60 which is too old to be supported by the triton GPU compiler, which is used as the backend. Triton only supports devices of CUDA Capability >= 7.0, but your device is of CUDA capability 5.2
```

Error logs: https://github.com/pytorch/tutorials/actions/runs/6281857867/job/17060687817

## Checklist


malfet (approved):
A bit ugly, but would work...

janeyx99:
@malfet is there a better solution

PR #2562, opened 2023-09-14
[ONNX 3] Add ONNX registry tutorial

titaiwangms:
Follow up https://github.com/pytorch/tutorials/pull/2541, this PR adds step by step guide to demonstrate end to end solution of how to address unsupported ATen/ONNX operators issue. Please review the last commit.

cc @thiagocrepaldi @abock @justinchuby @bowenbao @wschin

PR #2578, opened 2023-10-03
Add ONNX registry tutorial

titaiwangms:
Reland #2562 

Follow up https://github.com/pytorch/tutorials/pull/2541, this PR adds step by step guide to demonstrate end to end solution of how to address unsupported ATen/ONNX operators issue. Please review the last commit.

NOTE: To avoid merge conflicts with the previous two PRs (ONNX 1 and ONNX 2), this PR builds on top of them like ghstack. You can review the last commit.

cc @thiagocrepaldi @abock @justinchuby @BowenBao @wschin


beginner_source/onnx/onnx_registry_tutorial.py
```diff
@@ -0,0 +1,469 @@
...
+
+import onnxscript  # pip install onnxscript
+print(onnxscript.__version__)
+
+# NOTE: opset18 is the only version of ONNX operators we are
+# using in torch.onnx.dynamo_export for now.
+from onnxscript import opset18
+
+import onnxruntime  # pip install onnxruntime
+print(onnxruntime.__version__)
```
thiagocrepaldi, line 43:
```suggestion
```

I still don't see the value on this check in the middle of a series of tutorials, which share a common installation step for all tutorials.

If any check should be done, we should do it in the common installation step (proposed this change above in another comment)

It is repetitive to have the same dependency being checked for each tutorial again and again

beginner_source/onnx/onnx_registry_tutorial.py
```diff
@@ -0,0 +1,469 @@
...
+
+######################################################################
+# Unsupported ATen operators
+# ---------------------------------
+#
+# ATen operators are implemented by PyTorch, and the ONNX exporter team must manually implement the
+# conversion from ATen operators to ONNX operators through [ONNX Script](https://onnxscript.ai/). Although the ONNX exporter
+# team has been making their best efforts to support as many ATen operators as possible, some ATen
+# operators are still not supported. In this section, we will demonstrate how you can implement any
+# unsupported ATen operators, which can contribute back to the project through the PyTorch GitHub repo.
```
thiagocrepaldi, line 54:
```suggestion
# Although the ONNX exporter team has been making their best efforts to support all ATen operators, some
# operators might still not be supported yet. 
#
# In this section, we will demonstrate how you can implement missing ATen operators,
# which we encourage to be contributed back to the project through the PyTorch GitHub repo.
```

Moved part of the text to the introduction. that was good stuff to motivate users to keep reading the 3 possible scenarios later on

beginner_source/onnx/onnx_registry_tutorial.py
```diff
@@ -0,0 +1,469 @@
...
+#
+# If the model cannot be exported to ONNX, for instance, :class:`aten::add.Tensor` is not supported
+# by ONNX The error message can be found, and is as follows (for example,  ``aten::add.Tensor``):
+#    ``RuntimeErrorWithDiagnostic: Unsupported FX nodes: {'call_function': ['aten.add.Tensor']}. ``
```
thiagocrepaldi, line 58:
```suggestion
# In this section, we will assume that `aten::add.Tensor` is not supported by the ONNX registry, and we will demonstrate how to support it.
# When a model cannot be exported to ONNX due to a missing operator, the ONNX exporter will show an error message with something similar to:
#
# .. code-block:: python
#    RuntimeErrorWithDiagnostic: Unsupported FX nodes: {'call_function': ['aten.add.Tensor']}.
```

beginner_source/onnx/onnx_registry_tutorial.py
```diff
@@ -0,0 +1,469 @@
...
+#
+# To support unsupported ATen operators, we need two things:
```
thiagocrepaldi, line 60:
```suggestion
# To support unsupported ATen operators, we need:
```

beginner_source/onnx/onnx_registry_tutorial.py
```diff
@@ -0,0 +1,469 @@
...
+      {onnx_registry.is_registered_op(namespace='aten', op_name='add', overload='Tensor')}")
+
+
+######################################################################
+# In this example, we will assume that ``aten::add.Tensor`` is not supported by the ONNX registry,
+# and we will demonstrate how to support it. The ONNX registry allows user overrides for operator
+# registration. In this case, we will override the registration of ``aten::add.Tensor`` with our
+# implementation and verify it. However, this unsupported operator should return False when
+# checked with :meth:`onnx_registry.is_registered_op`.
+#
```
thiagocrepaldi, line 86:
```suggestion
# The ONNX registry allows user overrides for operator
# registration. In this case, we will override the registration of ``aten::add.Tensor`` with our
# implementation and verify it. However, this unsupported operator should return False when
# checked with :meth:`onnx_registry.is_registered_op`.
#
```

Moved part of the text to the beginning of the section

beginner_source/onnx/onnx_registry_tutorial.py
thiagocrepaldi: This section is focussed on the adding missing aten operators, but the example actually replaces an existing one, which is the topic of the next section.

I think we should add support to a truly missing operator? Maybe a variant of an existing operator that we don't really care about in having in the official registry?

I am trying to prevent confusion between the 3 scenarios. The missing operator section is using a trick of replacing existing operator, which beginners might find confusing

beginner_source/onnx/onnx_registry_tutorial.py
```diff
@@ -0,0 +1,469 @@
...
+aten_gelu_model = CustomGelu()
+input_gelu_x = torch.randn(3, 3)
+
+export_output = torch.onnx.dynamo_export(
+    aten_gelu_model, input_gelu_x, export_options=export_options
+    )
+
+
+######################################################################
+# Make sure the model uses :func:`custom_aten_gelu`` instead of
```
thiagocrepaldi, line 231:
```suggestion
# Make sure the model uses :func:`custom_aten_gelu` instead of
```

beginner_source/onnx/export_simple_model_to_onnx_tutorial.py
```diff
@@ -115,7 +116,7 @@ def forward(self, x):
 # ----------------------------------------------
 #
 # Now that we have our model saved in a file, we can visualize it with `Netron <https://github.com/lutzroeder/netron>`_.
-# Netron can either be installed on macos, Linux or Windows computers, or run directly from the browser.
+# Netron can either be installed on macOS, Linux or Windows computers, or run directly from the browser.
```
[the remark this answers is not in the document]
titaiwangms: Then I will leave it.

beginner_source/onnx/intro_onnx.py
```diff
@@ -32,13 +33,21 @@
...
 ------------
 
+PyTorch 2.1.0 or newer is required.
+
 The ONNX exporter depends on extra Python packages:
 
   - `ONNX <https://onnx.ai>`_
   - `ONNX Script <https://onnxscript.ai>`_
 
 They can be installed through `pip <https://pypi.org/project/pip/>`_:
```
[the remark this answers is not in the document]
titaiwangms: I believe this comment is at the wrong place.

beginner_source/onnx/intro_onnx.py
```diff
@@ -32,13 +33,21 @@
...
 
+.. note::
+   This tutorial leverages `onnxscript <https://github.com/microsoft/onnxscript#readme>`__
+   to create custom ONNX operators. onnxscript is a Python library that allows users to
+   create custom ONNX operators in Python. It is a prerequisite learning material for
+   this tutorial. Please make sure you have read the onnxscript tutorial before proceeding.
+
```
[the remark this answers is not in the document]
titaiwangms: This was requested by you in the original one. I am not sure why this is reviewed again. I think it's important to point out where symbolic functions are from now, and it's a note with specifying custom operators. I like the idea putting it in this page.

beginner_source/onnx/onnx_registry_tutorial.py
titaiwangms: I think version is especially important in registry tutorial because:
1. PyTorch calls graph api from onnxscript
2. torchlib depends on onnxscript version
3. onnxruntime supports varies across version

If we move this to intro_onnx.py, there will be only this code (I see no reason why we should coment "the version", instead of utilizing print out.), and I think intro_onnx.py is very good given current information it provides to users.

If we move this to export_simple_model_to_onnx.py, it's weird that we emphasize onnxscript version when you don't use it in that code.

This version check honestly doesn't block any code, and it actually provides users another debug (version), if their code doesn't work like tutorial. I really see no harm to keep it.

beginner_source/onnx/onnx_registry_tutorial.py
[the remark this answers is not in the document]
titaiwangms: Actually, it's the third case that our users value the most, not the first one.

This paragraph is particularly serving scenario 1, and would be too specific for an overview. Instead, I think the intro below pretty much explains what we are going to do/learn in this tutorial.

```python
# This tutorial is an introduction to ONNX registry, which
# empowers users to create their own ONNX registries enabling
# them to address unsupported operators in ONNX.
```

The term "unsupported operators" unbiasly states that users should refer to here when they have "any kind" unsupported operator issue (could be unsupported python op / ATen op / total custom op).

titaiwangms: Mostly as said above.

beginner_source/onnx/onnx_registry_tutorial.py
```diff
@@ -0,0 +1,469 @@
...
+#
+# If the model cannot be exported to ONNX, for instance, :class:`aten::add.Tensor` is not supported
+# by ONNX The error message can be found, and is as follows (for example,  ``aten::add.Tensor``):
+#    ``RuntimeErrorWithDiagnostic: Unsupported FX nodes: {'call_function': ['aten.add.Tensor']}. ``
+#
+# To support unsupported ATen operators, we need two things:
+# * The unsupported ATen operator namespace, operator name, and the
+#    corresponding overload (for example ``<namespace>::<op_name>.<overload> - aten::add.Tensor``),
+#    which can be found in the error message.
+# * The implementation of the operator in `ONNX Script <https://github.com/microsoft/onnxscript>`__.
```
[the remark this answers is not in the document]
titaiwangms: Thasnk, for example is preferred over e.g. https://github.com/pytorch/tutorials/pull/2562#discussion_r1340626476

beginner_source/onnx/onnx_registry_tutorial.py
titaiwangms: As replied above.

beginner_source/onnx/onnx_registry_tutorial.py
```diff
@@ -0,0 +1,469 @@
...
+"""
+
+
+###############################################################################
+# Overview
+# ~~~~~~~~
+#
+# This tutorial is an introduction to ONNX registry, which
+# empowers users to create their own ONNX registries enabling
+# them to address unsupported operators in ONNX.
```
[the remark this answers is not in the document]
titaiwangms: Replied in other comments

svekars:
Please resolve the merge conflicts

beginner_source/onnx/onnx_registry_tutorial.py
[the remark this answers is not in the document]
thiagocrepaldi: I do like part how the tutorial express what is the ultimate goal. This is pretty clear. In fact, most of the text is great in showing what is next.

My suggestion is to smooth out the transition from a happy path scenario ("I exported a model that didnt anything extra") presented on tutorial 2 to advanced scenario requiring knowledge of operator implementation that is needed by tutorial 3. 

So I proposed a modified version of the original text from the "Unsupported ATen operators" section to serve as an additional introduction to ATen operator (maintained by pytorch core), and how ONNX operators (maintained by onnx exporter)  interacts with the ONNX registry.

This is an attempt to build increasing levels of information before we get to coding. We have at least 3 concepts here. ATen operators, ONNX Operators and ONNX Registry. Discussing them briefly helps public users that dont do ONNX exporter every day to build a mental map on how things fit together.

We can iterate on the proposed text, but defining what is an operator and registry before jumping to the 3 scenarios that we can implement operators is a good thing.

beginner_source/onnx/onnx_registry_tutorial.py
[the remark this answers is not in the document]
thiagocrepaldi: The additional paragraph to the tutorial was commented on the previous comment.

There I mentioned the text above was slightly modified to cover not only unsupported operators, but also replacing existing operators. It wasn't a copy and paste change only. it can still be improved as discussed in the original comment

thiagocrepaldi: replied above too

beginner_source/onnx/onnx_registry_tutorial.py
thiagocrepaldi: I do agree 100% with you that we should not miss an aten operator on purpose just for the sake of the tutorial. I would never suggest that :)

In fact, I am Ok in keeping the current approach if we cannot find another way that frees user from hacking the registry themselves. In that case, I would create a new section/snippet block just for the un-registration part and make it part of the tutorial. That would 1) teach how to remove an operator from registry and 2) create a "physical" separation between the "adding" new operator from the "removing" the operator in the tutorial.

Nonetheless, I would like to inquire about options that make things simpler for the user.

For example, can we leverage the matching logic to register an equivalent operator with different names(overloads)? Maybe leveraging the `default` overload or another overload the . Can we try to register a `aten::add.Tensor.default` that we don't implement today because of our onnx-fx interpreter can match `aten::add.Tensor.default` to `aten::add.Tensor` that already exists?

Maybe something similar to what is being discussed at https://github.com/pytorch/pytorch/pull/109966 in which `aten::split(tensor,dim)` already exists but a `aten::split(tensor, dim, drop_remainder=False)` is added with the same behavior? #109966 could be improved to make this kind of scenario possible, if not already

beginner_source/onnx/README.txt
```diff
@@ -0,0 +1,14 @@
...
+1. intro_onnx.py
+    Introduction to ONNX
+    https://pytorch.org/tutorials/onnx/intro_onnx.html
+
+2. export_simple_model_to_onnx_tutorial.py
+    Export a PyTorch model to ONNX
+    https://pytorch.org/tutorials/beginner/onnx/export_simple_model_to_onnx_tutorial.html
+
+3. onnx_registry_tutorial.py
+    Introduction to ONNX
```
BowenBao, line 13:
```suggestion
    Introduction to ONNX Registry
```
Should it be like this? What does this format mean though, is it like this?
```
<number>. <file>
    <title>
    <link>
```

thiagocrepaldi: good catch.

It serves as table of content

beginner_source/onnx/intro_onnx.py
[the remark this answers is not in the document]
titaiwangms: Done

beginner_source/onnx/onnx_registry_tutorial.py
titaiwangms: Thanks, I am keeping this for now. We can have more detailed discussion on how to present version check when I am back.

beginner_source/onnx/onnx_registry_tutorial.py
titaiwangms: > For example, can we leverage the matching logic to register an equivalent operator with different names(overloads)? Maybe leveraging the default overload or another overload the . Can we try to register a aten::add.Tensor.default that we don't implement today because of our onnx-fx interpreter can match aten::add.Tensor.default to aten::add.Tensor that already exists?

In this approach, we would have to create that ATen overload ahead in torch side to demonstrate it, which is over complicated the three examples here. Otherwise, we never get the overload from FX.

> Maybe something similar to what is being discussed at https://github.com/pytorch/pytorch/pull/109966 in which aten::split(tensor,dim) already exists but a aten::split(tensor, dim, drop_remainder=False) is added with the same behavior? #109966 could be improved to make this kind of scenario possible, if not already

In that case, user will not be able to overwrite, because schema doesn't match. You have to remove the overload first, and register it with the new schema. Schema is a primary key to the whole matching, kwargs is part of them. 

Overall, I think this tutorial is full of new things, and we should expose overload usages later. Maybe after having some feedback on registry, and then we can go from there.

beginner_source/onnx/onnx_registry_tutorial.py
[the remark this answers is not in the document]
titaiwangms: Replied in another comment

beginner_source/onnx/onnx_registry_tutorial.py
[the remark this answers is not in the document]
titaiwangms: I see. Added the line.

beginner_source/onnx/onnx_registry_tutorial.py
titaiwangms: Moved and added ir.html

beginner_source/onnx/intro_onnx.py
```diff
@@ -0,0 +1,68 @@
...
+bytecode analysis that preserves the dynamic nature of the model instead of using traditional static tracing techniques.
+
+Dependencies
+------------
+
+PyTorch 2.1.0 or newer is required.
+
+The ONNX exporter depends on extra Python packages:
+
+  - `ONNX <https://onnx.ai>`_
```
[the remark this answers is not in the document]
titaiwangms: Not sure about this. But added.

titaiwangms:
> Please resolve the merge conflicts

@svekars Thanks, it's ready to go!

titaiwangms:
@thiagocrepaldi 

We can have more discussion and a follow up PR on exposing "ONNX varaints" when I am back.

svekars:
Closing this issue as the author is out of office. A new PR will be created and preserve the original ownership.

PR #2592, opened 2023-10-05
Use torch.export.Dim in export tutorial instead of torch.export.dynamic_dim

williamwen42:
As title. Will not work for 2.1 release since `torch.export.Dim` is not in 2.1.

So I don't think PR can't be merged until the next PR, but I'm getting it out early.

### DIFF
--- a/intermediate_source/torch_export_tutorial.py
+++ b/intermediate_source/torch_export_tutorial.py
@@ -104,6 +104,48 @@ print(exported_mod.graph_module)
 print(exported_mod.graph_signature)
 
 ######################################################################
+The graph produced by ``torch.export`` by default returns a graph containing
+only functional ATen operators. This operator set (or "opset") is around 2000 operators, and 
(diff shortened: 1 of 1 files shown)


intermediate_source/torch_export_tutorial.py
```diff
@@ -255,91 +260,107 @@ def forward(self, x, y):
...
+# We can relax this constraint using the ``dynamic_shapes`` argument of
+# ``torch.export.export()``, which allows us to specify (using ``torch.export.Dim``)
+# which dimensions of the input tensors are dynamic.
 #
 # ..
-#     [TODO] link to doc of dynamic_dim when it is available
+#     [TODO] link to doc of Dim when it is available
+#
+# For each tensor argument of the input callable, we can specify a mapping from the dimension
+# to a ``torch.export.Dim``.
```
angelayi, line 271: https://pytorch.org/docs/main/export.html#torch.export.Dim

PR #2594, opened 2023-10-05
Pin torchrl deps to 0.2.0

malfet:
As latests seems to be incompatible with latest tensordict release, see https://github.com/pytorch/rl/issues/1606

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,7 +25,7 @@ tensorboard
 jinja2==3.0.3
 pytorch-lightning
 torchx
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -25,7 +25,7 @@ tensorboard
 jinja2==3.0.3
 pytorch-lightning
 torchx
-torchrl
+torchrl==0.1.1
```
svekars, line 28:
```suggestion
torchrl==0.2.0
```

PR #2596, opened 2023-10-06
Fix eval compile timing in torch.compile tutorial

williamwen42:
Fixes https://github.com/pytorch/pytorch/issues/110666.

Moved placement of `torch.no_grad`. Added asserts for speedup values so that future regressions are detected sooner.

### DIFF
--- a/intermediate_source/torch_compile_tutorial.py
+++ b/intermediate_source/torch_compile_tutorial.py
@@ -139,8 +139,7 @@ def init_model():
 # ``mode`` argument, which we will discuss below.
 
 def evaluate(mod, inp):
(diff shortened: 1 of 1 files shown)


malfet (approved):

malfet:
Looks like the fix moves `torch.nograd` outside of `torch.compile`, which is OK if it's documented somewhere, that this is nto a supported way of using `torch.compile` (which feels wrong to me, it should just optimize given function, i.e. respect `torch.nograd` decorator)

williamwen42:
Let's push the patch for the tutorial for now, we will fix the no_grad issue in the background.

PR #2608, opened 2023-10-16
[quant] Add IR related recommendations for Quantizer tutorial

jerryzh168:
Summary:
att

Test Plan:
CI

Reviewers:

Subscribers:

Tasks:

Tags:

Fixes #ISSUE_NUMBER

## Checklist

### DIFF
--- a/prototype_source/pt2e_quantizer.rst
+++ b/prototype_source/pt2e_quantizer.rst
@@ -302,6 +302,106 @@ functions that are used in the example:
    `get_bias_qspec <https://github.com/pytorch/pytorch/blob/47cfcf566ab76573452787335f10c9ca185752dc/torch/ao/quantization/_pt2e/quantizer/utils.py#L53>`__
    can be used to get the ``QuantizationSpec`` from ``QuantizationConfig`` for a specific pattern.
 
+A Note on IR for PT2E Quantization Flow
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+IR means the intermediate representation of the model, for example, ``torch`` IR (``torch.nn`` modules, ``torch.nn.functional`` ops) or ``aten`` IR (``torch.ops.aten.linear``, ...). PT2E Quantization Flow is using pre autograd aten IR (the output of `torch.export` API) so that we support training. As is shown before, we need to match the operator or operator patterns before we can do annotations on them, So the question is how do we match the pattern?
+
+1. Matching ``aten`` IR directly
+------------------------------------
+
+The most straightforward way might be matching ``aten`` IR directly.
+
+Example::
+
+  for n in gm.graph.nodes:
+        if n.op != "call_function" or n.target not in [
+            torch.ops.aten.relu.default,
+            torch.ops.aten.relu_.default,
+        ]:
+            continue
+        relu_node = n
+        maybe_conv_node = n.args[0]
+        if (
+            not isinstance(maybe_conv_node, Node)
+            or maybe_conv_node.op != "call_function"
+            or maybe_conv_node.target
+            not in [
+                torch.ops.aten.conv1d.default,
+                torch.ops.aten.conv2d.default,
+            ]
+        ):
+            continue
+
+        # annotate conv and relu nodes
+        ...
+
+However one problem for using this IR is that the representation might change if the PyTorch implementation for modules or functional ops changed. But this could be unexpected since modeling users typically assume that when the eager mode model code doesn't change, they should get the same model representation after program capture as well. One concrete effect for this problem is that if a ``Quantizer`` do annotations based on recognizing ``aten`` IR patterns, then it may fail to recognzing the pattern after PyTorch version update, and the same eager mode floating point may be left unquantized.
+
+2. Using ``SubgraphMatcher``
+--------------------------------
+Because of this, we recommend people to recognize the pattern through ``SubgraphMatcher``, through capturing a ``torch`` IR pattern (with the same program capture used for capturing the floating point model), instead of using the ``aten`` IR pattern directly.
+
+Example::
+
+  def conv_relu_pattern(x, weight, bias):
+      conv = torch.nn.functional.conv2d(x, weight, bias)
+      relu = torch.nn.functional.relu(conv)
+      return relu
+
+  matcher = SubgraphMatcher(conv_relu_pattern)
+  matches = matcher.match(model)
+  for match in matches:
+      # find input and output of the pattern
+      # annotate the nodes
+      inputs, output = _find_input_and_output(match)
+      inputs[0].users[0].meta["quantization_annotation"] = ...
+      inputs[1].users[0].meta["quantization_annotation"] = ...
+      output.meta["quantization_annotation"] = ...
+
+With this, the ``Quantizer`` will still be valid even when the implementation for nn modules and functionals changes, the ``aten`` IR for floating point model will change, but since we capture the pattern again instead of hardcoding the ``aten`` IR for the pattern, we'll get the updated ``aten`` IR as well and will still be able to match the pattern.
+
+One caveat is that if inputs of the pattern has multiple users, we don't have a good way to identify which user node we want to annotate except for checking the aten op target.
+
+Another caveat is that we need to make sure we have an exhaustive list of examples (e.g. 2D, 3D, 4D inputs, real v.s. symbolic inputs, training=True v.s. training=False etc.) for the pattern to make sure cover different possible ``aten`` IR outcomes captured from the ``torch`` IR pattern.
+      
+3. Using ``SubgraphMatcherWithNameNodeMap``
+----------------------------------------------
+We also introduced a different SubgraphMatcher util called ``SubgraphMatcherWithNameNodeMap`` to make it easier to query the nodes that people want to annotate.
+
+Example::
+  
+  def conv_relu_pattern(x, weight, bias):
+      conv = torch.nn.functional.conv2d(x, weight, bias)
+      relu = torch.nn.functional.relu(conv)
+      # returns an additional dict that includes a map from name to node that we want to annotate
+      return relu, {"conv": conv, "relu": relu, "x": x, "weight": weight, "bias": bias}
+
+  matcher = SubgraphMatcherWithNameNodeMap(conv_relu_pattern)
+  matches = matcher.match(model)
+  for match in matches:
+      # find input and output of the pattern
+      # annotate the nodes
+      name_node_map = match.name_node_map
+      name_node_map["conv"].meta["quantization_annotation"] = ...
+      name_node_map["relu"].meta["quantization_annotation"] = ...
+
+This should be easier to use than the original ``SubgraphMatcher``, however, it may not work if some of the operators are captured as multiple ops, e.g. if ``torch.nn.functional.conv2d`` is captured as multiple ops, then the inputs ``x``, ``weight``, ``bias`` may not be the direct inputs to the final operator marked by ``conv``, in this case we'll need to revert to the same way of access mentioned in the previous example.
+
+Example::
+
+  for match in matches:
+      # find input and output of the pattern
+      # annotate the nodes
+      name_node_map = match.name_node_map
+      input_x = name_node_map["x"]
+      # since input_x may not be the input of name_node_map["conv"]
+      # we'll get the user nodes of this op and annotate that instead
+      input_x.users[0].meta["quantization_annotation"] = ...
+      ...
+
+
+Similar to ``SubgraphMatcher``, we also need to make sure to capture the PyTorch pattern with different example inputs to cover all possible ``aten`` IR variants.      
+
 Conclusion
 ^^^^^^^^^^^^^^^^^^^
 
--- a/prototype_source/pt2e_quantizer.rst
+++ b/prototype_source/pt2e_quantizer.rst
@@ -304,7 +304,7 @@ functions that are used in the example:
 
 A Note on IR for PT2E Quantization Flow
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-IR means the intermediate representation of the model, for example, ``torch`` IR (``torch.nn`` modules, ``torch.nn.functional`` ops) or ``aten`` IR (``torch.ops.aten.linear``, ...). PT2E Quantization Flow is using pre autograd aten IR (the output of `torch.export` API) so that we support training. As is shown before, we need to match the operator or operator patterns before we can do annotations on them, So the question is how do we match the pattern?
+IR means the intermediate representation of the model, for example, ``torch`` IR (``torch.nn`` modules, ``torch.nn.functional`` ops) or ``aten`` IR (``torch.ops.aten.linear``, ...). PT2E Quantization Flow is using pre autograd aten IR (the output of `torch.export` API) so that we support training. As is shown before, we need to match the operator or operator patterns before we can attach annotations on them, So the question is how do we match the pattern?
 
 1. Matching ``aten`` IR directly
 ------------------------------------
@@ -363,7 +363,7 @@ With this, the ``Quantizer`` will still be valid even when the implementation fo
 One caveat is that if inputs of the pattern has multiple users, we don't have a good way to identify which user node we want to annotate except for checking the aten op target.
 
 Another caveat is that we need to make sure we have an exhaustive list of examples (e.g. 2D, 3D, 4D inputs, real v.s. symbolic inputs, training=True v.s. training=False etc.) for the pattern to make sure cover different possible ``aten`` IR outcomes captured from the ``torch`` IR pattern.
-      
+
 3. Using ``SubgraphMatcherWithNameNodeMap``
 ----------------------------------------------
 We also introduced a different SubgraphMatcher util called ``SubgraphMatcherWithNameNodeMap`` to make it easier to query the nodes that people want to annotate.
@@ -400,7 +400,15 @@ Example::
       ...
 
 
-Similar to ``SubgraphMatcher``, we also need to make sure to capture the PyTorch pattern with different example inputs to cover all possible ``aten`` IR variants.      
+Similar to ``SubgraphMatcher``, we also need to make sure to capture the PyTorch pattern with different example inputs to cover all possible ``aten`` IR variants.
+
+4. General Recommentation
+----------------------------------------------
+Given the above UX and constraints/caveats, we recommend people to start with option 3. and make sure the first operator in the pattern are functional ops that map to one aten op (e.g. F.conv2d, F.linear), and provide enough variants of example inputs. (We may provide some (pattern, list of example_inputs) so people can just use them directly in the future)
+
+If people are uncertain if some functional operator is going to be traced into multiple aten operators then they can pick option 2.
+
+We would not recommend option 1. since that's the least stable in all three options.
 
 Conclusion
 ^^^^^^^^^^^^^^^^^^^
--- a/prototype_source/pt2e_quantizer.rst
+++ b/prototype_source/pt2e_quantizer.rst
@@ -385,7 +385,7 @@ Example::
       name_node_map["conv"].meta["quantization_annotation"] = ...
       name_node_map["relu"].meta["quantization_annotation"] = ...
 
-This should be easier to use than the original ``SubgraphMatcher``, however, it may not work if some of the operators are captured as multiple ops, e.g. if ``torch.nn.functional.conv2d`` is captured as multiple ops, then the inputs ``x``, ``weight``, ``bias`` may not be the direct inputs to the final operator marked by ``conv``, in this case we'll need to revert to the same way of access mentioned in the previous example.
+This should be easier to use than the original ``SubgraphMatcher`` around finding the node we want to annotate or use. However, it may not work if some of the operators are captured as multiple ops. For example, if ``torch.nn.functional.conv2d`` is captured as multiple ops, then the inputs ``x``, ``weight``, ``bias`` may not be the direct inputs to the final operator marked by ``conv``, in this case we'll need to revert to the same way of access mentioned in the previous example.
 
 Example::
 
--- a/prototype_source/pt2e_quantizer.rst
+++ b/prototype_source/pt2e_quantizer.rst
@@ -306,8 +306,8 @@ A Note on IR for PT2E Quantization Flow
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 IR means the intermediate representation of the model, for example, ``torch`` IR (``torch.nn`` modules, ``torch.nn.functional`` ops) or ``aten`` IR (``torch.ops.aten.linear``, ...). PT2E Quantization Flow is using pre autograd aten IR (the output of `torch.export` API) so that we support training. As is shown before, we need to match the operator or operator patterns before we can attach annotations on them, So the question is how do we match the pattern?
 
-1. Matching ``aten`` IR directly
-------------------------------------
+Motivation: Problem of Matching ``aten`` IR directly
+--------------------------------------------------------
 
 The most straightforward way might be matching ``aten`` IR directly.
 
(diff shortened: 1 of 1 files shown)


prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
...
+                torch.ops.aten.conv1d.default,
+                torch.ops.aten.conv2d.default,
+            ]
+        ):
+            continue
+
+        # annotate conv and relu nodes
+        ...
+
+However one problem for using this IR is that the representation might change if the PyTorch implementation for modules or functional ops changed. But this could be unexpected since modeling users typically assume that when the eager mode model code doesn't change, they should get the same model representation after program capture as well. One concrete effect for this problem is that if a ``Quantizer`` do annotations based on recognizing ``aten`` IR patterns, then it may fail to recognzing the pattern after PyTorch version update, and the same eager mode floating point may be left unquantized.
```
svekars, line 338:
```suggestion
However, an issue with this IR is that the representation might change if the PyTorch implementation for modules or functional ops changed. But this could be unexpected since modeling users typically assume that when the eager mode model code doesn't change, they should get the same model representation after program capture as well. One concrete effect for this problem is that if a ``Quantizer`` do annotations based on recognizing ``aten`` IR patterns, then it may fail to recognize the pattern after PyTorch version update, and the same eager mode floating point may be left unquantized.
```

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
...
+
+2. Using ``SubgraphMatcher``
+--------------------------------
+Because of this, we recommend people to recognize the pattern through ``SubgraphMatcher``, through capturing a ``torch`` IR pattern (with the same program capture used for capturing the floating point model), instead of using the ``aten`` IR pattern directly.
```
svekars, line 342:
```suggestion
Because of this, we recommend to recognize the pattern through ``SubgraphMatcher``, through capturing a ``torch`` IR pattern (with the same program capture used for capturing the floating point model), instead of using the ``aten`` IR pattern directly.
```

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
...
+
+Example::
```
svekars, line 344:
```suggestion
**Example:**::
```

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
...
+  matches = matcher.match(model)
+  for match in matches:
+      # find input and output of the pattern
+      # annotate the nodes
+      inputs, output = _find_input_and_output(match)
+      inputs[0].users[0].meta["quantization_annotation"] = ...
+      inputs[1].users[0].meta["quantization_annotation"] = ...
+      output.meta["quantization_annotation"] = ...
+
+With this, the ``Quantizer`` will still be valid even when the implementation for nn modules and functionals changes, the ``aten`` IR for floating point model will change, but since we capture the pattern again instead of hardcoding the ``aten`` IR for the pattern, we'll get the updated ``aten`` IR as well and will still be able to match the pattern.
```
svekars, line 361:
```suggestion
With this, the ``Quantizer`` will still be valid even when the implementation for the ``nn`` modules and functionals changes, the ``aten`` IR for floating point model will change, but since we capture the pattern again instead of hardcoding the ``aten`` IR for the pattern, we'll get the updated ``aten`` IR as well and will still be able to match the pattern.
```

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
...
+
+One caveat is that if inputs of the pattern has multiple users, we don't have a good way to identify which user node we want to annotate except for checking the aten op target.
```
svekars, line 363:
```suggestion
One caveat is that if inputs of the pattern has multiple users, we don't have a good way to identify which user node we want to annotate except for checking the ``aten`` op target.
```

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
...
+
+Another caveat is that we need to make sure we have an exhaustive list of examples (e.g. 2D, 3D, 4D inputs, real v.s. symbolic inputs, training=True v.s. training=False etc.) for the pattern to make sure cover different possible ``aten`` IR outcomes captured from the ``torch`` IR pattern.
```
svekars, line 365:
```suggestion
Another caveat is that we need to make sure we have an exhaustive list of examples (such as 2D, 3D, 4D inputs, real vs. symbolic inputs, ``training=True`` vs. ``training=False``, and so on) for the pattern to make sure cover different possible ``aten`` IR outcomes captured from the ``torch`` IR pattern.
```

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
...
+      
+3. Using ``SubgraphMatcherWithNameNodeMap``
+----------------------------------------------
+We also introduced a different SubgraphMatcher util called ``SubgraphMatcherWithNameNodeMap`` to make it easier to query the nodes that people want to annotate.
```
svekars, line 369:
```suggestion
We also introduced a different ``SubgraphMatcher`` utility called ``SubgraphMatcherWithNameNodeMap`` to make it easier to query the nodes that people want to annotate.
```

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
...
+  for match in matches:
+      # find input and output of the pattern
+      # annotate the nodes
+      name_node_map = match.name_node_map
+      name_node_map["conv"].meta["quantization_annotation"] = ...
+      name_node_map["relu"].meta["quantization_annotation"] = ...
+
+This should be easier to use than the original ``SubgraphMatcher`` around finding the node we want to annotate or use. However, it may not work if some of the operators are captured as multiple ops. For example, if ``torch.nn.functional.conv2d`` is captured as multiple ops, then the inputs ``x``, ``weight``, ``bias`` may not be the direct inputs to the final operator marked by ``conv``, in this case we'll need to revert to the same way of access mentioned in the previous example.
+
+Example::
```
svekars, line 390:
```suggestion
**Example:**::
```

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
    `get_bias_qspec <https://github.com/pytorch/pytorch/blob/47cfcf566ab76573452787335f10c9ca185752dc/torch/ao/quantization/_pt2e/quantizer/utils.py#L53>`__
    can be used to get the ``QuantizationSpec`` from ``QuantizationConfig`` for a specific pattern.
 
+A Note on IR for PT2E Quantization Flow
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+IR means the intermediate representation of the model, for example, ``torch`` IR (``torch.nn`` modules, ``torch.nn.functional`` ops) or ``aten`` IR (``torch.ops.aten.linear``, ...). PT2E Quantization Flow is using pre autograd aten IR (the output of `torch.export` API) so that we support training. As is shown before, we need to match the operator or operator patterns before we can do annotations on them, So the question is how do we match the pattern?
```
svekars, line 307:
```suggestion
**IR** stands for the **Intermediate Representation** of the model. For example, ``torch`` IR (``torch.nn`` modules, ``torch.nn.functional`` ops) or ``aten`` IR (``torch.ops.aten.linear``, ...). PT2E Quantization Flow is using pre autograd ATen IR (the output of ``torch.export`` API) so that we support training. As is shown before, we need to match the operator or operator patterns before we can do annotations on them. The question is, how do we match the pattern?
```

svekars (commented):
Some editorial nits.

svekars:
Can you please fix the pyspelling job?

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
...
+            continue
+
+        # annotate conv and relu nodes
+        ...
+
+However, an issue with this IR is that the representation might change if the PyTorch implementation for modules or functional ops changed. But this could be unexpected since modeling users typically assume that when the eager mode model code doesn't change, they should get the same model representation after program capture as well. One concrete effect for this problem is that if a ``Quantizer`` do annotations based on recognizing ``aten`` IR patterns, then it may fail to recognize the pattern after PyTorch version update, and the same eager mode floating point may be left unquantized.
+
+2. Using ``SubgraphMatcher``
+--------------------------------
+Because of this, we recommend to recognize the pattern through ``SubgraphMatcher``, through capturing a ``torch`` IR pattern (with the same program capture used for capturing the floating point model), instead of using the ``aten`` IR pattern directly.
```
andrewor14, line 342: I feel it makes sense to start with `SubgraphMatcher` since this is the recommended way?

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
...
+      conv = torch.nn.functional.conv2d(x, weight, bias)
+      relu = torch.nn.functional.relu(conv)
+      return relu
+
+  matcher = SubgraphMatcher(conv_relu_pattern)
+  matches = matcher.match(model)
+  for match in matches:
+      # find input and output of the pattern
+      # annotate the nodes
+      inputs, output = _find_input_and_output(match)
```
andrewor14, line 356: is this util something we provide?

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
...
+  for match in matches:
+      # find input and output of the pattern
+      # annotate the nodes
+      name_node_map = match.name_node_map
+      input_x = name_node_map["x"]
+      # since input_x may not be the input of name_node_map["conv"]
+      # we'll get the user nodes of this op and annotate that instead
+      input_x.users[0].meta["quantization_annotation"] = ...
+      ...
+
```
andrewor14, line 401: Should we add (4) using `find_sequential_partitions`? I feel a lot of people will try to look at our `XNNPACKQuantizer` example when writing their own quantizers. It may be confusing to see a way to implement quantizers that are not described in the tutorials.

Alternatively, do we have plans to rewrite `XNNPACKQuantizer` to use either (2) or (3)?

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
...
+----------------------------------------------
+We also introduced a different ``SubgraphMatcher`` utility called ``SubgraphMatcherWithNameNodeMap`` to make it easier to query the nodes that people want to annotate.
+
+Example::
+  
+  def conv_relu_pattern(x, weight, bias):
+      conv = torch.nn.functional.conv2d(x, weight, bias)
+      relu = torch.nn.functional.relu(conv)
+      # returns an additional dict that includes a map from name to node that we want to annotate
+      return relu, {"conv": conv, "relu": relu, "x": x, "weight": weight, "bias": bias}
```
andrewor14, line 377: Did we resolve the weight issue? Can we return an `nn.Parameter` here?

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,106 @@ functions that are used in the example:
...
+      return relu
+
+  matcher = SubgraphMatcher(conv_relu_pattern)
+  matches = matcher.match(model)
+  for match in matches:
+      # find input and output of the pattern
+      # annotate the nodes
+      inputs, output = _find_input_and_output(match)
+      inputs[0].users[0].meta["quantization_annotation"] = ...
+      inputs[1].users[0].meta["quantization_annotation"] = ...
```
andrewor14, line 358: Actually from reading the example (2) doesn't seem super user friendly. Should we put (3) `SubgraphMatcherWithNameNodeMap` at the top as the recommended way? I feel it'll handle 90% of the patterns we want to match. For something like LSTM or MHA maybe we need something more advanced like (1) and (2)

andrewor14 (approved):
Looks good. Just a few questions about the ordering / how this relates to our existing `XNNPACKQuantizer`.

prototype_source/pt2e_quantizer.rst
jerryzh168: yeah I feel it's better to rewrite XNNPACKQuantizer to use (2) probably since (2) looks at graph instead of relying on metadata, cc @kimishpatel let me know what you think.

prototype_source/pt2e_quantizer.rst
jerryzh168: I'm planning to add a paragraph for recommendation, this is just the order we introduce them

prototype_source/pt2e_quantizer.rst
jerryzh168: yeah I think we should probably provide some API here, I haven't decided on the concrete util names/signatures yet though, we can update this a bit later

prototype_source/pt2e_quantizer.rst
jerryzh168: not yet, nn.Parameter is not supported. also I haven't tested if this would work for linear modules yet actually.

prototype_source/pt2e_quantizer.rst
jerryzh168: yeah I was thinking of having (3) as the recommended one, but there is actually an caveat that we are assuming some ops not being traced into multiple ops (e.g. F.linear, F.conv2d etc.)

jerryzh168:
> Can you please fix the pyspelling job?

@svekars actually I'm not sure how to fix these actually: https://github.com/pytorch/tutorials/actions/runs/6539478971/job/17757612269?pr=2608

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,114 @@ functions that are used in the example:
...
+
+Example::
+  
+  def conv_relu_pattern(x, weight, bias):
+      conv = torch.nn.functional.conv2d(x, weight, bias)
+      relu = torch.nn.functional.relu(conv)
+      # returns an additional dict that includes a map from name to node that we want to annotate
+      return relu, {"conv": conv, "relu": relu, "x": x, "weight": weight, "bias": bias}
+
+  matcher = SubgraphMatcherWithNameNodeMap(conv_relu_pattern)
```
kimishpatel, line 379: can we actually remove 3? This is just adding confusion because I dont have any realistic scenario where this will be useful. When you want to "annotate" intermediate nodes, meaning "mark" them as "annotated", then that can be done via nodes returned via subgraph matcher anyway. So honestly this is just adding confusion.

prototype_source/pt2e_quantizer.rst
kimishpatel: I need to think if `find_sequential_partitions`, if its implementation is changed, will be useful to find some arbitrary sequential patterns. For now, it is not clear, so lets not add more confusion here.

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,114 @@ functions that are used in the example:
...
+      # we'll get the user nodes of this op and annotate that instead
+      input_x.users[0].meta["quantization_annotation"] = ...
+      ...
+
+
+Similar to ``SubgraphMatcher``, we also need to make sure to capture the PyTorch pattern with different example inputs to cover all possible ``aten`` IR variants.
+
+4. General Recommentation
+----------------------------------------------
+Given the above UX and constraints/caveats, we recommend people to start with option 3. and make sure the first operator in the pattern are functional ops that map to one aten op (e.g. F.conv2d, F.linear), and provide enough variants of example inputs. (We may provide some (pattern, list of example_inputs) so people can just use them directly in the future)
```
kimishpatel, line 407: I would really really stay away from "recommending". People implementing Quantizer should make their choice. When you "recommend" it feels like you are implicitly supporting something, and when that doesnt work, "people" will come to you complaining.

We should just highlight different ways of graph matching, that is it. And this is not specific to quantization, so ideally this should have its own user manual somewhere.

kimishpatel (changes requested):
left some comments

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,114 @@ functions that are used in the example:
...
+            continue
+
+        # annotate conv and relu nodes
+        ...
+
+However one problem for using this IR is that the representation might change if the PyTorch implementation for modules or functional ops changed. But this could be unexpected since modeling users typically assume that when the eager mode model code doesn't change, they should get the same model representation after program capture as well. One concrete effect for this problem is that if a ``Quantizer`` do annotations based on recognizing ``aten`` IR patterns, then it may fail to recognzing the pattern after PyTorch version update, and the same eager mode floating point may be left unquantized.
+
+2. Using ``SubgraphMatcher``
+--------------------------------
+Because of this, we recommend people to recognize the pattern through ``SubgraphMatcher``, through capturing a ``torch`` IR pattern (with the same program capture used for capturing the floating point model), instead of using the ``aten`` IR pattern directly.
```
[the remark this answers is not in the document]
jerryzh168: "people" means backend developers.

The motivation for giving a recommendation is introduced in L338, basically trying to make sure the pattern matching is robust against pytorch code changes. If we don't give the recommendation I think users will still expect the same python model + same quantization code + same quantizer will give the same quantized model when they update their pytorch version, which may not be true if quantizer writers are not cautious about this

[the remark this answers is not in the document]
jerryzh168: probably not officially defined, L307 gives some examples

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,114 @@ functions that are used in the example:
...
+      # find input and output of the pattern
+      # annotate the nodes
+      inputs, output = _find_input_and_output(match)
+      inputs[0].users[0].meta["quantization_annotation"] = ...
+      inputs[1].users[0].meta["quantization_annotation"] = ...
+      output.meta["quantization_annotation"] = ...
+
+With this, the ``Quantizer`` will still be valid even when the implementation for nn modules and functionals changes, the ``aten`` IR for floating point model will change, but since we capture the pattern again instead of hardcoding the ``aten`` IR for the pattern, we'll get the updated ``aten`` IR as well and will still be able to match the pattern.
+
+One caveat is that if inputs of the pattern has multiple users, we don't have a good way to identify which user node we want to annotate except for checking the aten op target.
```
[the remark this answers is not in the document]
jerryzh168: I'm assuming L338 is what everyone wants, do you feel this is not true?

prototype_source/pt2e_quantizer.rst
jerryzh168: I think 3 is actually compatible with the uses in 2 as well, and it can give an easier way to access the node, maybe I can remove 2 and talk about how to use 3 to do annotation without referring to internal nodes, it will be something like:

```
input = match["x"]
weight = match["weight"]
input_user = input.users[0]
input_user.meta["..."] = ...
```

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,114 @@ functions that are used in the example:
...
+
+Similar to ``SubgraphMatcher``, we also need to make sure to capture the PyTorch pattern with different example inputs to cover all possible ``aten`` IR variants.
+
+4. General Recommentation
+----------------------------------------------
+Given the above UX and constraints/caveats, we recommend people to start with option 3. and make sure the first operator in the pattern are functional ops that map to one aten op (e.g. F.conv2d, F.linear), and provide enough variants of example inputs. (We may provide some (pattern, list of example_inputs) so people can just use them directly in the future)
+
+If people are uncertain if some functional operator is going to be traced into multiple aten operators then they can pick option 2.
+
+We would not recommend option 1. since that's the least stable in all three options.
```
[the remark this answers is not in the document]
jerryzh168: I don't understand why 1 would be better than torch IR? what is the criteria here?

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,114 @@ functions that are used in the example:
    `get_bias_qspec <https://github.com/pytorch/pytorch/blob/47cfcf566ab76573452787335f10c9ca185752dc/torch/ao/quantization/_pt2e/quantizer/utils.py#L53>`__
    can be used to get the ``QuantizationSpec`` from ``QuantizationConfig`` for a specific pattern.
 
+A Note on IR for PT2E Quantization Flow
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ezyang, line 306: Who is the target audience for this note? Wouldn't I expect them to already know what an IR is?

jerryzh168: this is targeting backend developers, like ARM, Qualcomm, Intel, TensorRT etc. yeah they probably know what an IR is. you mean I can delete the explanation of IR or something? the explanation is also explaining what we mean by torch IR, aten IR etc.

prototype_source/pt2e_quantizer.rst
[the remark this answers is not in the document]
kimishpatel: Thats fair. My main concern is with "recommendation". I would have presented alternatives with pros and cons.

ezyang:
Modulo disagreements with @kimishpatel, the doc seems fine, but you should workshop it with one of the potential users and check if the APIs actually are sufficient for them in practice. User feedback matters.

prototype_source/pt2e_quantizer.rst
kimishpatel: TBH we are giving 2 options neither of which are solving the problem you describe in the code snippet in a good way. If decomposition is a real problem 3 doesnt help. If decomposition is not a problem, I can buy that 3 is slightly better, in which case I would just keep 3 and remove 2. 3 is strictly a superset of 2 anyway and there is not significant difference in UX between 2 and 3.

prototype_source/pt2e_quantizer.rst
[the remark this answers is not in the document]
kimishpatel: i understand the stability argument. My point is that _in practice_ torch IR is very close to aten IR. And if you dont recommend it, I would just remove option 1 and just have this doc talk be "how-to-pattern-match for quantizers"

kimishpatel:
> Modulo disagreements with @kimishpatel, the doc seems fine, but you should workshop it with one of the potential users and check if the APIs actually are sufficient for them in practice. User feedback matters.

lol. Please do tell me where you disagree though

prototype_source/pt2e_quantizer.rst
jerryzh168: > 3 is strictly a superset of 2

yeah agree, maybe I should just remove 2?

we do want to protect against decomposition (one op traced into multiple ops) I think, but for some ops functional op is the same as aten already: https://github.com/pytorch/pytorch/blob/main/torch/nn/functional.py#L2011, unless we expect these will change in the future, we don't necessarily need to worry about them being traced into multiple ops actually.

If we can refer to intermediate node, it's slightly easier to do the annotate I think. But to be more robust, I can also just say always use 3 and always annotate input_users, instead of looking at intermediate node. do you think we should do this then?

prototype_source/pt2e_quantizer.rst
jerryzh168: I do plan to provide some utils actually, since now we separated patterns matching and annotation, we could maintain a centralized place for patterns

still don't quite following the "recommendation" point, but I feel at this point looks like we can just recommend one API

prototype_source/pt2e_quantizer.rst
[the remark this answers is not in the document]
jerryzh168: OK, I'll present this in a different way, this is also a motivation of why we want to have other ways of matching

jerryzh168:
> > Modulo disagreements with @kimishpatel, the doc seems fine, but you should workshop it with one of the potential users and check if the APIs actually are sufficient for them in practice. User feedback matters.
> 
> lol. Please do tell me where you disagree though

I think Ed meant the disagreement between you and me on what options to recommend. I plan to restructure this and use option 1 as motivation and remove option 2

prototype_source/pt2e_quantizer.rst
kimishpatel: > If we can refer to intermediate node, it's slightly easier to do the annotate I think. But to be more robust, I can also just say always use 3 and always annotate input_users, instead of looking at intermediate node. do you think we should do this then?

That sounds good to me.

prototype_source/pt2e_quantizer.rst
kimishpatel: Actually now that I think of it, we should actually do a pattern cache of some sort that can be used not just for quantization but also for delegation.

kimishpatel (approved):

prototype_source/pt2e_quantizer.rst
jerryzh168: yes, Greg also mentioned this before, we can probably capture some of these patterns beforehand and users can use them for matching directly. maybe these could be used for delegation as well if you want to match the pattern and then expand the pattern to cover q/dq ops etc.

prototype_source/pt2e_quantizer.rst
```diff
@@ -302,6 +302,76 @@ functions that are used in the example:
...
+      bias_node = name_node_map["bias"]
+      output_node = name_node_map["relu"]
+      input_node.users[0].meta["quantization_annotation"] = ...
+      weight_node.users[0].meta["quantization_annotation"] = ...
+      bias_node.users[0].meta["quantization_annotation"] = ...
+      output_node.meta["quantization_annotation"] = ...
+
+With this, the ``Quantizer`` will still be valid even when the implementation for nn modules and functionals changes, the ``aten`` IR for floating point model will change, but since we capture the pattern again instead of hardcoding the ``aten`` IR for the pattern, we'll get the updated ``aten`` IR as well and will still be able to match the pattern.
+
+One caveat is that if inputs of the pattern has multiple users, we don't have a good way to identify which user node we want to annotate except for checking the aten op target.
```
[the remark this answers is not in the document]
jerryzh168: yes that's correct

svekars (approved):

PR #2616, opened 2023-10-23
Implemented fixes in 2:4 sparsity tutorial

lvoursl:
## Description
I was running [this tutorial](https://pytorch.org/tutorials/prototype/semi_structured_sparse.html) on 2:4 sparsity and found a few issues that I'm fixing in this PR. More concrete:
1. Block with the model definition wasn't rendered on the website since there was some stop symbol in source markdown. 
2. One variable wasn't defined in the code and one comma was missed in another place -- so some code wasn't working.
3. Model didn't converge to the same level of performance after pruning (gap was substantial in F1 metric). The problem seems to be related to the learning rate schedule in HuggingFace Transformers: during fine-tuning of pruned model learning rate went to 0 super quickly and the model params were not updating at all. To fix it I simply set LR schedule to constant, so we have consistent results in that case. 

cc: @jcaip

### DIFF
--- a/prototype_source/semi_structured_sparse.rst
+++ b/prototype_source/semi_structured_sparse.rst
@@ -315,6 +315,7 @@ Now that those are defined, we just need one additional helper function, which w
 We will get started by loading our model and tokenizer, and then setting up our dataset.
 
 .. code:: python
+
     # load model
     model_name = "bert-base-cased"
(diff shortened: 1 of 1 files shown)


jcaip (approved):
lgtm, thank you for the changes!

PR #2645, opened 2023-11-04
Replace  usage of copy.deepcopy() - Convolution/Batch Norm Fuser in FX

### LINKED ISSUE #2331: Update tutorial to avoid use of copy.deepcopy()- Convolution/Batch Norm Fuser in FX
Update the  [Building a convolution/batch norm fuser in FX](https://github.com/pytorch/tutorials/blob/6ae56a1a2639308550a69c5006e879d331f05b6d/intermediate_source/fx_conv_bn_fuser.py#L107) to avoid the use of copy.deepcopy().
Tutorial link: https://github.com/pytorch/tutorials/blob/6ae56a1a2639308550a69c5006e879d331f05b6d/intermediate_source/fx_conv_bn_fuser.py#L107

See issue #2177  for more information

cc @eellison @suo @gmagogsfm @jamesr66a @msaroufim @SherlockNoMad @albanD @svekars @carljparker @NicolasHug @kit1980 @subramen

MirMustafaAli:
Fixes #2331 

## Description
Replacing the use of `copy.deepcopy()` in _Convolution/Batch Norm Fuser in FX tutorials_   with use of `load_state_dict`   as mentioned in [Deep copying PyTorch modules](https://discuss.pytorch.org/t/deep-copying-pytorch-modules/13514). 

cc @eellison @suo @gmagogsfm @jamesr66a @msaroufim @SherlockNoMad @albanD @sekyondaMeta @svekars @carljparker @NicolasHug @kit1980 @subramen

### DIFF
--- a/intermediate_source/fx_conv_bn_fuser.py
+++ b/intermediate_source/fx_conv_bn_fuser.py
@@ -104,7 +104,9 @@ def fuse_conv_bn_eval(conv, bn):
     module `C` such that C(x) == B(A(x)) in inference mode.
     """
     assert(not (conv.training or bn.training)), "Fusion only for eval!"
-    fused_conv = copy.deepcopy(conv)
+    fused_conv = type(conv)(conv.in_channels, conv.out_channels, conv.kernel_size)
+    fused_conv.load_state_dict(conv.state_dict())
+    fused_conv.eval()
 
     fused_conv.weight, fused_conv.bias = \
         fuse_conv_bn_weights(fused_conv.weight, fused_conv.bias,
--- a/intermediate_source/fx_conv_bn_fuser.py
+++ b/intermediate_source/fx_conv_bn_fuser.py
@@ -150,7 +150,9 @@ def replace_node_module(node: fx.Node, modules: Dict[str, Any], new_module: torc
 
 
 def fuse(model: torch.nn.Module) -> torch.nn.Module:
-    model = copy.deepcopy(model)
+    model, state_dict = type(model)(), model.state_dict()
+    model.load_state_dict(state_dict)
+    model.eval()
     # The first step of most FX passes is to symbolically trace our model to
     # obtain a `GraphModule`. This is a representation of our original model
     # that is functionally identical to our original model, except that we now


intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -104,7 +104,9 @@ def fuse_conv_bn_eval(conv, bn):
     module `C` such that C(x) == B(A(x)) in inference mode.
     """
     assert(not (conv.training or bn.training)), "Fusion only for eval!"
-    fused_conv = copy.deepcopy(conv)
+    fused_conv = type(conv)(conv.in_channels, conv.out_channels, conv.kernel_size)
```
msaroufim, line 107: This fix seems weird? I'd guess the better is to implement a proper `__deepcopy__() for nn modules? @albanD 

This popular thread seems to validate this https://discuss.pytorch.org/t/deep-copying-pytorch-modules/13514 but idk if this is what we want people to actually do?

MirMustafaAli: > This fix seems weird? The right to do feels like its implementing a proper `__deepcopy__()` for nn modules? @albanD
> 
> This popular thread seems to validate this fix https://discuss.pytorch.org/t/deep-copying-pytorch-modules/13514 but idk if this is what we want people to actually do?

we can save and load the model, found from [2385](https://github.com/pytorch/tutorials/pull/2385) . Other than this, is there other way which i am missing that will help me make a plausible fix ?

albanD: The problem is that Module is a complex enough class that deepcopying it is very challenging (the same way we don't recommend you serialize it as-is but only the state_dict).
deepcopy() work in most simple cases but it is expected to fail sometimes.
If you only have a regular Conv2d kernel, doing deepcopy or a new constructor is pretty much the same thing though.

MirMustafaAli: how would you want me to proceed with the PR?

NicolasHug: I think what @albanD is saying is that in this specific case `deepcopy`-ing a conv layer is just fine, i.e. the original code probably doesn't need to be changed.

intermediate_source/fx_conv_bn_fuser.py
```diff
@@ -150,7 +152,9 @@ def replace_node_module(node: fx.Node, modules: Dict[str, Any], new_module: torc
 
 
 def fuse(model: torch.nn.Module) -> torch.nn.Module:
-    model = copy.deepcopy(model)
+    model, state_dict = type(model)(), model.state_dict()
```
NicolasHug, line 155: That's only going to work for models that do not take any parameters to `__init__()`.

@svekars with this + https://github.com/pytorch/tutorials/pull/2645/files#r1391396959, I'm tempted to think that the originally issue is probably irrelevant for this tutorial. Even if `copy.deepcopy(model)` may not be perfect, it's still better than any alternative that has been proposed so far. Perhaps we could close the original issue and still provide credits to the contributor for their efforts?

svekars:
Closing this and the issue and will give half credit.

PR #2674, opened 2023-11-12
Add No Support Label for ROCm GPU in pytorch profiler tutorial 

### LINKED ISSUE #2014: profiler ROCm tracing result was wrong
I try to profile a resnet50 model based on https://pytorch.org/tutorials/intermediate/tensorboard_profiler_tutorial.html.
But the trace result in chrome://tracing is weird.
image
The profiler should record from 0ms but here record from nearly 250 ms. And the active step didn't launch the kernel on GPU.
The schedule was set up to wait=1, warmup=1, active=1 .

cc @aaronenyeshi @chaekit @sekyondaMeta @svekars @carljparker @NicolasHug @kit1980 @subramen @robieta

guptaaryan16:
Fixes #2014

## Description
This PR is in relation to the issue of ROCm support in PyTorch Profiler, which I have seen can give some bugs on testing with PyTorch API. The official files in `PyTorch/Profiler` does not have an option for its support yet as given here: https://github.com/pytorch/pytorch/blob/7f1cbc8b5a7de8794c36833baa47bb1343833589/torch/profiler/profiler.py#L35

ROCm usage can give some unexpected results so right now it's best to mark the usage unsupported. I plan to raise this issue with the PyTorch team as it will be important to support ROCm GPUs as their usage grows.

## Checklist

cc @aaronenyeshi @chaekit @jcarreiro @sekyondaMeta @svekars @carljparker @NicolasHug @kit1980 @subramen

### DIFF
--- a/intermediate_source/tensorboard_profiler_tutorial.py
+++ b/intermediate_source/tensorboard_profiler_tutorial.py
@@ -7,7 +7,7 @@ to detect performance bottlenecks of the model.
 Introduction
 ------------
 PyTorch 1.8 includes an updated profiler API capable of
-recording the CPU side operations as well as the CUDA kernel launches on the GPU side (ROCm AMD GPUs are not supported).
+recording the CPU side operations as well as the CUDA kernel launches on the GPU side (``AMD ROCm™`` GPUs are not supported).
 The profiler can visualize this information
 in TensorBoard Plugin and provide analysis of the performance bottlenecks.
 
@@ -57,10 +57,11 @@ import torchvision.transforms as T
 # Transform it to the desired format and use ``DataLoader`` to load each batch.
 
 transform = T.Compose(
-    [T.Resize(224),
-     T.ToTensor(),
-     T.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))])
-train_set = torchvision.datasets.CIFAR10(root='./data', train=True, download=True, transform=transform)
+    [T.Resize(224), T.ToTensor(), T.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))]
+)
+train_set = torchvision.datasets.CIFAR10(
+    root="./data", train=True, download=True, transform=transform
+)
 train_loader = torch.utils.data.DataLoader(train_set, batch_size=32, shuffle=True)
 
 ######################################################################
@@ -68,7 +69,7 @@ train_loader = torch.utils.data.DataLoader(train_set, batch_size=32, shuffle=Tru
 # To run on GPU, move model and loss to GPU device.
 
 device = torch.device("cuda:0")
-model = torchvision.models.resnet18(weights='IMAGENET1K_V1').cuda(device)
+model = torchvision.models.resnet18(weights="IMAGENET1K_V1").cuda(device)
 criterion = torch.nn.CrossEntropyLoss().cuda(device)
 optimizer = torch.optim.SGD(model.parameters(), lr=0.001, momentum=0.9)
 model.train()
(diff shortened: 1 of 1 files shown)


intermediate_source/tensorboard_profiler_tutorial.py
```diff
@@ -7,7 +7,7 @@
 Introduction
 ------------
 PyTorch 1.8 includes an updated profiler API capable of
-recording the CPU side operations as well as the CUDA kernel launches on the GPU side.
+recording the CPU side operations as well as the CUDA kernel launches on the GPU side (``AMD ROCm™`` GPUs are not supported).
```
malfet, line 10: @jeffdaily is this indeed the case?

malfet: Also, why are you adding a backtricks here?

jeffdaily: ROCm has full upstream support of both modern pytorch profiling via kineto + ROCm's libroctracer, as well as the older autograd profiler via ROCm's roctx.

Run any application, pytorch or otherwise, using ROCm's rocprof.  The PyTorch GPU traces will be collected as part of your rocprof output.

malfet (changes requested):
To further substantiate the claim, can you please specify in PR description why your are making it (I.e. you've tried to run the tutorial on GPU such-n-such using PyTorch-X.Y.Z and instead of certain output you got something else. Because right now you just add a blank statement, which would be very hard to verify

intermediate_source/tensorboard_profiler_tutorial.py
guptaaryan16: @malfet I quoted this function [`supported_activities`](https://github.com/pytorch/pytorch/blob/7f1cbc8b5a7de8794c36833baa47bb1343833589/torch/profiler/profiler.py#L35) in official PyTorch API which does not mention anything about ROCm GPUs. I suppose they should have something here for the ROCm profiling if that's the case.

Also the backtricks is due to an error in PySpelling test where `ROCm` is not in dictionary, so the test caught it as a misspelled word. I suppose I can either do this or add it in a custom dictionary list.

jeffdaily: We should update that comment.  ROCm profiler uses roctracer library to trace on-device HIP kernels.  Passing `ProfilerActivity.CUDA` is the correct activity to use for both CUDA and ROCm due to ROCm PyTorch's strategy of "masquerading" as CUDA so that users do not have to make any changes to their pytorch models when running on ROCm.  PyTorch chose to expose "CUDA" in public APIs and we chose to reuse them to make the transition to ROCm from CUDA easier on users.

jeffdaily: See https://pytorch.org/docs/stable/notes/hip.html.

guptaaryan16:
@malfet I was testing the issue #2014 but since I did not have specific hardware(i.e. AMD GPU) for the testing the trace I assumed that the person who raised the issue has provided a correct trace file which seems to give wrong output. Since the trace is fine for CPU and CUDA, on testing, and there as no mention of ROCm in profiler API if they are supported, I raised this PR. Can you verify this tutorial and https://pytorch.org/tutorials/intermediate/tensorboard_profiler_tutorial.html on ROCm and see if the output trace is correct? Seems like it is off in time for about 250ms which is not the case for CPU and CUDA devices. Thanks @jeffdaily

jeffdaily:
I'm assigning this to @hongxiayang to verify.

hongxiayang:
> I'm assigning this to @hongxiayang to verify.

https://github.com/pytorch/tutorials/pulls

guptaaryan16:
Ok seems like this issue is resolved, so I am closing this PR

PR #2677, opened 2023-11-13
Changes to pytorch tutorial for Spatial Transformer Networks to reflect the suggestion by https://github.com/theRealSuperMario

### LINKED ISSUE #1117: But in Spatial Transformer Network?
Hi,

I am opening this issue because I noticed a weird behavior of the the spatial transformer networks implementation (https://github.com/pytorch/tutorials/blob/78e91c54dd0cd4fb0d02dfcc86fe94d16ab03df6/intermediate_source/spatial_transformer_tutorial.py#L57)

I summarized my findings [here](https://github.com/theRealSuperMario/pytorch_stn). In short, what is happening is that
when the input is normalised and then fed to the STN, the `F.grid_sample` call adds a zero-padding, however, the normalisation changes the background value from `0` to `-mean/std`. 
(https://github.com/pytorch/tutorials/blob/78e91c54dd0cd4fb0d02dfcc86fe94d16ab03df6/intermediate_source/spatial_transformer_tutorial.py#L127)

This causes the STN to collapse very early and to actually never learn the correct transformation. You can actually see that in the example code already (https://pytorch.org/tutorials/intermediate/spatial_transformer_tutorial.html), because the learnt transformation is zooming OUT instead of zooming IN on the digits. For the original 28 x 28 images, this is not such a big problem, However, when you continue to cluttered MNIST as in the original publication, the difference is huge. Once again, please have a look [here](https://github.com/theRealSuperMario/pytorch_stn).

I think the tutorial for the STN should be updated and also include the cluttered MNIST example because that is what drives the point home. I would volunteer to do so, if the get the permission to go ahead.

Unfortunately, most other implementations I was able to find on the web also have this bug.

ver2king:
Fixes #1117 

## Checklist

- [x] Add a sharable link for the distorted MNIST dataset size 60*60 (credit to **Author**: 'Sandro Braun <https://github.com/theRealSuperMario>')
- [x] Add two additional Spatial Transformer Net (STNs) that digest the distorted MNIST dataset's image size, and with different padding modes in the "torch.grid_sample" layer (credit to **Author**: 'Sandro Braun <https://github.com/theRealSuperMario>')
- [x] Compare the performance of all STNs (for both the original and the distorted MNIST datasets)

cc @sekyondaMeta @svekars @carljparker @NicolasHug @kit1980 @subramen


svekars (commented):
Please remove all the changes you added from the `.idea` dir.

PR #2680, opened 2023-11-14
Fix to match library export

### LINKED ISSUE #2679: [BUG] - Failed to import Checkpoint for hyperparameter tuning tutorial
### Add Link

https://pytorch.org/tutorials/beginner/hyperparameter_tuning_tutorial.html

### Describe the bug

The example in the tutorial and raw Github file fail to run. The tutorial imports Checkpoint from ray.air when the Class is defined inside ray.train.

Documentation: https://docs.ray.io/en/latest/train/api/doc/ray.train.Checkpoint.html#ray.train.Checkpoint

Which throws the following error:

```
ImportError: cannot import name 'Checkpoint' from 'ray.air'
```

### Describe your environment

* Platform: MacOS Ventura 13.4.1, M2
* CUDA: No, MPS
* PyTorch version: 2.1.0

undecidedapollo:
Fixes #2679 

## Description
Update the tutorial example to reference ray.train.Checkpoint instead of the non-existent ray.air.Checkpoint

## Checklist

### DIFF
--- a/beginner_source/hyperparameter_tuning_tutorial.py
+++ b/beginner_source/hyperparameter_tuning_tutorial.py
@@ -49,7 +49,8 @@ from torch.utils.data import random_split
 import torchvision
 import torchvision.transforms as transforms
 from ray import tune
-from ray.air import Checkpoint, session
+from ray.train import Checkpoint
+from ray.air import session
 from ray.tune.schedulers import ASHAScheduler
 
 ######################################################################


undecidedapollo:
Closed this PR as it seems the API has changed a good bit since the tutorial was written and updating this tutorial will take a bit more work. Example error message notifying of some of the changes:

```python
AttributeError: The new `ray.train.Checkpoint` class does not support `from_dict()`. Instead, only directories are supported.

Example to store a dictionary in a checkpoint:

import os, tempfile
import ray.cloudpickle as pickle
from ray import train
from ray.train import Checkpoint

with tempfile.TemporaryDirectory() as checkpoint_dir:
  with open(os.path.join(checkpoint_dir, 'data.pkl'), 'wb') as fp:
    pickle.dump({'data': 'value'}, fp)

  checkpoint = Checkpoint.from_directory(checkpoint_dir)
  train.report(..., checkpoint=checkpoint)

Example to load a dictionary from a checkpoint:

if train.get_checkpoint():
  with train.get_checkpoint().as_directory() as checkpoint_dir:
    with open(os.path.join(checkpoint_dir, 'data.pkl'), 'rb') as fp:
      data = pickle.load(fp)
 ```
I'm still new to pytorch so may try again once I get a better feel of how the library works.

PR #2682, opened 2023-11-15
add a note to fix PyTorch issue 113019

jingxu10:
Fixes https://github.com/pytorch/pytorch/issues/113019

## Checklist

cc @sekyondaMeta @svekars @carljparker @NicolasHug @kit1980 @subramen

### DIFF
--- a/prototype_source/pt2e_quant_ptq_x86_inductor.rst
+++ b/prototype_source/pt2e_quant_ptq_x86_inductor.rst
@@ -109,6 +109,10 @@ We will start by performing the necessary imports, capturing the FX Graph from t
 
 Next, we will have the FX Module to be quantized.
 
+.. note::
+
+   Steps below should be run inside the ``torch.no_grad()`` scope.
+
 2. Apply Quantization
 ----------------------------
 


prototype_source/pt2e_quant_ptq_x86_inductor.rst
```diff
@@ -109,6 +109,10 @@ We will start by performing the necessary imports, capturing the FX Graph from t
 
 Next, we will have the FX Module to be quantized.
 
+.. note::
+
+   Steps below should be run inside the ``torch.no_grad()`` scope.
```
jerryzh168, line 114: only the lowering step I think? cc @leslie-fang-intel

leslie-fang-intel: Thanks for the comment @jerryzh168. Yean, to be more concrete, I believe the `lowering` and `benchmark` steps must run with ``torch.no_grad()`` scope. But since it's PTQ quant we are talking about here, I think the `prepare` and `convert` steps can also run with `torch.no_grad` context, right?

jerryzh168: `prepare` and `convert` can run with `torch.no_grad`, but whether these should be run in `torch.no_grad` is talking about requirement though, so the question is are these required or not?

leslie-fang-intel: Make sense. For PTQ, I think `prepare` and `convert` are not required to run under `torch.no_grad`. And the `lowering` and `benchmark` steps are must have. @jingxu10 could you help to make the changes accordingly?

PR #2686, opened 2023-11-16
fix: typo error

Foundsheep:
Fix typo.

cc @svekars @brycebortree


beginner_source/introyt/modelsyt_tutorial.py
```diff
@@ -342,7 +342,7 @@ def forward(self, sentence):
 # the 6x6 input.
 # 
 # **Normalization layers** re-center and normalize the output of one layer
-# before feeding it to another. Centering the and scaling the intermediate
+# before feeding it to another. Centering them and scaling the intermediate
```
svekars, line 345:
```suggestion
# before feeding it to another. Centering and scaling the intermediate
```

Foundsheep:
@svekars could you help with this PR, in terms of commit status(?). It seems like this PR is in pending because of that, but I can't figure out how to make it.

PR #2689, opened 2023-11-19
Update reinforcement_q_learning.py - use named tuples rather than indices

Rubydesic:
## Description
Tutorial uses index numbers to access the values of named tuple returned from torch.max() which is a little unclear. This changes it to use the names "indices" and "values" instead. This is a trivial change so I did not make a separate issue.

## Checklist

cc @vmoens @nairbv


vmoens (approved):
Thanks it's cleaner like this!

PR #2701, opened 2023-12-08
Add `gtag()` helper function used by star-rating

### LINKED ISSUE #2696: [BUG] - No longer collecting star-rating data for tutorials, since transition to GA4
### Add Link

The problem occurs with all the tutorials, but here is an example that I have been working with:

https://pytorch.org/tutorials/beginner/basics/quickstart_tutorial.html

### Describe the bug

Near the footer of each tutorial page is a UX of five stars which provides an opportunity for a user to rate the tutorial. (Refreshing the page resets the star rating.)

Prior to the transition to GA4 in June, 2023, we received event data when a user clicked on one of these stars. Since the transition to GA4 we are no longer seeing this data on the Google Analytics site.

One possibility is that the data _is_ still collected, but that I just don't see where it is reported on the Google Analytics site; the reporting UX for GA4 is different than for Universal Analytics (UA)--the previous version of Google Analytics. This seems unlikely given my exploration of the site and the reporting UX.

Another possibility is that the mechanism that we were using is no longer supported by GA4. This seems more likely because the code to send the star rating still exists on our tutorial pages--but it no longer appears to do anything. That said, it could be possible to update that the code to make it compatible with GA4. 

In the event that the existing code cannot be adapted to GA4, a fallback option is to use the expanded click-detection of GA4 to detect when the stars on clicked on--and report that using the GA4 reporting UX.

### Describe your environment

MacBook Pro ~ 14-inch, 2021
- Apple M1 Pro
- 16 GB RAM
- Storage: 303.33 GB available of 494.38 GB

Google Chrome ~ Version 119.0.6045.199 (Official Build) (arm64)

carljparker:
Originally proposed this change in the `pytorch/pytorch_sphinx_theme` repo as PR#188, but SveKars suggested we try it out in `pytorch/tutorials` which does not have as large a scope.

Fixes #2696

## Description
Add JavaScript helper function used by star-rating code...as well as other customer JavaScript in the turtorials.

## Checklist

### DIFF
--- a/_templates/layout.html
+++ b/_templates/layout.html
@@ -17,7 +17,11 @@ if((window.location.href.indexOf("/prototype/")!= -1) && (window.location.href.i
 {{ super() }}
 <script>
 
-  
+<!--
+  Helper function to make it easier to call dataLayer.push()
+-->
+function gtag(){window.dataLayer.push(arguments);}
+
 //add microsoft link
 
 if(window.location.href.indexOf("/beginner/basics/")!= -1)


_templates/layout.html
```diff
@@ -17,7 +17,11 @@
 {{ super() }}
 <script>
 
-  
+<!--
+  Helper function to make it easier to call dataLayer.push()
+-->
```
svekars, line 22:
```suggestion
<!-- Helper function to make it easier to call dataLayer.push() -->
```

_templates/layout.html
```diff
@@ -17,7 +17,9 @@
 {{ super() }}
 <script>
 
-  
+<!-- Helper function to make it easier to call dataLayer.push() -->
```
svekars, line 20:
```suggestion
// Helper function to make it easier to call dataLayer.push() 
```

svekars (approved):

carljparker:
Thanks, @svekars, for catching the issue with the comment markers.

PR #2728, opened 2024-01-10
Add missing `plt.show()` statement to Beginner/DCGAN Face Tutorial

Soooda:
## Description
I was following the instructions in the tutorial to set up the dataset loader and visualise its outputs according to the code block:
```python
plt.figure(figsize=(8,8))
plt.axis("off")
plt.title("Training Images")
plt.imshow(np.transpose(vutils.make_grid(real_batch[0].to(device)[:64], padding=2, normalize=True).cpu(),(1,2,0)))
```
It configures the plt graphing options but misses the final `plt.show()` command to open GUI. I ran the code above and nothing happened.

## Checklist


beginner_source/dcgan_faces_tutorial.py
```diff
@@ -265,7 +265,7 @@
 plt.axis("off")
 plt.title("Training Images")
 plt.imshow(np.transpose(vutils.make_grid(real_batch[0].to(device)[:64], padding=2, normalize=True).cpu(),(1,2,0)))
-
+plt.show()
```
svekars, line 268: My guess is that you are running this as a Python script? `plt.show()` is not necessary in notebooks which was the primary targeted environment for this tutorial. We can add this as a comment.
```suggestion
# If running as a Python script, uncomment this line to display the image: 
# plt.show()
```

svekars: Actually, let's just add it.

svekars (approved):

PR #2731, opened 2024-01-12
Fix the Title: Underline too short warnings

svekars:
Fixing the WARNING: Title underline too short
There are over 100 of those warnings in the build. While fine in the html, when built into a notebook, they produce broken output. 

cc @brycebortree


brycebortree (approved):
LGTM!

PR #2741, opened 2024-01-22
Process notebooks with pypandoc and a custom pandoc filter

svekars:
To fix a number of formatting issues with notebooks, adding pypandoc to process the notebooks and a custom filter to handle admonitions and blocks of code that have video frames. In addition to that, adding a post-processing script to clean up the resulting notebooks and remove the unneeded {=html} syntax

### DIFF
--- a/.github/workflows/build-tutorials.yml
+++ b/.github/workflows/build-tutorials.yml
@@ -187,7 +187,8 @@ jobs:
           echo "rm /opt/cache/bin/*" | docker exec -u root -i "${container_name}" bash
 
           docker exec -t "${container_name}" sh -c ".jenkins/build.sh"
-
+          docker exec -t "${container_name}" python "./post_process_notebooks.py"
+  
       - name: Upload docs preview
         uses: seemethere/upload-artifact-s3@v5
         if: ${{ github.event_name == 'pull_request' }}
--- a/conf.py
+++ b/conf.py
@@ -107,7 +107,10 @@ sphinx_gallery_conf = {
                             "# https://pytorch.org/tutorials/beginner/colab\n"
                             "%matplotlib inline"),
     'reset_modules': (reset_seeds),
-    'ignore_pattern': r'_torch_export_nightly_tutorial.py'
+    'ignore_pattern': r'_torch_export_nightly_tutorial.py',
+        'pypandoc': {'extra_args': ['--mathjax'],
+                     'filters': ['./custom_pandoc_filter.py'],
+        },
 }
 
 if os.getenv('GALLERY_PATTERN'):
--- a/custom_pandoc_filter.py
+++ b/custom_pandoc_filter.py
@@ -1,0 +1,68 @@
+from pandocfilters import toJSONFilter, Div, RawBlock, Para, Str, Space, Link, Code, CodeBlock
+import markdown
+import re
+
+def to_markdown(item):
+    if item['t'] == 'Str':
+        return item['c']
+    elif item['t'] == 'Space':
+        return ' '
+    elif item['t'] == 'Link':
+        # Assuming the link text is always in the first item
+        return f"[{item['c'][1][0]['c']}]({item['c'][2][0]})"
+    elif item['t'] == 'Code':
+        return f"`{item['c'][1]}`"
+    elif item['t'] == 'CodeBlock':
+        return f"```\n{item['c'][1]}\n```"
+
+def process_admonitions(key, value, format, meta):
+    if key == 'Div':
+        [[ident, classes, keyvals], contents] = value
+        if 'note' in classes:
+            color = '#54c7ec'
+            label = 'NOTE:'
+        elif 'tip' in classes:
+            color = '#6bcebb'
+            label = 'TIP:'
+        elif 'warning' in classes:
+            color = '#e94f3b'
+            label = 'WARNING:'
+        else:
+            return
+
+        note_content = []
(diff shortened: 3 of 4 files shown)


requirements.txt
```diff
@@ -4,10 +4,14 @@
...
+nbsphinx
 docutils==0.16
 sphinx-copybutton
-tqdm==4.66.1
-numpy==1.24.4
+pypandoc==1.12
+pandocfilters
+markdown
+tqdm
+numpy
```
malfet, line 14: Please do not unpin numpy and tqdm, but rather specify versions that do work for you (all those pins were added when updates broke tutorials)

.jenkins/custom_pandoc_filter.py
```diff
@@ -0,0 +1,132 @@
...
+            label = 'TIP:'
+        elif 'warning' in classes:
+            color = '#e94f3b'
+            label = 'WARNING:'
+        else:
+            return
+
+        note_content = []
+        for block in contents:
+            if 't' in block and block['t'] == 'Para':
```
malfet, line 46: Nit
```suggestion
            if block.get('t') == 'Para':
```

.jenkins/custom_pandoc_filter.py
```diff
@@ -0,0 +1,132 @@
...
+                for item in block['c']:
+                    if item['t'] == 'Str':
+                        note_content.append(Str(item['c']))
+                    elif item['t'] == 'Space':
+                        note_content.append(Space())
+                    elif item['t'] == 'Link':
+                        note_content.append(Link(*item['c']))
+                    elif item['t'] == 'Code':
+                        note_content.append(Code(*item['c']))
+            elif 't' in block and block['t'] == 'CodeBlock':
```
malfet, line 56: Nit (`dict.get` returns `None` if such index does not exist)
```suggestion
            elif block.get('t') == 'CodeBlock':
```

.jenkins/custom_pandoc_filter.py
```diff
@@ -0,0 +1,132 @@
...
+                                item_html += '<li>' + ''.join(to_markdown(i) for i in list_item[0]['c']) + '</li>'
+                            item_html += '</ul>'
+                    columns[column_num] += item_html
+                    column_num = (column_num + 1) % 2
+            columns = [column + '</div>' for column in columns]
+            return {'t': 'RawBlock', 'c': ['html', ''.join(columns)]}
+
+def is_code_block(item):
+    return item['t'] == 'Code' and 'octicon' in item['c'][1]
+def process_all(key, value, format, meta):
```
malfet, line 124: Nit (keep two empty lines between function definitions)
```suggestion


def process_all(key, value, format, meta):
```

.jenkins/custom_pandoc_filter.py
```diff
@@ -0,0 +1,132 @@
...
+    new_value = process_admonitions(key, value, format, meta)
+    if new_value is None:
+        new_value = process_images(key, value, format, meta)
+    if new_value is None:
+        new_value = process_grids(key, value, format, meta)
+    return new_value
+if __name__ == "__main__":
```
malfet, line 131: Nit (two empty lines between top level blocks)
```suggestion


if __name__ == "__main__":
```

.jenkins/custom_pandoc_filter.py
```diff
@@ -0,0 +1,132 @@
...
+            return {'t': 'RawBlock', 'c': ['html', ''.join(columns)]}
+
+def is_code_block(item):
+    return item['t'] == 'Code' and 'octicon' in item['c'][1]
+def process_all(key, value, format, meta):
+    new_value = process_admonitions(key, value, format, meta)
+    if new_value is None:
+        new_value = process_images(key, value, format, meta)
+    if new_value is None:
+        new_value = process_grids(key, value, format, meta)
```
malfet, line 129: Nit (this expresses what you are doing here slightly better, isn't it?)
```suggestion
    for transform in [process_admonitions, process_images, process_grids]:
        new_value = transform(key, value, format, meta)
        if new_value is not None:
            break
```

.jenkins/custom_pandoc_filter.py
```diff
@@ -0,0 +1,132 @@
...
+    # load correctly in the notebook.
+    if key == 'Image':
+        [ident, classes, keyvals], caption, [src, title] = value
+        if not src.startswith('http'):
+            while src.startswith('../'):
+                src = src[3:]
+            if src.startswith('/_static'):
+                src = src[1:]
+            src = 'https://pytorch.org/tutorials/' + src
+        return {'t': 'Image', 'c': [[ident, classes, keyvals], caption, [src, title]]}
```
malfet, line 95: Nit (explicitly return `None`, but return early) I might be wrong, in C++ it's considered a bad behavior to assume function will return None by default, but in Python it might be ok, though still a good thing to do implicitly
```suggestion
    if key != 'Image':
        return None
    [ident, classes, keyvals], caption, [src, title] = value
    if not src.startswith('http'):
        while src.startswith('../'):
            src = src[3:]
        if src.startswith('/_static'):
            src = src[1:]
        src = 'https://pytorch.org/tutorials/' + src
        
    return {'t': 'Image', 'c': [[ident, classes, keyvals], caption, [src, title]]}
```

malfet (approved):
Left a few comments (most particular, do not unpin the version of dependencies), but overall looks good to me

PR #2743, opened 2024-01-23
Fix calculation of current item number in optimization_tutorial.py

### LINKED ISSUE #2742: [BUG] - Current item calculation in Optimization tutorial prints wrong number when data set length doesn't split evenly to batches
### Add Link

https://pytorch.org/tutorials/beginner/basics/optimization_tutorial.html#full-implementation

### Describe the bug

On line 17 in Full Implementation code of Optimization tutorial the current item is calcuated like this
```
loss, current = loss.item(), (batch + 1) * len(X)
```
which gives the wrong number on the last item if data set size doesn't split evenly to batches.

### Describe your environment

web site

anttikup:
Current item calculation gave the wrong number when dataset doesn't split to batches equally.

Fixes #2742

## Description
Change `(batch + 1) * len(X)` to `batch * batch_size + len(X)`

## Checklist

cc @subramen @albanD


albanD (approved):
Thanks

PR #2754, opened 2024-01-29
Fix batch size calculation in dist_tuto

petergtz:
## Description
Batch size (`bsz`) must be an `int`, not a `float`. This change fixes it, basically doing the same as in https://github.com/seba-1511/dist_tuto.pth/blob/a552567061a9985cdcfe72ecb9b47e4630d6a7fe/train_dist.py#L85 (the sample linked from the tutorial).

## Checklist
- [x] This is a trivial tutorial fix. I have not created an issue for it.


wconstab (approved):
thanks for the fix!

PR #2761, opened 2024-02-02
Remove recipes_source/recipes from gitignore

clee2000:
Pretty sure we should not be ignoring recipes_source/recipes

Also it keeps messing with my ability to find files in vscode


malfet (approved):
Sorry, I understand now what it does?

PR #2762, opened 2024-02-03
Switch multigpu runner

clee2000:
Follows https://github.com/pytorch/pytorch/pull/105759/files

Old commits: 22min
This PR: 17min


malfet (approved):

malfet (commented):
Would be good to add time to finish the tests before/after the change (i.e. hopefully we are not slower than before)

PR #2768, opened 2024-02-08
Removed redundant variable and fixed attribute 'Lock issue on google-colab 

sajidahmed12:
Removed some redundant variable declaration and fixed attribute 'Lock issue on google-colab.

[1]. There was a common variable named train_iter and its called around 3-4 times which was redundant.  ```train_iter = iter(AG_NEWS(split="train"))``` called 3-4 times 

[2]. Added  `!pip install -U portalocker>=2.0.0` into a cell to ensure its dependency and added a suggestion to restart the kernal due to this [issue from stack-overflow](https://stackoverflow.com/questions/76302971/question-in-pytorch-transformer-tutorial-about-nonetype-object-has-no-attribut).


sajidahmed12:
Sure, already done, Please check the latest PR.

PR #2784, opened 2024-03-03
Update tensor_tutorial.py

decanbay:
matrix AxB is not equalt to A x B.T 
so 
tensor x tensor is tensor @ tensor , it is not tensor@tensor.T

Fixes #ISSUE_NUMBER

## Checklist

cc @subramen @albanD


albanD (commented):
I'm not sure what your point is? The code as updated will not run.

decanbay:
> I'm not sure what your point is? The code as updated will not run.

what do you mean by "as updated will not run?" The point is this sectoion of the tutorial is wrong and can be confusing to people. 
No other comment here.

albanD:
Ho sorry I though the "tensor" object was defined here https://github.com/pytorch/tutorials/blob/729eb0716881e268dabc5e704d188f66b356e641/beginner_source/blitz/tensor_tutorial.py#L81

And with these shapes, you need the transpose for the matmul to make sense, otherwise you would get a runtime error.
I'm still not sure what the difference is. If you want to make less weird on how "tensor" is re-used, I would just create tensor1 and tensor2 with appropriate shapes and do the matmul between these two things.

PR #2791, opened 2024-03-06
(Re-land) Update ray to 2.7.2

malfet:
To address numerous CVE that older Ray releases were affected by see https://www.anyscale.com/blog/update-on-ray-cves-cve-2023-6019-cve-2023-6020-cve-2023-6021-cve-2023-48022-cve-2023-48023

Use legacy APIs by setting RAY_AIR_NEW_PERSISTENCE_MODE environment variable to zero.

Move logic of adding fake properties to `sys.stdout` from `main` to top level imports, as `ray.tune` will expect `sys.stdout` to have an `encoding`, which is missing in sphinx
Please note, that this increases tutorial runtime from 4 to 10 min

### DIFF
--- a/beginner_source/hyperparameter_tuning_tutorial.py
+++ b/beginner_source/hyperparameter_tuning_tutorial.py
@@ -52,6 +52,9 @@ from ray import tune
 from ray.air import Checkpoint, session
 from ray.tune.schedulers import ASHAScheduler
 
+# FIXME: migrate to ray.train.Checkpoint and remove following line
+os.environ["RAY_AIR_NEW_PERSISTENCE_MODE"]="0"
+
 ######################################################################
 # Most of the imports are needed for building the PyTorch model. Only the last three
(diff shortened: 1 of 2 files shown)


beginner_source/hyperparameter_tuning_tutorial.py
```diff
@@ -452,8 +455,8 @@ def main(num_samples=10, max_num_epochs=10, gpus_per_trial=2):
     # Fixes ``AttributeError: '_LoggingTee' object has no attribute 'fileno'``.
     # This is only needed to run with sphinx-build.
     import sys
-
-    sys.stdout.fileno = lambda: False
+    if not not hasattr(sys.stdout, "encoding"):
```
malfet, line 458:
```suggestion
    if not hasattr(sys.stdout, "encoding"):
```

svekars (approved):

Issue #2693, opened 2023-11-29
optimizer.zero_grad call in wrong place in quickstart tutorial


Flagy:
In the quickstart tutorial linked here:
https://github.com/pytorch/tutorials/blob/main/beginner_source/basics/quickstart_tutorial.py

the call to optimizer.zero_grad function is not done in the right place, it should be put as first line in the training loop as explained in the optimization tutorial: https://pytorch.org/tutorials/beginner/basics/optimization_tutorial.html

The results of the network without applying this change are poor and not coherent with the goal of the tutorial.

garyyaoresearch:
> @garyyaoresearch you are right, actually in both tutorials the call to zero_grad is right after the call to step. However I've seen many times that zero_grad is actually called before the bacward and step. Also the accepted answer in the pytorch discussion [here](https://discuss.pytorch.org/t/what-step-backward-and-zero-grad-do/33301/7) appears to suggest the order as zero_grad, backward and then step in the training loop.
> 
> But the real problem for me was that the performances of the model are very poor on the MNIST_fashion dataset on the [quickstart_tutorial](https://github.com/pytorch/tutorials/blob/main/beginner_source/basics/quickstart_tutorial.py) and they improve when the zero_grad is put before.
> 
> Let me know if you found the same issue on the performances of the model on the indicated tutorial.

Would it possible to show us result logs for comparison? Is it possible that when making the change, you didn't restart the notebook or reinitialize the model?

albanD:
As discussed in https://github.com/pytorch/tutorials/pull/2800 I think we do want to keep the current order to ensure minimal memory use which is a generally important concern these days while it will preserve the same numerical result for these use cases.

PR #2800, opened 2024-03-15
Reset the gradients before compute them in loss.backward()

### LINKED ISSUE #2799: I think `optimizer.zero_grad()` should go before `loss.backward()`
I think `optimizer.zero_grad()` should go before `loss.backward()` to ensure that gradients are zeroed out before the calculation of the gradients that takes place inside  `loss.backward()`.

https://github.com/pytorch/tutorials/blob/30e14df26091ef3ac58c8327e1e5c5cbf578f382/beginner_source/basics/quickstart_tutorial.py#L157

jmarintur:
Fixes #2799 and #2693

## Description
Change backward propagation steps to the usual order. 

## Checklist

- [x] The issue that is being fixed is referred in the description
- [] Only one issue is addressed in this pull request
- [] Labels from the issue that this PR is fixing are added to this pull request

cc @subramen @albanD

### DIFF
--- a/beginner_source/basics/quickstart_tutorial.py
+++ b/beginner_source/basics/quickstart_tutorial.py
@@ -152,9 +152,9 @@ def train(dataloader, model, loss_fn, optimizer):
         loss = loss_fn(pred, y)
 
         # Backpropagation
+        optimizer.zero_grad()
         loss.backward()
         optimizer.step()
-        optimizer.zero_grad()
 
         if batch % 100 == 0:
             loss, current = loss.item(), (batch + 1) * len(X)


albanD (changes requested):
Both are equivalent in terms of the computed values and we actually prefer to zero_grad after the optimizer step to clear the memory and not keep it around during fw/ beginning of bw.

jmarintur:
Hi @albanD, funny thing that the accepted answer in [here](https://discuss.pytorch.org/t/what-step-backward-and-zero-grad-do/33301) was to do it before (I do it too), but I can perfectly understand why in scenarios when you want to free memory as soon as you can, doing it right after may be the best strategy. Thanks for your clarification. Perhaps we should also close https://github.com/pytorch/tutorials/issues/2799 and https://github.com/pytorch/tutorials/issues/2693. Wdyt?

albanD:
In general, any order works if you're not memory constrained (which was the case for most people in 2018 :D ). But these days, I think people care a lot more about memory use!

geopapa11:
Hi @albanD out of curiosity why placing `optimizer.zero_grad()` right after `optimizer.step()` is more memory-efficient? 

Is it because you are resetting all gradients right before you execute the next step of the internal loop `for batch, (X, y) in enumerate(dataloader)` and hence you release the respective memory before the execution of the forward propagation step?

While when `optimizer.zero_grad()` is before `optimizer.step()`, this means that memory is allocated for the gradients before we enter into the computation of the forward propagation step. And this means that less memory is available for forward propagation? 

Is that the logic? Thanks in advance for your response! 😊

PR #2824, opened 2024-04-04
Add tutorial for swap_tensors in nn.Module

mikaylagawarecki:
as title


recipes_source/recipes/swap_tensors.py
```diff
@@ -0,0 +1,246 @@
...
+# ``nn.Module.load_state_dict()``
+# --------------------------------
+# At present, depending on the value of the ``assign`` keyword argument passed
+# to ``load_state_dict()``, there are two ways to load the ``state_dict``:
+#
+# 1. ``assign=False``: in-place copy (i.e. ``param.copy_(state_dict['param'])``)
+# 2. ``assign=True``: ``__setattr__`` (i.e. ``module.param = state_dict['param']``)
+#
+#
+# Each has its own limitations -- ``assign=False`` imposes the constraint that
```
svekars, line 160:
```suggestion
# Each approach has its own limitations -- ``assign=False`` imposes the constraint that
```

recipes_source/recipes/swap_tensors.py
```diff
@@ -0,0 +1,246 @@
...
+# the type of the parameter in the state_dict must be the same as the type of
```
svekars, line 161:
```suggestion
# the type of the parameter in the ``state_dict`` must be the same as the type of
```

recipes_source/recipes/swap_tensors.py
```diff
@@ -0,0 +1,246 @@
...
+                   )
+
+with torch.device("meta"):
+    m = nn.Linear(3, 5)
+    m.apply(fn)
+
+#################################################################################
+# We can then load the ``state_dict``. Observe that we use ``assign=True`` because
+# for biases, we want to preserve the properties of the tensor in the ``state_dict``
+# (i.e. we do not want the bias to be on the ``meta`` device after loading).
```
svekars, line 222:
```suggestion
# (for example, we do not want the bias to be on the ``meta`` device after loading).
```

recipes_source/recipes/swap_tensors.py
```diff
@@ -0,0 +1,246 @@
...
+state_dict = nn.Linear(3, 5).state_dict()
+print(f"state_dict:\n {state_dict}")
+m.load_state_dict(state_dict, assign=True)
+print(f"After: id(weight)={id(m.weight)}, id(bias)={id(m.bias)}")
+print(f"m.state_dict() after load_state_dict():\n {m.state_dict()}")
+
+#################################################################################
+# The above is a toy example of how we can use the new extension point in
+# ``nn.Module.load_state_dict()``. One can also imagine alternate scenarios such
+# as when we have tensor subclasses in the state_dict and plain ``nn.Parameters``/
```
svekars, line 236:
```suggestion
# as when we have tensor subclasses in the ``state_dict`` and plain ``nn.Parameters``/
```

recipes_source/recipes/swap_tensors.py
```diff
@@ -0,0 +1,246 @@
...
+# tensors in the module or when both are tensor subclasses. Based on the use
+# case, we can define the ``__torch_function__`` handler for ``module_load``
+# to apply the transforms as needed.
+#
+# Conclusion
+# ----------
+# In this tutorial, we learned about ``swap_tensors``, the importance
```
svekars, line 243:
```suggestion
# In this recipe, we learned about ``swap_tensors``, the importance
```

svekars (commented):
Thanks so much @mikaylagawarecki! Looks great - I added a few editorial suggestions - feel free to add/ignore as needed.

recipes_source/recipes_index.rst
```diff
@@ -151,6 +151,13 @@ Recipes are bite-sized, actionable examples of how to use specific PyTorch featu
    :link: ../recipes/torch_logs.html
    :tags: Basics
 
+.. customcarditem::
+   :header: Extension points in nn.Module for loading state_dict and tensor subclasses
+   :card_description: New extension points in nn.Module.
+   :image: ../_static/img/thumbnails/cropped/generic-pytorch-logo.png
+   :link: ../recipes/recipes/module_load_state_dict_tips.html
```
svekars, line 158: this should be ``recipes/recipes/swap_tensors.html``

mikaylagawarecki: whoops! good catch, thank you!

recipes_source/recipes/swap_tensors.py
```diff
@@ -0,0 +1,246 @@
...
+t1 = torch.arange(2)
+t2 = torch.arange(3)
+print(f"Before swapping, t1: {t1}, t2: {t2}")
+torch.utils.swap_tensors(t1, t2)
+print(f"After swapping, t1: {t1}, t2: {t2}")
+
+################################################################################
+# More specifically, ``swap_tensors`` swaps the python ``__class__``, ``__dict__``
+# and ``__slots__`` of the two tensors, as well as their associated ``at::Tensor``.
+#
```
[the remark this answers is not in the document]
mikaylagawarecki: Hm I think this is discussed in `Application to ``nn.Module``` right below, so I didn't add anything here

recipes_source/recipes/swap_tensors.py
```diff
@@ -0,0 +1,237 @@
...
+#
+# Application to ``nn.Module``
+# ----------------------------
+# This utility is pertinent to ``nn.Module`` when a Python object outside
+# of the module holds a reference to parameters of the module. If an ``nn.Module``
+# modifies any of its parameters out of place, the object holding references to
+# the parameters will not see the change. A classic example of this is the
+# optimizer, which holds a reference to the parameters of the ``nn.Module``.
```
albanD, line 41: nit: mention that the impact here is silent correctness where the optimizer step() will run without error but the Module weights will never be updated!

recipes_source/recipes/swap_tensors.py
```diff
@@ -0,0 +1,237 @@
...
+# ``nn.Module.to()`` and related methods
+# --------------------------------------
+# This includes methods that change the device of the module (such as ``nn.Module.cpu()``),
+# methods that change the ``dtype`` of the module (such as ``nn.Module.float()``)
+# as well as methods that allow the module to be materialized
+# (such as ``nn.Module.to_empty()``).
+#
+# At first glance, it might be non-intuitive that these methods are able to
+# modify the parameters of the module in-place. The existing approach has been
+# to set the ``.data`` of the module under the hood (``param.data = new_param``).
```
albanD, line 61: Can we not name it? :D I don't want people to see this and realize they can do it. "The existing approach has been to use a nasty hack dating back from the first days of PyTorch."

recipes_source/recipes/swap_tensors.py
```diff
@@ -0,0 +1,237 @@
...
+#
+# Notably, the existing approach does not work in these cases:
+#
+# * when using ``__torch_dispatch__`` subclasses
+# * when ``param`` and ``new_param`` do not have the same type
```
albanD, line 66: I think you need to define "type" here.

recipes_source/recipes/swap_tensors.py
```diff
@@ -0,0 +1,237 @@
...
+# This includes methods that change the device of the module (such as ``nn.Module.cpu()``),
+# methods that change the ``dtype`` of the module (such as ``nn.Module.float()``)
+# as well as methods that allow the module to be materialized
+# (such as ``nn.Module.to_empty()``).
+#
+# At first glance, it might be non-intuitive that these methods are able to
+# modify the parameters of the module in-place. The existing approach has been
+# to set the ``.data`` of the module under the hood (``param.data = new_param``).
+#
+# Notably, the existing approach does not work in these cases:
```
albanD, line 63: We can mention Tensor with different c++ representation like sparse or xla device.

recipes_source/recipes/swap_tensors.py
```diff
@@ -0,0 +1,237 @@
...
+class MyQuantizedLinearWeight(torch.Tensor):
+    @staticmethod
+    def __new__(cls, elem, scale, **kwargs):
+        return torch.Tensor._make_wrapper_subclass(
+            cls,
+            elem.shape,
+            dtype=elem.dtype,
+            layout=elem.layout,
+            device=elem.device,
+            requires_grad=elem.requires_grad,
```
albanD, line 84: This creates a leaf. Please take requires_grad as a kwarg.
Also you can remove **kwargs since we don't use it

recipes_source/recipes/swap_tensors.py
```diff
@@ -0,0 +1,237 @@
...
+print(f"After: id(m.weight)={id(m.weight)}, id(m.bias)={id(m.bias)}")
+print(f"m.weight.dtype: {m.weight.dtype}")
+print(f"m.weight.elem.dtype: {m.weight.elem.dtype}")
+print(f"m.bias.dtype: {m.bias.dtype}")
+torch.__future__.set_swap_module_params_on_conversion(False)
+
+################################################################################
+# ``nn.Module.load_state_dict()``
+# --------------------------------
+# Depending on the value of the ``assign`` keyword argument passed
```
albanD, line 144: While this is historically accurate, I'm wondering if we want to twist the story here a little bit to better fit what we want today:
- assign=False preserves the properties of param and only takes the value from the state dict.
- assign=True preserves the properties and values of the state dict object.

It was done before with copy_ and setattr. But now has a different implementation since we have swap tensor and we can remove the limitations of the previous impl.

recipes_source/recipes/swap_tensors.py
```diff
@@ -0,0 +1,237 @@
...
+# Assume that we expect the ``state_dict`` to contain plain tensors and the
+# module to contain ``MyQuantizedLinearWeight`` parameters where we want the
+# tensors in the ``state_dict`` to be transformed into the subclass. Then we
+# can define a ``__torch_function__`` handler for ``torch.Tensor.module_load``
+# as such:
+
+@classmethod
+def custom_torch_function(cls, func, types, args=(), kwargs=None):
+    kwargs = {} if kwargs is None else kwargs
+    def module_load(dest, src, assign=False):
```
albanD, line 182: nit: not sure what this indirection brings rather than having these two lines in the if statement below?

albanD (commented):
Very cool! Small suggestions only.

recipes_source/recipes/swap_tensors.py
mikaylagawarecki: Updated to say Python ``type()``, do you think this is sufficiently expressive?

recipes_source/recipes/module_load_state_dict_tips.py
```diff
@@ -152,8 +153,13 @@ def my_processing_function(key, device):
 # ``nn.Module.parameters()``, the optimizer must be initialized after the module
 # is loaded from state dict if ``assign=True`` is passed.
 
+# As of PyTorch 2.3.0, one can use ``torch.__future__.set_swap_module_params_on_conversion`` to
+# avoid this caveat. This `recipe <https://pytorch.org/tutorials/recipes/recipes/module_load_state_dict_tips.html>`_
```
mikaylagawarecki, line 157:
```suggestion
# avoid this caveat. This `recipe <https://pytorch.org/tutorials/recipes/recipes/swap_tensors.html>`_
```

albanD (approved):
Sounds good! Thanks!

svekars (approved):

PR #2830, opened 2024-04-09
Use optimizer.zero_grad(set_to_none=True) instead of optimizer.zero_g…

josephyooo:
…rad()

## Description

This parameter sets the grads to none rather than zero, generally reducing memory footprint and improving performance. The risk is that the grads will appear and behave differently if the user tries to access and manually perform ops on it.

References:
https://pytorch.org/docs/master/generated/torch.optim.Optimizer.zero_grad.html#torch.optim.Optimizer.zero_grad
https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html

## Checklist

- [x] I have not created an issue fixed by this trivial tutorial patch.

cc @albanD


albanD (commented):
We made this the default ~a year ago. So not needed anymore!

PR #2838, opened 2024-04-15
[export] Add non-strict and derived dims tutorials

angelayi:


intermediate_source/torch_export_tutorial.py
```diff
@@ -180,8 +181,68 @@ def forward(self, x):
     tb.print_exc()
 
 ######################################################################
-# The sections below demonstrate some ways you can modify your code
-# in order to remove graph breaks.
+# Non-Strict Export
+# -----------------
+#
+# To trace the program, torch.export uses TorchDynamo, a bytecode analysis
```
svekars, line 187:
```suggestion
# To trace the program, ``torch.export`` uses TorchDynamo, a byte code analysis
```

intermediate_source/torch_export_tutorial.py
```diff
@@ -180,8 +181,68 @@ def forward(self, x):
...
+# engine, to symbolically analyze the Python code and build a graph based on the
+# results. This analysis allows torch.export to provide stronger guarantees
```
svekars, line 189:
```suggestion
# results. This analysis allows ``torch.export`` to provide stronger guarantees
```

intermediate_source/torch_export_tutorial.py
```diff
@@ -180,8 +181,68 @@ def forward(self, x):
...
+# about safety, but not all Python code is supported, causing these graph
+# breaks.
+#
+# So in PyTorch 2.3, we introduced a new mode of exporting called
```
svekars, line 193:
```suggestion
# To address this inconvenience, in PyTorch 2.3, we introduced a new mode of exporting called
```

intermediate_source/torch_export_tutorial.py
```diff
@@ -180,8 +181,68 @@ def forward(self, x):
...
+class Bad2(torch.nn.Module):
+    def forward(self, x):
+        x.data[0, 0] = 3
+        return x
+
+bad2_nonstrict = export(Bad2(), (torch.randn(3, 3),), strict=False)
+print(bad2_nonstrict.module()(torch.ones(3, 3)))
+
+######################################################################
+# - calling unsupported functions (such as many built-in functions) traces
```
svekars, line 212:
```suggestion
# - Calling unsupported functions (such as many built-in functions) traces
```

intermediate_source/torch_export_tutorial.py
```diff
@@ -180,8 +181,68 @@ def forward(self, x):
...
+    def forward(self, x):
+        x = x + 1
+        return x + id(x)
+
+bad3_nonstrict = export(Bad3(), (torch.randn(3, 3),), strict=False)
+print(bad3_nonstrict)
+print(bad3_nonstrict.module()(torch.ones(3, 3)))
+
+######################################################################
+# - Unsupported Python language features (e.g. throwing exceptions, match
```
svekars, line 227:
```suggestion
# - Unsupported Python language features (such as throwing exceptions, match
```

intermediate_source/torch_export_tutorial.py
```diff
@@ -180,8 +181,68 @@ def forward(self, x):
...
+        except:
+            x = x + 2
+        return x
+
+bad4_nonstrict = export(Bad4(), (torch.randn(3, 3),), strict=False)
+print(bad4_nonstrict.module()(torch.ones(3, 3)))
+
+
+######################################################################
+# However, there are still some features that requires rewrites to the original
```
svekars, line 244:
```suggestion
# However, there are still some features that require rewrites to the original
```

svekars (commented):
Thanks so much, @angelayi! A quick editorial pass - looks good! =)

intermediate_source/torch_export_tutorial.py
```diff
@@ -365,6 +426,29 @@ def forward(self, x, y):
 except Exception:
     tb.print_exc()
 
+######################################################################
+# We can also describe one dimension in terms of other.
+
+class DerivedDimExample(torch.nn.Module):
```
pianpwk, line 432: There's some crazier stuff that Avik's built into derived dims that users probably don't expect to work, but actually do work, maybe it's worth putting those on display? Stuff like (from test_export.py)

```
class Foo(torch.nn.Module):
    def forward(self, z, y):
        return z[1:] + y[1::3]

foo = Foo()

z, y = torch.randn(4), torch.randn(10)

dx = torch.export.Dim("dx", min=3, max=6)
dz = dx + 1
dy = dz * 3 + 1  # works
ep = export(
    foo,
    (z, y),
    dynamic_shapes=({0: dimz}, {0: dimy}),
)
assert ep.module()(torch.randn(7), torch.randn(19)).size()[0] == 6
```

We also allow static dim specs (e.g. dimx = 5, crashes if the value is wrong), as just a nice to have, maybe that's worth adding in a previous basic shapes example? I can also help with that.

intermediate_source/torch_export_tutorial.py
```diff
@@ -365,6 +426,29 @@ def forward(self, x, y):
 except Exception:
     tb.print_exc()
 
+######################################################################
+# We can also describe one dimension in terms of other.
```
pianpwk, line 430: I think the restriction here is as a linear relation, A * Dim + B... not sure if there's more though. Avik would know best.

pianpwk:
This looks great overall!!

svekars (approved):

PR #2839, opened 2024-04-17
Remove line overwriting output

### LINKED ISSUE #2827: Misleading example for per-sample gradient
In the example of per-sample gradient, the following line can be misleading since the `predictions` of a net are logits:  
https://github.com/pytorch/tutorials/blob/08a61b7cae9d00312d0029b1f86a248ec1253a83/intermediate_source/per_sample_grads.py#L49

The correct way should be: 
``` python
return F.nll_loss(F.log_softmax(predictions, dim=-1), targets) 
```

Would appreciate if this can be corrected.

jmarintur:
Fixes #2827

## Description
In the forward function of the CNN located in per_sample_grads.py, before the return, there is a line, L45, that is overwriting the output: 

https://github.com/pytorch/tutorials/blob/08a61b7cae9d00312d0029b1f86a248ec1253a83/intermediate_source/per_sample_grads.py#L42-L46

We simply need to remove it.

## Checklist
- [x] The issue that is being fixed is referred in the description.
- [x] Only one issue is addressed in this pull request.
- [x] Labels from the issue that this PR is fixing are added to this pull request.

cc @albanD


albanD (approved):
Yep!

cc @svekars

svekars (approved):

PR #2846, opened 2024-04-24
Pin torch version to 2.3 and rebuild docker image

clee2000:
pin torch version to most recent (2.3) to trigger docker rebuild
This way if we (or a pr author) wants to update the torch version they can just change the pin


svekars (approved):

svekars:
Thank you!

PR #2876, opened 2024-05-18
Tutorial timestamps

### LINKED ISSUE #2848: 💡 [REQUEST] - Add a Timestamp to Every Tutorial
### 🚀 Descirbe the improvement or the new tutorial

I would like to request the addition of timestamps to the tutorial pages for providing the information about when the tutorial was last updated or revised.

_No response_

### Additional context

_No response_

cc @svekars @brycebortree

brcolli:
Fixes #2848

## Description
I wrote a script that goes to each .py and .rst file in beginner_source, intermediate_source, and advanced_source directories. The timestamp format is "Updated: HH:MM AM/PM, mm dd, yyyy". It appends the timestamp to the end of the file. At first, I added it to the line above or below the author line, but this had a lot of edge cases, depending on how a particular person wrote their tutorial. I could broach this again, but I figured that was an overengineered solution to a simple problem.

The timestamp is generated by looking at git logs for the specific file and seeing when it was last committed.
Alternative solution was looking at last modified time for the file, but that isn't concrete or modular.
Another possible alternative is adding a check in the pipeline to ensure the user manually inserts the line before they commit. This is the safest but least user friendly.

I created a .sh script to get called during the build-tutorials workflow. It is placed at the very beginning of the manager job or right before running make docs in the worker job.
This does add a few seconds to the build time, but nothing major.

I was able to test this on my forked repo with a simple runner, but I obviously couldn't test this in your dedicated pipeline.

## Checklist


.jenkins/update_timestamps.py
```diff
@@ -0,0 +1,57 @@
...
+def get_last_commit_timestamp_for_file(file_path: str) -> str:
+    """Get the last commit timestamp for a file.
+
+    Args:
+        file_path (str): Path to file
+
+    Returns:
+        str: Last committed timestamp string
+    """
+    git_command = ["git", "log", "-1", "--format=%at", "--", file_path]
```
svekars, line 15:
```suggestion
    git_command = ["git", "log", "-1", '--date=format:%B %d, %Y', "--format=%at", "--", file_path]
```

.jenkins/update_timestamps.py
```diff
@@ -0,0 +1,57 @@
...
+    Args:
+        file_path (str): Path to file
+    """
+    with open(file_path, 'r') as file:
+        lines = file.readlines()
+
+    # Get current timestamp
+    timestamp = get_last_commit_timestamp_for_file(file_path)
+    timestamp_line = f'{"# " if file_path.endswith("py") else ""}**Updated:** *{timestamp}*\n'
+    timestamp_pattern = r'\*\*Updated:\*\*\s\**\d{1,2}:\d{2} [AP]M, \w+ \d{1,2}, \d{4}\*'
```
svekars, line 37: I would rather try to avoid regexes like this

.jenkins/update_timestamps_batch.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+SOURCEDIR=$1
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+
+directories=("$SOURCEDIR/beginner_source" "$SOURCEDIR/intermediate_source" "$SOURCEDIR/advanced_source")
```
svekars, line 7: We have other directories as well and what if we add a new gallery directory, this won't work anymore? You could try perhaps to get them from the conf.py sphinx-gallery settings?

svekars (changes requested):
Thank you for taking a stab at this! A few things:
* I like that you are using git log. Can we please change the format from: "--format=%at" to `'--date=format:%B %d, %Y' "--format=%cd'"`
* I would really prefer the timestamp to appear at the top of the page, after the title perhaps, rather than at the bottom. Also, it should say:  `Last Updated: <timestamp>`, with font: 300 italics. 
* I see that right now, it is only working for beginner/, intermediate/ and advanced/ paths. so any subdirs like beginner/basics are excluded. Can it be fixed?

.jenkins/update_timestamps.py
brcolli: I updated it! For my own knowledge, may I ask why? Is it because it's too specific? I was trying to be specific just so I didn't grab any lines unexpectedly from the tutorial files.

.jenkins/update_timestamps_batch.sh
brcolli: I updated it to get the files from the python script get_sphinx_filenames. This is what is used in conf.py, so can I assume that is a complete set?

.jenkins/update_timestamps.py
brcolli: I also think it looks better if it's right after the title, but this comes with some caveats. Since we do not standardize the title across .py and .rst files, I would have to get very broad. Many people start files in many different ways. Often they follow the line with a separator like === or *** but some files don't have this. I would love your input, the best I came up with was "Define title as first line with alphabet characters, except in .rst files that begin with `". Is that an acceptable definition of "title"?

brcolli:
I also think it looks better if it's right after the title, but this comes with some caveats. Since we do not standardize the title across .py and .rst files, I would have to get very broad. Many people start files in many different ways. Often they follow the line with a separator like === or *** but some files don't have this. I would love your input, the best I came up with was "Define title as first line with alphabet characters, except in .rst files that begin with `". Is that an acceptable definition of "title"?

brcolli:
Finally, you requested that the font is weight 300. Is that the default weight used by our stylings for the HTML/CSS files?

svekars:
> I also think it looks better if it's right after the title, but this comes with some caveats. Since we do not standardize the title across .py and .rst files, I would have to get very broad. Many people start files in many different ways. Often they follow the line with a separator like === or *** but some files don't have this. I would love your input, the best I came up with was "Define title as first line with alphabet characters, except in .rst files that begin with `". Is that an acceptable definition of "title"?

You could implement a different approach - instead of trying to match that in .rst/.py, you could have timestamps say in a yml file and load them with .js based after the HTML <title> tag as we can't control which type of heading authors will implement.

PR #2920, opened 2024-06-10
FIX: Update `optimization_tutorial.py` to expose `batch_size` in `train_loop()`

loganthomas:
## Description
`batch_size` is not currently exposed as `train_loop()` parameter

## Checklist

- [X] Labels from the issue that this PR is fixing are added to this pull request

cc @subramen @albanD @sekyondaMeta @svekars @kit1980 @brycebortree


svekars:
Can you please specify which docathon issue it fixes?

loganthomas:
Sorry, this doesn't solve an existing issue. I saw this when going through the tutorial myself. I tried to remove the docathon label but I didn't know how.

PR #2945, opened 2024-06-19
FIX: Update `optimization_tutorial.py` to expose `batch_size` in `train_loop()`

loganthomas:
## Description
`batch_size` is not currently exposed as `train_loop()` parameter
- Re-implements #2920 (accidental deletion from forked repo)

## Checklist

- [X] Labels from the issue that this PR is fixing are added to this pull request

cc @subramen @albanD @sekyondaMeta @svekars @kit1980 @brycebortree

### DIFF
--- a/beginner_source/basics/optimization_tutorial.py
+++ b/beginner_source/basics/optimization_tutorial.py
@@ -147,7 +147,7 @@ optimizer = torch.optim.SGD(model.parameters(), lr=learning_rate)
 # We define ``train_loop`` that loops over our optimization code, and ``test_loop`` that
 # evaluates the model's performance against our test data.
 
-def train_loop(dataloader, model, loss_fn, optimizer):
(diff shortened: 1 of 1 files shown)


subramen:
Hi, thanks for the PR but the proposed change is not correct. The `dataloader` takes in the batch size arg (as seen here https://github.com/pytorch/tutorials/blob/0740801fa8df2464f4d931d36a4bd7ada9cd0519/beginner_source/basics/optimization_tutorial.py#L47)

loganthomas:
> The `dataloader` takes in the batch size arg (as seen here)


Yes, but part of the `train_loop` uses the `batch_size` for reporting: https://github.com/pytorch/tutorials/blob/0740801fa8df2464f4d931d36a4bd7ada9cd0519/beginner_source/basics/optimization_tutorial.py#L166

PR #2954, opened 2024-06-24
Updating "Classifying Names with a Character-Level RNN"

### LINKED ISSUE #1166: Char RNN classification with batch size
I'm replicating [this example](https://github.com/pytorch/tutorials/blob/master/intermediate_source/char_rnn_classification_tutorial.py) for a **classification** with a **char-rnn**.
```python
for iter in range(1, n_iters + 1):
    category, line, category_tensor, line_tensor = randomTrainingExample()
    output, loss = train(category_tensor, line_tensor)
    current_loss += loss
```
I see that every epoch only 1 example is taken and random. I would like that each epoch **all the dataset** is taken with a specific **batch size** of examples. I can adjust the code to do this myself but I was wondering if some flags already exist.

Thank you

mgs28:
Fixes #1166

## Description
Updating Sean's excellent RNN classification tutorial that is now 8 years old and missing some newer pytorch functionality.
 
* Cannot use default Dataloader  to select batch sizes because "stack expects each tensor to be of equal size" but each of the names are of different length. However, updated the code to use mini batches without Dataloader functionality. 
* Introducing pytorch's Datasets class, we show how to split the data into train and test datasets which changes the training explanation. 
* Rewrote pieces of the tutorial to use three classes to improve re-use (Data, DataSet and RNN).
* Added a little more explanation to how RNNs score multi-character strings and their 2D matrix of tensors.  
* Changed evaluation from random training examples to an entire the test set.  
* removed some of the command line explanations since notebooks are used more often. 
* tried to preserve as much of the original text, functions and style as possible. 

## Checklist

cc @albanD


mgs28:
@svekars - it looks like you are active a lot in this repo, any chance you could help me with this? Thanks!

mgs28:
Added functionality to process training data in mini batches to satisfy original story. However, had to use numpy + random to create batch indices from a given dataset.

mgs28:
@svekars -  would you please help me with this tutorial update pull request or point me to someone who could?

svekars:
cc: @spro

mgs28:
Sorry about the spelling errors! I ran pyspelling and re-ran make html for the tutorial. This should pass those CI steps now.

I also added a story for me to come back and update the CONTRIBUTING.md to include some of these checks. (https://github.com/pytorch/tutorials/issues/2969) 

Thanks @spro @svekars !

mgs28:
@spro and @svekars - I significantly cut the training time although it is faster on my CPU than GPU. It runs in 72 seconds on my local CPU. I also added some default device so it looks for your CUDA build machines to hopefully make it faster. 

Thanks!

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -3,6 +3,7 @@
 NLP From Scratch: Classifying Names with a Character-Level RNN
 **************************************************************
 **Author**: `Sean Robertson <https://github.com/spro>`_
+**Updated**: `Matthew Schultz <https://github.com/mgs28>`_
```
svekars, line 6: we typically don't add Update. Please remove this

mgs28: done!

mgs28: @svekars  - anything else I can do to make this better? thanks!

mgs28: @svekars - hello, are there other items I should address here? I appreciate your help with this!

jbschlosser (commented):
Thanks for the PR! At a high level, I have the following comments / concerns:
* I think it's a good idea to update the tutorial to utilize PyTorch's `Dataset` and `DataLoader` abstractions.
    * I'm not sold on the need for the `NameData` class. It adds what I consider unnecessary complexity / code. It's perfectly simple to do any conversions between names <-> tensors as simple standalone functions utilized by the dataset.
* I agree that proper train / test / validation splits should be done in the tutorial, so that's a nice addition.
* I'm okay with defining a custom `RNN` module for illustration purposes, although in practice we'd encourage the use of `nn.RNN`, and I'm not sure if this existed when the tutorial was originally written. One thing the officially provided module in `torch.nn` provides is better perf due to e.g. cuDNN-accelerated kernels. If we don't use this within the tutorial, I think it should at least be mentioned and recommended for the perf benefits. All that said, I have some comments on the `RNN` module defined in the tutorial:
    * It's a bit confusing to redefine it multiple times in the tutorial, adding stuff to it each time. I'd recommend a single definition.
    * The `learn()` API does not belong on `RNN` and I suggest leaving the training logic in a standalone `train()`. This way, it's more PyTorch-idiomatic and easier for users to switch to some third-party training API (e.g. ignite, PyTorch Lightning, etc.).

mgs28:
@jbschlosser - thank you for the lovely suggestions to improve. If possible, I'd like to split into two things:

First, the edits to my existing content.
1. Excellent point on NameData. I removed it and used helper functions.
2. I'm glad you like the Dataset addition - that was my prompt for doing this.
3. Thanks for letting me know the multiple definitions were confusing. I focused it on one simple one, particularly since training was split from the object.
4. I added training as a separate function. I will go learn more about those third party trainers since I haven't used them.

Secondly, I would really like to use the nn.RNN if possible. There are very few tutorials that mention them and everyone seems to drive their RNN builds off this tutorial. However to solve this task, I think I need a network with layers like [57, 128, 18] and it looks like the default Elman networks are stuck at [57, 18] with layers. 

Is best practice to inherit from nn.RNN and add my own fully connected output layer or am I misunderstanding something? 

Thanks!

mgs28:
To make it simpler, I assume extending the nn.RNN class might look like (which runs about 40% faster) 

class MyRNN(nn.RNN):
    def __init__(self,input_size, hidden_size, output_size):
        super(MyRNN, self).__init__(input_size, hidden_size)
        
        self.h2o = nn.Linear(hidden_size, output_size)
        self.softmax = nn.LogSoftmax(dim=1)
    
    def forward(self, line_tensor):
        # Pass the input through the RNN layers
        rnn_out, hidden = super(MyRNN, self).forward(line_tensor)
        output = self.h2o(hidden[0])
        output = self.softmax(output)
        
        return output

jbschlosser:
> Is best practice to inherit from nn.RNN and add my own fully connected output layer or am I misunderstanding something?

Rather than inherit, we generally encourage composition. In this case, something like:
```
# better name needed :)
class MyRNN(nn.Module):
    def __init__(self, ...):
        super().__init__()
        self.rnn = nn.RNN(...)
        self.h2o = nn.Linear(...)

    ...
    def forward(self, x):
        x = self.rnn(x)
        x = self.h2o(x)
        return F.softmax(x, dim=-1)
```

mgs28:
Thanks! I used nn.rnn in composition and changed some of the surrounding text. That forced the addition of a few terms to the repo dictionary. I appreciate you teaching me something again and hopefully the tutorial is better for it.

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -178,114 +218,58 @@ def lineToTensor(line):
 # held hidden state and gradients which are now entirely handled by the
 # graph itself. This means you can implement a RNN in a very "pure" way,
 # as regular feed-forward layers.
-#
-# This RNN module implements a "vanilla RNN" an is just 3 linear layers 
-# which operate on an input and hidden state, with a ``LogSoftmax`` layer 
-# after the output.
+# 
+# This CharRNN class implements an RNN with three components. 
+# First, we use the `nn.RNN implemnentation <https://pytorch.org/docs/stable/generated/torch.nn.RNN.html>`__
```
jbschlosser, line 223:
```suggestion
# First, we use the `nn.RNN implementation <https://pytorch.org/docs/stable/generated/torch.nn.RNN.html>`__ 
```

jbschlosser (commented):
Looks pretty good, thanks for the updates!

The confusion matrix does look quite a bit worse than pre-changes though; can this be entirely attributed to the use of a test set? Maybe we need a bit more time for training to converge?

mgs28:
@jbschlosser  - thanks! 

1. The training converges around 35-40 iterations. I am happy to train until convergence or leave some room for the reader. I also left it lighter so that build time is faster for the CI bot. What would you prefer? 

2. Comparisons with the original confusion matrix: It's train/test split + training time. If I train with 

all_losses = train(rnn, alldata, n_epoch=100, learning_rate=0.1, report_every=5) 

and evaluate on alldata then I get a bright diagonal line that looks pretty similar to original. I imagine with some parameter tuning then I could get closer.

jbschlosser:
> I also left it lighter so that build time is faster for the CI bot. What would you prefer?

It's a good point that we want to balance this some. That said, I think it'd be nice to be a little bit closer to the original (at least somewhat of a diagonal line confusion matrix). Hopefully we can strike a reasonable balance where we're beginning to see a diagonal line trend. No need to spend a ton of time parameter tuning though :) that's okay left as an exercise to the reader.

mgs28:
No problem @jbschlosser  - I tuned some of the parameters to get pretty close to a diagonal confusion matrix. It still gets a little confused between English and Scottish as well as Korean and Chinese. However, there's a strong diagonal in the confusion matrix. Thanks!

jbschlosser:
Thanks for the updates! I'm good with it; will let @svekars make the final approval :)

mgs28:
@svekars  - do you have any further comments on this update? I'm happy to address. Thanks!

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -164,9 +126,87 @@ def lineToTensor(line):
...
+print (f"The name 'Ahn' becomes {lineToTensor('Ahn')}") #notice 'A' sets the 27th index to 1
 
-print(lineToTensor('Jones').size())
+#########################
+# Congratulations, you have built the foundational tensor objects for this learning task! You can use a similar approach 
+# for other RNN tasks with text.
+#
+# Next, we need to combine all our examples into a dataset so we can train, text and validate our models. For this, 
+# we will use the `Dataset and DataLoader <https://pytorch.org/tutorials/beginner/basics/data_tutorial.html>` classes 
+# to hold our dataset. Each Dataset needs to implement three functions: __init__, __len__, and __getitem__.
```
svekars, line 141:
```suggestion
# Next, we need to combine all our examples into a dataset so we can train text and validate our models. For this, 
# we will use the `Dataset and DataLoader <https://pytorch.org/tutorials/beginner/basics/data_tutorial.html>`__ classes 
# to hold our dataset. Each Dataset needs to implement three functions: ``__init__``, ``__len__``, and ``__getitem__``. 
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -164,9 +126,87 @@ def lineToTensor(line):
...
+        data_item = self.data[idx]
+        data_label = self.labels[idx]
+        data_tensor = self.data_tensors[idx]
+        label_tensor = self.labels_tensors[idx] 
+
+        return label_tensor, data_tensor, data_label, data_item
+
+
+#########################
+#Here we can load our example data into the NamesDataset
```
svekars, line 192:
```suggestion
#Here we can load our example data into the ``NamesDataset``:
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -164,9 +126,87 @@ def lineToTensor(line):
...
+
+alldata = NamesDataset("data/names")
+print(f"loaded {len(alldata)} items of data")
+print(f"example = {alldata[0]}")
+
+#########################
+#Using the dataset object allows us to easily split the data into train and test sets. Here we create a 80/20 
+#split but the torch.utils.data has more useful utilities. Here we specify a generator since we need to use the
```
svekars, line 200: Can you add a space between the initial # sign and the first letter of the character for consistency with other tutorials?
```suggestion
# split but the ``torch.utils.data`` has more useful utilities. Here we specify a generator since we need to use the 
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -164,9 +126,87 @@ def lineToTensor(line):
...
+#Here we can load our example data into the NamesDataset
+
+alldata = NamesDataset("data/names")
+print(f"loaded {len(alldata)} items of data")
+print(f"example = {alldata[0]}")
+
+#########################
+#Using the dataset object allows us to easily split the data into train and test sets. Here we create a 80/20 
+#split but the torch.utils.data has more useful utilities. Here we specify a generator since we need to use the 
+#same device as torch defaults to above.
```
svekars, line 201:
```suggestion
#same device as PyTorch defaults to above. 
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -164,9 +126,87 @@ def lineToTensor(line):
...
+#Using the dataset object allows us to easily split the data into train and test sets. Here we create a 80/20 
+#split but the torch.utils.data has more useful utilities. Here we specify a generator since we need to use the 
+#same device as torch defaults to above. 
+
+train_set, test_set = torch.utils.data.random_split(alldata, [.85, .15], generator=torch.Generator(device=device).manual_seed(2024))
+
+print(f"train examples = {len(train_set)}, validation examples = {len(test_set)}")
+
+#########################
+#Now we have a basic dataset containing 20074 examples where each example is a pairing of label and name. We have also
```
svekars, line 208:
```suggestion
# Now we have a basic dataset containing **20074** examples where each example is a pairing of label and name. We have also 
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -178,114 +218,58 @@ def lineToTensor(line):
 # held hidden state and gradients which are now entirely handled by the
 # graph itself. This means you can implement a RNN in a very "pure" way,
 # as regular feed-forward layers.
-#
-# This RNN module implements a "vanilla RNN" an is just 3 linear layers 
-# which operate on an input and hidden state, with a ``LogSoftmax`` layer 
-# after the output.
+# 
+# This CharRNN class implements an RNN with three components. 
+# First, we use the `nn.RNN implementation <https://pytorch.org/docs/stable/generated/torch.nn.RNN.html>`__
```
svekars, line 223:
```suggestion
# First, we use the `nn.RNN implementation <https://pytorch.org/docs/stable/generated/torch.nn.RNN.html>`__.
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -178,114 +218,58 @@ def lineToTensor(line):
...
 # graph itself. This means you can implement a RNN in a very "pure" way,
 # as regular feed-forward layers.
-#
-# This RNN module implements a "vanilla RNN" an is just 3 linear layers 
-# which operate on an input and hidden state, with a ``LogSoftmax`` layer 
-# after the output.
+# 
+# This CharRNN class implements an RNN with three components. 
+# First, we use the `nn.RNN implementation <https://pytorch.org/docs/stable/generated/torch.nn.RNN.html>`__ 
+# , next we define a layer that maps the RNN hidden layers to our output and finally we apply a softmax. Using nn.RNN
```
svekars, line 224:
```suggestion
# Next, we define a layer that maps the RNN hidden layers to our output. And finally, we apply a ``softmax`` function. Using ``nn.RNN``
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -178,114 +218,58 @@ def lineToTensor(line):
...
 # as regular feed-forward layers.
-#
-# This RNN module implements a "vanilla RNN" an is just 3 linear layers 
-# which operate on an input and hidden state, with a ``LogSoftmax`` layer 
-# after the output.
+# 
+# This CharRNN class implements an RNN with three components. 
+# First, we use the `nn.RNN implementation <https://pytorch.org/docs/stable/generated/torch.nn.RNN.html>`__ 
+# , next we define a layer that maps the RNN hidden layers to our output and finally we apply a softmax. Using nn.RNN 
+# leads to a significant improvement in performance (e.g. cuDNN-accelerated kernals) versus implementing
```
svekars, line 225:
```suggestion
# leads to a significant improvement in performance, such as cuDNN-accelerated kernels, versus implementing 
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -178,114 +218,58 @@ def lineToTensor(line):
...
-#
-# This RNN module implements a "vanilla RNN" an is just 3 linear layers 
-# which operate on an input and hidden state, with a ``LogSoftmax`` layer 
-# after the output.
+# 
+# This CharRNN class implements an RNN with three components. 
+# First, we use the `nn.RNN implementation <https://pytorch.org/docs/stable/generated/torch.nn.RNN.html>`__ 
+# , next we define a layer that maps the RNN hidden layers to our output and finally we apply a softmax. Using nn.RNN 
+# leads to a significant improvement in performance (e.g. cuDNN-accelerated kernals) versus implementing 
+# each layer as a nn.Linear. It also simplifies the implementation in forward().
```
svekars, line 226:
```suggestion
# each layer as a ``nn.Linear``. It also simplifies the implementation in ``forward()``.
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -178,114 +218,58 @@ def lineToTensor(line):
...
-# each language) and a next hidden state (which we keep for the next
-# step).
-#
 
-input = letterToTensor('A')
-hidden = torch.zeros(1, n_hidden)
-
-output, next_hidden = rnn(input, hidden)
+###########################
+#We can then create a RNN with 57 input nodes, 128 hidden nodes and 18 outputs.
```
svekars, line 249:
```suggestion
# We can then create an RNN with 57 input nodes, 128 hidden nodes, and 18 outputs:
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -178,114 +218,58 @@ def lineToTensor(line):
...
-#
-
-input = lineToTensor('Albert')
-hidden = torch.zeros(1, n_hidden)
-
-output, next_hidden = rnn(input[0], hidden)
-print(output)
-
+# We can then pass our Tensor to the RNN to get a predicted output and 
+# use a helper function, label_from_output, to get a text label for the class.
```
svekars, line 258:
```suggestion
# After that we can pass our Tensor to the RNN to obtain a predicted output. Subsequently,  
# we use a helper function, ``label_from_output``, to derive a text label for the class.

```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -295,93 +279,66 @@ def randomTrainingExample():
...
-n_iters = 100000
-print_every = 5000
-plot_every = 1000
-
-
-
-# Keep track of losses for plotting
-current_loss = 0
-all_losses = []
+# We do this by defining a train() function which trains on a given dataset with minibatches. RNNs
```
svekars, line 282:
```suggestion
# We do this by defining a ``train()`` function which trains the model on a given dataset using minibatches. RNNs 
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -295,93 +279,66 @@ def randomTrainingExample():
...
-print_every = 5000
-plot_every = 1000
-
-
-
-# Keep track of losses for plotting
-current_loss = 0
-all_losses = []
+# We do this by defining a train() function which trains on a given dataset with minibatches. RNNs 
+# train similar to other networks so for completeness we include a batched training method here.
```
svekars, line 283:
```suggestion
# RNNs are trained similarly to other networks; therefore, for completeness, we include a batched training method here.
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -295,93 +279,66 @@ def randomTrainingExample():
...
+# The loop (for i in batch) computes the losses for each of the items in the batch before adjusting the
```
svekars, line 284:
```suggestion
# The loop (``for i in batch``) computes the losses for each of the items in the batch before adjusting the 
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -295,93 +279,66 @@ def randomTrainingExample():
...
-
-
-
-# Keep track of losses for plotting
-current_loss = 0
-all_losses = []
+# We do this by defining a train() function which trains on a given dataset with minibatches. RNNs 
+# train similar to other networks so for completeness we include a batched training method here.
+# The loop (for i in batch) computes the losses for each of the items in the batch before adjusting the 
+# weights. This is repeated until the number of epochs is reached.
```
svekars, line 285:
```suggestion
# weights. This operation is repeated until the number of epochs is reached. 
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -295,93 +279,66 @@ def randomTrainingExample():
...
+    return all_losses
 
-def timeSince(since):
-    now = time.time()
-    s = now - since
-    m = math.floor(s / 60)
-    s -= m * 60
-    return '%dm %ds' % (m, s)
+##########################################################################
+# We can now train a dataset with mini batches for a specified number of epochs
```
svekars, line 336: mini batches or minibatches? I see all three versions online: mini batch, mini-batch, minibatch. It seems like it should be one word and I'd rather not use hyphens, so "minibatch" maybe a better way. 
```suggestion
# We can now train a dataset with minibatches for a specified number of epochs
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -460,72 +414,20 @@ def evaluate(line_tensor):
...
-#
-#    -  Any word -> language
-#    -  First name -> gender
-#    -  Character name -> writer
-#    -  Page title -> blog or subreddit
-#
 # -  Get better results with a bigger and/or better shaped network
 #
-#    -  Add more linear layers
+#    -  Vary the hyperparameters to improve performance (e.g. change epochs, batch size, learning rate )
```
svekars, line 423:
```suggestion
#    -  Adjust the hyperparameters to enhance performance, such as changing the number of epochs, batch size, and learning rate
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -460,72 +414,20 @@ def evaluate(line_tensor):
...
-#    -  First name -> gender
-#    -  Character name -> writer
-#    -  Page title -> blog or subreddit
-#
 # -  Get better results with a bigger and/or better shaped network
 #
-#    -  Add more linear layers
+#    -  Vary the hyperparameters to improve performance (e.g. change epochs, batch size, learning rate ) 
 #    -  Try the ``nn.LSTM`` and ``nn.GRU`` layers
+#    -  Change the size of the layers (e.g. fewer or more hidden nodes, additional linear layers)
```
svekars, line 425:
```suggestion
#    -  Modify the size of the layers, such as increasing or decreasing the number of hidden nodes or adding additional linear layers
```

mgs28:
@svekars  - Sorry on two accounts: 

1. Sorry for the delay... Some personal stuff come up and tonight was the earliest I could address. 
2. I made review of your editing changes harder. I manually made some of your changes via a commit and accepted some here in the UI. I apologize for that. I was not thinking tonight. 

I made one departure from your edits. You suggested removing a comma in "train text and validate our models" but when I re-read it, I actually meant "train, test and validate our models". I updated the text to read as such. 

I apologize for my delay and muddying up this review but greatly appreciate your feedback on this. All your points were great. Thanks!

mgs28:
@svekars  - Happy Tuesday. Is there anything else I need to do to help merge this tutorial in? 

Thanks!

svekars:
Looks good - would be great to also fix these user warnings:

Screenshot 2024-12-03 at 12 13 09 PM

mgs28:
@svekars  - Fixed it.  Sorry I missed that warning! Thanks

svekars (approved):

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -169,9 +131,87 @@ def lineToTensor(line):
         tensor[li][0][letterToIndex(letter)] = 1
     return tensor
 
-print(letterToTensor('J'))
+#########################
+# Here are some examples of how to use lineToTensor() for a single and multiple character string.
```
svekars, line 135:
```suggestion
# Here are some examples of how to use ``lineToTensor()`` for a single and multiple character string.
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -169,9 +131,87 @@ def lineToTensor(line):
...
+        self.data_dir = data_dir #for provenance of the dataset
+        self.load_time = time.localtime #for provenance of the dataset 
+        labels_set = set() #set of all classes
+
+        self.data = []
+        self.data_tensors = []
+        self.labels = [] 
+        self.labels_tensors = [] 
+
+        #read all the txt files in the specified directory
```
svekars, line 167:
```suggestion
        #read all the ``.txt`` files in the specified directory
```

intermediate_source/char_rnn_classification_tutorial.py
```diff
@@ -183,114 +223,58 @@ def lineToTensor(line):
...
-# As you can see the output is a ``<1 x n_categories>`` Tensor, where
-# every item is the likelihood of that category (higher is more likely).
-#
+def label_from_output(output, output_labels):
+    top_n, top_i = output.topk(1)
+    label_i = top_i[0].item()
+    return output_labels[label_i], label_i
 
+input = lineToTensor('Albert')
+output = rnn(input) #this is equivalent to output = rnn.forward(input)
```
svekars, line 270:
```suggestion
output = rnn(input) #this is equivalent to ``output = rnn.forward(input)``
```

svekars:
@mgs28 I really want to merge this but I wonder what happened in the update that increased the time of the build for this tutorial from [~6~ minutes](https://pytorch.org/tutorials/intermediate/char_rnn_classification_tutorial.html#exercises) to [22 minutes](https://docs-preview.pytorch.org/pytorch/tutorials/2954/intermediate/char_rnn_classification_tutorial.html#exercises)

mgs28:
@svekars  - I also want you to merge it in. :) 

**History:** The training is different (batches instead of random samples) and if you look at the comments around July 11th, we tinkered with the parameters to make the confusion matrix better.  

My best suggestion is to set device = cpu at the beginning of the script. With 10% of my laptop CPU, it takes about 4 minutes. Can you share some of the specs of the deployment machine? That might give me a bit more to dig into.  Thanks!

svekars:
> @svekars - I also want you to merge it in. :)
> 
> **History:** The training is different (batches instead of random samples) and if you look at the comments around July 11th, we tinkered with the parameters to make the confusion matrix better.
> 
> My best suggestion is to set device = cpu at the beginning of the script. With 10% of my laptop CPU, it takes about 4 minutes. Can you share some of the specs of the deployment machine? That might give me a bit more to dig into. Thanks!

It's a 1 GPU machine. Can you try with device = cpu?

mgs28:
@svekars  - good idea. I removed the device options at the top of the script. pyspelling passed and the build html looks good. Please let me know if I missed something. Thanks!

svekars:
well, now its[43 minutes](https://docs-preview.pytorch.org/pytorch/tutorials/2954/intermediate/char_rnn_classification_tutorial.html#exercises)...

mgs28:
@svekars  - well... clearly that's the wrong direction... I reverted the changes. 

What's your goal? 

In July and more recently, I compared the old code to this new addition. 
- I added an optimizer (SGD) instead of manually updating weights in the prior tutorial
- I am using datasets (original reason for the update) instead of randomly selecting 100,000 examples
- Batches require training over 9.8x data (55 iterations through 17,063 points = 938,465 examples). We're about 3x faster per example but training more examples. 
- The original article tests on the training data and we changed to withhold a separate dataset. This makes the performance look worse so we increased training time.  

All of these changes make this tutorial more in line with best practices. To make the time goal, I can reduce training by a percentage. 

* Original training (estimate 22 minutes)
accuracy = 0.796745240688324
confusion matrix has a strong diagonal

* Half training (estimate 11 minutes)
accuracy = 0.7974094748497009
confusion matrix has a less strong diagonal

* Quarter training (estimate 5 minutes)
accuracy = 0.8003985285758972
confusion matrix has no diagonal 

I checked in the 11 minute version for you which I think is a good compromise. Do you agree?

svekars:
Agree on the 11 minutes compromise - thank you for adjusting! Please add a note that for the purpose of example the number of epochs is reduced and we recommend to set it to the a larger number for higher level of accuracy.

mgs28:
@svekars  - good idea and done. I added a comment where training is specified. The first item in the end exercises also mentions hyperparameter tuning as a good place to start. Does that work? Thanks!

PR #2959, opened 2024-06-28
Updated linalg functions in tensors_deeper_tutorial.py

### LINKED ISSUE #2933: [BUG] - UserWarning:  Using torch.cross without specifying the dim arg is deprecated.
### Add Link

https://pytorch.org/tutorials/beginner/introyt/tensors_deeper_tutorial.html

### Describe the bug

image

```
/var/lib/workspace/beginner_source/introyt/tensors_deeper_tutorial.py:462: UserWarning:

Using torch.cross without specifying the dim arg is deprecated.
Please either pass the dim explicitly or simply use torch.linalg.cross.
The default value of dim will change to agree with that of linalg.cross in a future release. (Triggered internally at ../aten/src/ATen/native/Cross.cpp:62.)
```

### Describe your environment

I don't actually run it but find the web page has such a warning.

ignaciobartol:
Fixes #2933

## Description
Hi, after reading #2933, there are a few things that I noted while attempting to solve this issue:
1. While `torch.cross` is more flexible and allows you for more complex cross products, I believe the expected behavior is usually the one from `torch.linalg.cross` implementation. In [this discussion](https://github.com/pytorch/pytorch/issues/62810) it explains why *blindly* using `torch.cross` can lead you to an intersting bug-hunting. 
2. The function `torch.svd` is deprecated and will be removed in future pytorch release. 
3. For completion and to avoid beginner user confusion, I also changed to use `torch.linalg.matmul` which actually calls `torch.matmul`, but I think it would be better from a beginner user standpoint, to refer them to `torch.linalg` for all linear algebra related operations. 
4. It would be nice to add maybe a tag/flag in the youtube video [here](https://youtu.be/r7QDUPb2dCM?t=1019) mentioning the updates to linear algebra operations.  
5. Added ref to `torch.linalg` docs. 

## Checklist

- [x] The issue that is being fixed is referred in the description

cc @albanD


albanD (approved):
Looks great and thank you for the detailed PR description!

PR #2963, opened 2024-07-08
Update data_loading_tutorial.py, correct iterator vs iterable

martin-kokos:
DataLoader doesn't implement `__next__()` therefore it can't be in iterator

## Description
```python
    dl = torch.utils.data.DataLoader(dataset, batch_size=2)
    X, y = next(dl)
           ^^^^^^^^
TypeError: 'DataLoader' object is not an iterator
```
https://docs.python.org/3/glossary.html#term-iterator
https://docs.python.org/3/glossary.html#term-iterable

## Checklist

cc @svekars @brycebortree @albanD


martin-kokos:
I am not going to sign the CLA. Do what you will.

PR #2972, opened 2024-07-15
Fix reference to dcp in loading example

dabauxi:
## Description

Imported as `DCP` but used as `dcp`. Changed import to `dcp` to have same import style as saving example.

## Checklist


wz337 (approved):
lgtm

PR #2983, opened 2024-07-24
A tutorial on pin_memory and non_blocking usage

vmoens:
This is a draft PR on the proper usage of `pin_memory` and `non_blocking` when sending data from CPU to GPU (and GPU to CPU).

## Description

There is some confusion about the proper use of pin_memory and non_blocking in the community.
To be convinced about it, one can simply look for the [`.pin_memory.to(`](https://github.com/search?q=%22.pin_memory%28%29.to%28%22&type=code) pattern on github (about 2k results), when this command is always slower than a simple call to `to(device)`.

The responsibility lies in part in the pytorch doc itself where it is recommended (implicitly) to call `pin_memory` [before calling `to`](https://pytorch.org/docs/stable/generated/torch.Tensor.to.html) with `non_blocking=True` and several such occurrences on the pytorch forum too.

Some refs on the topic:
https://discuss.pytorch.org/t/what-is-the-disadvantage-of-using-pin-memory/1702/13
https://discuss.pytorch.org/t/non-blocking-memory-transfer-to-gpu/188941
https://discuss.pytorch.org/t/should-we-set-non-blocking-to-true/38234
https://discuss.pytorch.org/t/how-is-explicit-pin-memory-different-from-just-calling-to-and-let-cuda-handle-it/197422

I use TensorDict to demonstrate how to use pin_memory across threads.

A follow-up will be to link back the pytorch doc / docstrings of to / pin_memory etc to this tutorial.

The syntax should be fixed and integrated better in the lib. Conclusion and additional resources are missing.
Still, feedback is more than welcome!

The tutorial requires tensordict v0.5 which will be released in the coming days.

cc @shagunsodhani @albanD @janeyx99 @dstaay-fb @mikaylagawarecki  @ptrblck


intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,429 @@
+# -*- coding: utf-8 -*-
+"""
+A guide on good usage of `non_blocking` and `pin_memory()` in PyTorch
```
svekars, line 3: We need to use .rst syntax:
- double backticks instead of single ones
- Links should be `Text <link>`_ not []()

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,429 @@
...
+=====================================================================
+
+TL;DR
+-----
```
svekars, line 7: I suggest we remove the TL;DR - the text below can be just the intro abstract right after the title. Can you follow this template: https://github.com/pytorch/tutorials/blob/main/beginner_source/template_tutorial.py  
Add What you will learn and Prerequisites and Author

intermediate_source/pinmem_nonblock.py
vmoens: Good point, I was planning on doing that - currently the script is a mere export from a regular ipynb, I should make a second pass and correct all the links etc!

intermediate_source/pinmem_nonblock.py
vmoens: Sure!

vmoens:
@svekars Happy to get some more feedback.
On my side, I only see these items to be done:
- edit figures to make them look prettier
- add additional info (links etc)
- check that we have the most demonstrative examples (ie play a bit with the setups to highlight the differences better)

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+# make up the virtual memory, which is an abstraction of the total resources available.
+# In short, the virtual memory makes it so that the available space is larger than what can be found on RAM in isolation
+# and creates the illusion that the main memory is larger than it actually is.
+#
+# In normal circumstances, a regular CPU tensor is _paged_, which means that it is divided in blocks called _pages_ that
+# can live anywhere in the virtual memory (both in RAM or on disk). As mentioned earlier, this has the advantage that
+# the memory seems larger than what the main memory actually is.
+#
+# Typically, when a program accesses a page that is not in RAM, a "page fault" occurs and the operating system (OS) then brings
+# back this page into RAM (_swap in_ or _page in_).
```
mikaylagawarecki, line 79: nit: all these `_bla bla_` seem to be rendered as links

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+# .. figure:: /_static/img/pinmem.png
+#    :alt:
+#
+# CUDA and (non-)pageable memory
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   .. _pinmem_cuda_pageable_mem:
+#
+# To understand how CUDA copies a tensor from CPU to CUDA, let's consider the two scenarios above:
+# - If the memory is page-locked, the device can access the memory directly in the main memory. The memory addresses are well
```
mikaylagawarecki, line 96: nit: This bullet point is not rendering as a bullet point

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+#
+# In practice, when calling :meth:`~torch.Tensor.to`, PyTorch always makes a call to
+# `cudaMemcpyAsync <https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g85073372f776b4c4d5f89f7124b7bf79>`_.
+# If ``non_blocking=False`` (default), a ``cudaStreamSynchronize`` will be called after each and every ``cudaMemcpyAsync``, making
+# the call to :meth:`~torch.Tensor.to` blocking in the main thread.
+# If ``non_blocking=True``, no synchronization is triggered, and the main thread on the host is not blocked.
+# Therefore, from the host perspective, multiple tensors can be sent to the device simultaneously,
+# as the thread does not need to wait for one transfer to be completed to initiate the other.
+#
+# .. note:: In general, the transfer is blocking on the device side (even if it isn't on the host side):
```
mikaylagawarecki, line 120: n00b qn: tbh this is something that I have always been confused about, but I can't seem to get overlap for copies.

I tried this with the following script on A100 devgpu but I can't seem to get overlap, do you know what might be wrong here?
Screenshot 2024-07-26 at 12 37 45 PM



<details closed>
<summary>Click for Script</summary>
<br>

```python

import torch
from torch.cuda import Stream


def benchmark_with_profiler(
    benchmark_fn,
    *benchmark_fn_args,
    **benchmark_fn_kwargs,
) -> None:
    torch._C._profiler._set_cuda_sync_enabled_val(True)
    wait, warmup, active = 1, 1, 2
    num_steps = wait + warmup + active
    rank = 0
    with torch.profiler.profile(
        activities=[
            torch.profiler.ProfilerActivity.CPU,
            torch.profiler.ProfilerActivity.CUDA,
        ],
        schedule=torch.profiler.schedule(
            wait=wait, warmup=warmup, active=active, repeat=1, skip_first=1
        ),
        on_trace_ready=torch.profiler.tensorboard_trace_handler("./")
        if not rank
        else None,
        record_shapes=True,
        profile_memory=True,
        with_stack=False,
        with_flops=True,
        with_modules=False,
    ) as prof:
        for step_idx in range(1, num_steps + 1):
            benchmark_fn(*benchmark_fn_args, **benchmark_fn_kwargs)
            if rank is None or rank == 0:
                prof.step()

s = Stream()

torch.manual_seed(42)
t1_cpu = torch.randn(1024 ** 2 * 5, pin_memory=True)
t2_cpu = torch.randn(1024 ** 2 * 5, pin_memory=True)

assert torch.cuda.is_available()
device = torch.device("cuda", torch.cuda.current_device())

def inner():
    with torch.cuda.stream(s):
        t2_cuda = t2_cpu.to(device, non_blocking=True)
        t2_h2d_event = s.record_event()
    t1_cuda = t1_cpu.to(device, non_blocking=True)
    t1_h2d_event = torch.cuda.current_stream().record_event()
    t1_h2d_event.synchronize()
    t2_h2d_event.synchronize()

benchmark_with_profiler(inner)

```

</details>

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+
+
+######################################################################
+# Let's see the call stack with a regular ``to(device)`` first:
+#
+
+print("Call to `to(device)`", profile_mem("copy_to_device(*tensors)"))
+
+######################################################################
+# and now the ``non_blocing`` version:
```
mikaylagawarecki, line 302: nit:
```suggestion
# and now the ``non_blocking`` version:
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+
+######################################################################
+# The results are without any doubt better when using ``non_blocking=True``, as all transfers are initiated simultaneously
+# on the host side and only one synchronization is done.
+#
+# The benefit will vary depending on the number and the size of the tensors as well as depending on the hardware being
+# used.
+#
+# .. note:: Interestingly, the blocking ``to("cuda")`` actually performs the same asynchronous device casting operation
+#   (``cudaMemcpyAsync``) as the one with ```non_blocking=True`` with a synchronization point after each copy.
```
mikaylagawarecki, line 319: nit: 
```suggestion
#   (``cudaMemcpyAsync``) as the one with ``non_blocking=True`` with a synchronization point after each copy.
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+#   How is the system's architecture influencing data transfer speeds (e.g., bus speeds, network latency)?
+#
+# Additionally, allocating a large number of tensors or sizable tensors in pinned memory can monopolize a substantial
+# portion of RAM.
+# This reduces the available memory for other critical operations, such as paging, which can negatively impact the
+# overall performance of an algorithm.
+#
+
+######################################################################
+# Conclusion
```
mikaylagawarecki, line 601: Nit: this and additional resources are not rendering

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+#
+#   .. _pinmem_cuda_pageable_mem:
+#
+# To understand how CUDA copies a tensor from CPU to CUDA, let's consider the two scenarios above:
+# - If the memory is page-locked, the device can access the memory directly in the main memory. The memory addresses are well
+#   defined and functions that need to read these data can be significantly accelerated.
+# - If the memory is pageable, all the pages will have to be brought to the main memory before being sent to the GPU.
+#   This operation may take time and is less predictable than when executed on page-locked tensors.
+#
+# More precisely, when CUDA sends pageable data from CPU to GPU, it must first create a page-locked copy of that data
```
janeyx99, line 101: and "page-locked" is the same thing as "pinned"?

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+# If ``non_blocking=False`` (default), a ``cudaStreamSynchronize`` will be called after each and every ``cudaMemcpyAsync``, making
+# the call to :meth:`~torch.Tensor.to` blocking in the main thread.
+# If ``non_blocking=True``, no synchronization is triggered, and the main thread on the host is not blocked.
+# Therefore, from the host perspective, multiple tensors can be sent to the device simultaneously,
+# as the thread does not need to wait for one transfer to be completed to initiate the other.
+#
+# .. note:: In general, the transfer is blocking on the device side (even if it isn't on the host side):
+#   the copy on the device cannot occur while another operation is being executed.
+#   However, in some advanced scenarios, multiple copies or copy and kernel
+#   executions can be done simultaneously on the GPU side. To enable this, three requirements must be met:
```
janeyx99, line 123: Is the following subject to change? / Maybe include the source/docs that support these facts if those exist?

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+# Typically, when a program accesses a page that is not in RAM, a "page fault" occurs and the operating system (OS) then brings
+# back this page into RAM (_swap in_ or _page in_).
+# In turn, the OS may have to _swap out_ (or _page out_) another page to make room for the new page.
+#
+# In contrast to pageable memory, a _pinned_ (or _page-locked_ or _non-pageable_) memory is a type of memory that cannot
+# be swapped out to disk.
+# It allows for faster and more predictable access times, but has the downside that it is more limited than the
+# pageable memory (aka the main memory).
+#
+# .. figure:: /_static/img/pinmem.png
```
janeyx99, line 87: this diagram is epic.

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+#   .. _pinmem_pt_perspective:
+#
+# ``pin_memory()``
+# ~~~~~~~~~~~~~~~~
+#
+#   .. _pinmem_pinmem:
+#
+# PyTorch offers the possibility to create and send tensors to page-locked memory through the
+# :meth:`~torch.Tensor.pin_memory` method and constructor arguments.
+# Cpu tensors on a machine where a cuda is initialized can be cast to pinned memory through the :meth:`~torch.Tensor.pin_memory`
```
janeyx99, line 146:
```suggestion
# CPU tensors on a machine where CUDA is initialized can be cast to pinned memory through the :meth:`~torch.Tensor.pin_memory`
```
nit

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+ax.set_title("Runtime (pin-mem and non-blocking)")
+ax.set_xticks([0, 1, 2])
+ax.set_xticklabels(strategies)
+plt.setp(ax.get_xticklabels(), rotation=45, ha="right", rotation_mode="anchor")
+ax.legend(loc="upper left", ncols=3)
+
+plt.show()
+
+del tensors, tensors_pinned
+_ = gc.collect()
```
janeyx99, line 397: The graph that gets displayed at this point is interesting! To me, it basically says:
- copying from pinned and pageable are both equally fast, as pinning doesn't save much time at all
- pinning and copying is a bad strat and should be avoided, because it would already happen under the hood.

I think it'd be good to have a paragraph summarizing the point of your benchmarks here. It is compelling!

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+
+
+######################################################################
+# Other copy directions (GPU -> CPU, CPU -> MPS etc.)
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   .. _pinmem_otherdir:
+#
+# Until now, we have operated under the assumption that asynchronous copies from the CPU to the GPU are safe.
+# This is generally true because CUDA automatically handles synchronization to ensure that the data being accessed is
```
janeyx99, line 407: Is this true because of streams? Like stream ordering will automatically enforce this determinism?

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+try:
+    i = -1
+    for i in range(100):
+        cpu_tensor = tensor.to("cpu", non_blocking=True)
+        torch.testing.assert_close(
+            cpu_tensor.mean(), torch.tensor(500_000, dtype=torch.double)
+        )
+    print("No test failed with non_blocking")
+except AssertionError:
+    print(f"One test failed with non_blocking: {i}th assertion!")
```
janeyx99, line 432: Nit:
```suggestion
    print(f"Some test failed with non_blocking: {i}th assertion!")
```

One implies that only one failed in my head. This is a minor nit and could just be me though

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+# the data is accessed.
+#
+# Practical recommendations
+# -------------------------
+#
+#   .. _pinmem_recom:
+#
+# We can now wrap up some early recommendations based on our observations:
+#
+# In general, ``non_blocking=True`` will provide a good throughput, regardless of whether the original tensor is or
```
janeyx99, line 462:
```suggestion
# In general, ``non_blocking=True`` will provide good throughput, regardless of whether the original tensor is or
```
this is phrased a tad weirdly. Maybe "will enable good throughput"?

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+#
+# One might now legitimately ask what use there is for the :meth:`~torch.Tensor.pin_memory` method.
+# In the following section, we will explore further how this can be used to accelerate the data transfer even more.
+#
+# Additional considerations
+# -------------------------
+#
+#   .. _pinmem_considerations:
+#
+# PyTorch notoriously provides a :class:`~torch.utils.data.DataLoader` class which constructor accepts a
```
janeyx99, line 476:
```suggestion
# PyTorch notoriously provides a :class:`~torch.utils.data.DataLoader` class whose constructor accepts a
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+# accelerate data transfers if memory pinning is inherently blocking.
+#
+# The key lies in the DataLoader's use of a separate thread to handle the transfer of data from pageable to pinned
+# memory, thus preventing any blockage in the main thread.
+#
+# To illustrate this, we will use the TensorDict primitive from the homonymous library.
+# When invoking :meth:`~tensordict.TensorDict.to`, the default behavior is to send tensors to the device asynchronously,
+# followed by a single call to ``torch.device.synchronize()`` afterwards.
+#
+# Additionally, ``TensorDict.to()`` includes a ``non_blocking_pin`` option  which initiates multiple threads to execute
```
janeyx99, line 488: so is the tl;dr "pin_memory is useless unless you can put it on a diff thread"?

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+colors = ["tab:blue", "tab:red", "tab:orange", "tab:green"]
+
+ax.bar(xlabels, values, label=bar_labels, color=colors)
+
+ax.set_ylabel("Runtime (ms)")
+ax.set_title("Device casting runtime")
+ax.set_xticks([])
+ax.legend()
+
+plt.show()
```
janeyx99, line 536: from this graph, it _still_ looks like pinning memory is useless because the red bar (the nonblock copy) is just as fast! so why would anyone ever use pin_memory still?

janeyx99 (commented):
super informative!

intermediate_source/pinmem_nonblock.py
shagunsodhani: I would probably make it even more explicit: `print(f"{i}th test failed with non_blocking. Skipping remaining tests")`

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+# followed by a single call to ``torch.device.synchronize()`` afterwards.
+#
+# Additionally, ``TensorDict.to()`` includes a ``non_blocking_pin`` option  which initiates multiple threads to execute
+# ``pin_memory()`` before proceeding with to ``to(device)``.
+# This approach can further accelerate data transfers, as demonstrated in the following example.
+#
+# .. code-block:: bash
+#
+#    # Install tensordict with the following command
+#    !pip3 install https://github.com/pytorch/tensordict
```
shagunsodhani, line 495: Tried running this on google colab and ran into the following

```
Collecting https://github.com/pytorch/tensordict
  Downloading https://github.com/pytorch/tensordict
     / 543.5 kB 2.0 MB/s 0:00:00
  ERROR: Cannot unpack file /tmp/pip-unpack-3xp1u__v/tensordict (downloaded from /tmp/pip-req-build-ifldo3u0, content-type: text/html; charset=utf-8); cannot detect archive format
ERROR: Cannot determine archive format of /tmp/pip-req-build-ifldo3u0
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+#
+
+from tensordict import TensorDict
+import torch
+from torch.utils.benchmark import Timer
+
```
shagunsodhani, line 501:
```suggestion
import matplotlib.pyplot as plt
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+#
+#   - :ref:`Memory management basics <pinmem_mem>`
+#   - :ref:`CUDA and (non-)pageable memory <pinmem_cuda_pageable_mem>`
+#   - :ref:`Asynchronous vs. Synchronous Operations with non_blocking=True <pinmem_async_sync>`
+#
+# - :ref:`A PyTorch perspective <pinmem_pt_perspective>`
+#
+#   - :ref:`pin_memory <pinmem_pinmem>`
+#   - :ref:`non_blocking=True <pinmem_nb>`
+#   - :ref:`Synergies <synergies>`
```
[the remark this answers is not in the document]
vmoens:
```suggestion
#   - :ref:`Synergies <pinmem_synergies>
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,625 @@
...
+#
+# Memory management basics
+# ~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   .. _pinmem_mem:
+#
+# When one creates a CPU tensor in PyTorch, the content of this tensor needs to be placed
+# in memory. The memory we talk about here is a rather complex concept worth looking at carefully.
+# We distinguish two types of memories that are handled by the Memory Management Unit: the main memory (for simplicity)
+# and the disk (which may or may not be the hard drive). Together, the available space in disk and RAM (physical memory)
```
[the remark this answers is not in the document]
vmoens:
```suggestion
# and the swap space on disk (which may or may not be the hard drive). Together, the available space in disk and RAM (physical memory)
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,429 @@
...
+# make up the virtual memory, which is an abstraction of the total resources available.
+# In short, the virtual memory makes it so that the available space is larger than what can be found on RAM in isolation
+# and creates the illusion that the main memory is larger than it actually is.
+#
+# In normal circumstances, a regular CPU tensor is _paged_, which means that it is divided in blocks called _pages_ that
```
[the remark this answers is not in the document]
vmoens:
```suggestion
# In normal circumstances, a regular CPU tensor is pageable which means that it is divided in blocks called pages that
```

intermediate_source/pinmem_nonblock.py
vmoens, line 79:
```suggestion
# back this page into RAM ("swap in" or "page in").
```

intermediate_source/pinmem_nonblock.py
vmoens:
```suggestion
#
# - If the memory is page-locked, the device can access the memory directly in the main memory. The memory addresses are well
```

intermediate_source/pinmem_nonblock.py
vmoens, line 432:
```suggestion
    print(f"{i}th test failed with non_blocking. Skipping remaining tests")
```

intermediate_source/pinmem_nonblock.py
vmoens: that's a problem with the CI I think, we have a limited number of cores there.
On depgpu it renders better. That being said I did not manage to find anything where pin_memory() brought a real non-trivial benefit on any machine...

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,626 @@
...
+#   How is the system's architecture influencing data transfer speeds (e.g., bus speeds, network latency)?
+#
+# Additionally, allocating a large number of tensors or sizable tensors in pinned memory can monopolize a substantial
+# portion of RAM.
+# This reduces the available memory for other critical operations, such as paging, which can negatively impact the
+# overall performance of an algorithm.
+#
+
+######################################################################
+# Conclusion
```
vmoens, line 602:
```suggestion
#
# Conclusion
```

intermediate_source/pinmem_nonblock.py
vmoens: I think so, @albanD can you confirm?

intermediate_source/pinmem_nonblock.py
vmoens: Got it, will do

intermediate_source/pinmem_nonblock.py
vmoens: That's my understanding!

intermediate_source/pinmem_nonblock.py
vmoens: Plus you can get some limited speed-up with limited threads on some machines, but at the cost of using more workers so your preproc/logging/whatever it is you use your CPUs for may become slower because of that. My take is that in general people should have a clear use case for pin-memory or not use it at all.

For instance, here's a scenario where I could consider pin-mem:
In distributed settings, you will sometimes pass a ton of tensors from worker 0 to worker 1, and worker 1 will send these tensors to CUDA.
With tensordict you could:
(1) consolidate all tensors in a single storage on worker 0
(2) send that single storage to worker 1 (faster than sending the tensors separately)
(3) receive that storage in pin-memory (not sure how to do that exactly, you'd need to hack the reductions methods appropriately, but I guess it's doable) - NOT receive in pageable mem and send to page-locked!
(4) send that to cuda

That will be much faster than sending all of the tensors, receiving them all and sending them to cuda with/without pin-mem.

@dstaay-fb

vmoens: (I wanted to call this tuto: don't use pin_memory but that was going a bit too far haha)

intermediate_source/pinmem_nonblock.py
vmoens: Seems like it's less about concurrent cpu->device copies and more about copy while doing another op.


<details closed>
<summary>Click for Script</summary>
<br>

```python
import torch
from torch.cuda import Stream


def benchmark_with_profiler(
    benchmark_fn,
    *benchmark_fn_args,
    **benchmark_fn_kwargs,
) -> None:
    torch._C._profiler._set_cuda_sync_enabled_val(True)
    wait, warmup, active = 1, 1, 2
    num_steps = wait + warmup + active
    rank = 0
    with torch.profiler.profile(
        activities=[
            torch.profiler.ProfilerActivity.CPU,
            torch.profiler.ProfilerActivity.CUDA,
        ],
        schedule=torch.profiler.schedule(
            wait=wait, warmup=warmup, active=active, repeat=1, skip_first=1
        ),
        on_trace_ready=torch.profiler.tensorboard_trace_handler("./")
        if not rank
        else None,
        record_shapes=True,
        profile_memory=True,
        with_stack=False,
        with_flops=True,
        with_modules=False,
    ) as prof:
        for step_idx in range(1, num_steps + 1):
            benchmark_fn(*benchmark_fn_args, **benchmark_fn_kwargs)
            if rank is None or rank == 0:
                prof.step()

s = Stream()

torch.manual_seed(42)
t1_cpu_pinned = torch.randn(1024 ** 2 * 5, pin_memory=True)
t2_cpu_pageable = torch.randn(1024 ** 2 * 5, pin_memory=False)
t3_cuda = torch.randn(1024 ** 2 * 5, device="cuda:0")

assert torch.cuda.is_available()
device = torch.device("cuda", torch.cuda.current_device())

def inner():
    with torch.cuda.stream(s): # Comment to get main stream
        t2_cuda = t2_cpu_pageable.to(device, non_blocking=True)  # comment and uncomment following for pinned
        # t1_cuda = t1_cpu_pinned.to(device, non_blocking=True)
        t2_h2d_event = s.record_event()
    t3_cuda_mul = t3_cuda * t3_cuda * t3_cuda
    t1_h2d_event = torch.cuda.current_stream().record_event()
    t1_h2d_event.synchronize()
    t2_h2d_event.synchronize()

benchmark_with_profiler(inner)
```
</details>

Playing with the following code (commenting/uncommenting things and executing in same stream or different):

1. non_blocking copy from pinned tensor on dedicated stream, followed by a multiplication on main stream
image
2. non_blocking copy from pinned tensor on main stream, followed by a multiplication on main stream
image
4. non_blocking copy from pageable tensor on dedicated stream, followed by a multiplication on main stream
image

I will edit the doc to say this:

If the tensor is in pinned mem, the device has a free DMA and the copy is done in a separate stream, CUDA can execute another op in the main stream concurrently to the copy - but not another copy.

Agreed? (maybe it could do another copy on a third stream??)

mikaylagawarecki: Yea I think the updated wording mostly makes sense except the last bit! Not sure whether we want to very conclusively say that another copy is not possible as I am not 100% sure whether there is absolutely no scenario under which it could do another copy as well 😅  (perhaps H2D and D2H could be overlapped??)

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,710 @@
...
+
+def inner(pinned: bool, streamed: bool):
+    with torch.cuda.stream(s) if streamed else contextlib.nullcontext():
+        if pinned:
+            t1_cuda = t1_cpu_pinned.to(device, non_blocking=True)
+        else:
+            t2_cuda = t2_cpu_paged.to(device, non_blocking=True)
+        t2_h2d_event = s.record_event()
+    # This operation can be executed during the CPU to GPU copy iff the tensor is pinned and the copy is
+    #  done in the other stream
```
mikaylagawarecki, line 160: kind of a nit, and also clarification on my end:

is it more like if the tensor was not pinned
- If the mul was huge enough, the mul could be overlapped once the internal copy from pageable --> pinned was done?

vmoens: you could overlap two mul on two streams if that's the question, but not a copy (from pageable) and a `mul`

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,728 @@
...
+# Therefore, from the host perspective, multiple tensors can be sent to the device simultaneously,
+# as the thread does not need to wait for one transfer to be completed to initiate the other.
+#
+# .. note:: In general, the transfer is blocking on the device side (even if it isn't on the host side):
+#   the copy on the device cannot occur while another operation is being executed.
+#   However, in some advanced scenarios, a copy and a kernel execution can be done simultaneously on the GPU side.
+#   As the following example will show, three requirements must be met to enable this:
+#
+#   1. The device must have at least one free DMA (Direct Memory Access) engine. Modern GPU architectures such as Volterra,
+#      Tesla or H100 devices have more than one DMA engine.
```
svekars, line 128:
```suggestion
#      Tesla, or H100 devices have more than one DMA engine.
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,728 @@
...
+#
+# PyTorch offers the possibility to create and send tensors to page-locked memory through the
+# :meth:`~torch.Tensor.pin_memory` method and constructor arguments.
+# CPU tensors on a machine where CUDA is initialized can be cast to pinned memory through the :meth:`~torch.Tensor.pin_memory`
+# method. Importantly, ``pin_memory`` is blocking on the main thread of the host: it will wait for the tensor to be copied to
+# page-locked memory before executing the next operation.
+# New tensors can be directly created in pinned memory with functions like :func:`~torch.zeros`, :func:`~torch.ones` and other
+# constructors.
+#
+# Let us check the speed of pinning memory and sending tensors to cuda:
```
svekars, line 253:
```suggestion
# Let us check the speed of pinning memory and sending tensors to CUDA:
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,728 @@
...
+_ = gc.collect()
+
+######################################################################
+#
+# We can observe that casting a pinned-memory tensor to GPU is indeed much faster than a pageable tensor, because under
+# the hood, a pageable tensor must be copied to pinned memory before being sent to GPU.
+#
+# However, contrary to a somewhat common belief, calling :meth:`~torch.Tensor.pin_memory()` on a pageable tensor before
+# casting it to GPU should not bring any significant speed-up, on the contrary this call is usually slower than just
+# executing the transfer. This makes sense, since we're actually asking python to execute an operation that CUDA will
```
svekars, line 321:
```suggestion
# executing the transfer. This makes sense, since we're actually asking Python to execute an operation that CUDA will
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,728 @@
...
+# perform anyway before copying the data from host to device.
+#
+# .. note:: The pytorch implementation of
```
svekars, line 324:
```suggestion
# .. note:: The PyTorch implementation of
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,728 @@
...
+ax.legend(loc="upper left", ncols=3)
+
+plt.show()
+
+del tensors, tensors_pinned
+_ = gc.collect()
+
+
+######################################################################
+# Other copy directions (GPU -> CPU, CPU -> MPS etc.)
```
svekars, line 508:
```suggestion
# Other copy directions (GPU -> CPU, CPU -> MPS)
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,728 @@
...
+#   How many tensors are transferred in a typical operation?
+#
+# - **Size of the tensors to be sent**
+#
+#   What is the size of the tensors being transferred? A few large tensors or many small tensors may not benefit from
+#   the same transfer program.
+#
+# - **System Architecture**
+#
+#   How is the system's architecture influencing data transfer speeds (e.g., bus speeds, network latency)?
```
svekars, line 696:
```suggestion
#   How is the system's architecture influencing data transfer speeds (for example, bus speeds, network latency)?
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,728 @@
...
+#   .. _pinned_memory_resources:
+#
+#  If you are dealing with issues with memory copies when using CUDA devices or want to learn more about
+#  what was discussed in this tutorial, check the following references:
+#
+#  - `CUDA toolkit memory management doc <https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html>`_
+#  - `CUDA pin-memory note <https://forums.developer.nvidia.com/t/pinned-memory/268474>`_
+#  - `How to Optimize Data Transfers in CUDA C/C++ <https://developer.nvidia.com/blog/how-optimize-data-transfers-cuda-cc/>`_
+#  - tensordict :meth:`~tensordict.TensorDict.to` method;
+#
```
svekars, line 728: I thin there is a bit of an issue with rendering this part because of an extra space in all those liens. If you have just one space instead of two, the indentation will be correct. 
Screenshot 2024-07-30 at 9 22 20 AM

svekars (commented):
Just a couple very minor editorial nits - looks great, otherwise!

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,728 @@
...
+#   .. _pinned_memory_resources:
+#
+#  If you are dealing with issues with memory copies when using CUDA devices or want to learn more about
+#  what was discussed in this tutorial, check the following references:
+#
+#  - `CUDA toolkit memory management doc <https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html>`_
+#  - `CUDA pin-memory note <https://forums.developer.nvidia.com/t/pinned-memory/268474>`_
+#  - `How to Optimize Data Transfers in CUDA C/C++ <https://developer.nvidia.com/blog/how-optimize-data-transfers-cuda-cc/>`_
+#  - tensordict :meth:`~tensordict.TensorDict.to` method;
```
vmoens, line 727:
```suggestion
# If you are dealing with issues with memory copies when using CUDA devices or want to learn more about
# what was discussed in this tutorial, check the following references:
#
# - `CUDA toolkit memory management doc <https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html>`_;
# - `CUDA pin-memory note <https://forums.developer.nvidia.com/t/pinned-memory/268474>`_;
# - `How to Optimize Data Transfers in CUDA C/C++ <https://developer.nvidia.com/blog/how-optimize-data-transfers-cuda-cc/>`_;
# - tensordict :meth:`~tensordict.TensorDict.to` method.
```

svekars (approved):

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,728 @@
...
+benchmark_with_profiler(streamed=True, pinned=False)
+
+######################################################################
+#
+# .. figure:: /_static/img/pinmem/trace_streamed0_pinned1.png
+#    :alt:
+#
+# Sending a pageable tensor to GPU on a separate stream is also a blocking operation:
+
+benchmark_with_profiler(streamed=False, pinned=True)
```
mikaylagawarecki, line 218:
```suggestion
benchmark_with_profiler(streamed=True, pinned=False)
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,728 @@
...
+
+benchmark_with_profiler(streamed=False, pinned=False)
+
+######################################################################
+# .. figure:: /_static/img/pinmem/trace_streamed0_pinned0.png
+#    :alt:
+#
+# Using a pinned tensor doesn't change the trace much, both operations are still executed consecutively:
+
+benchmark_with_profiler(streamed=True, pinned=False)
```
mikaylagawarecki, line 209: From the text and profile, think this line of code should be swapped with the one below

```suggestion
benchmark_with_profiler(streamed=False, pinned=True)
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,728 @@
...
+# the hood, a pageable tensor must be copied to pinned memory before being sent to GPU.
+#
+# However, contrary to a somewhat common belief, calling :meth:`~torch.Tensor.pin_memory()` on a pageable tensor before
+# casting it to GPU should not bring any significant speed-up, on the contrary this call is usually slower than just
+# executing the transfer. This makes sense, since we're actually asking Python to execute an operation that CUDA will
+# perform anyway before copying the data from host to device.
+#
+# .. note:: The PyTorch implementation of
+#   `pin_memory <https://github.com/pytorch/pytorch/blob/5298acb5c76855bc5a99ae10016efc86b27949bd/aten/src/ATen/native/Memory.cpp#L58>`_
+#   which relies on creating a brand new storage in pinned memory through `cudaHostAlloc <https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gb65da58f444e7230d3322b6126bb4902>`
```
mikaylagawarecki, line 326: This link is not rendering nicely, maybe missing an `_` after

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,728 @@
...
+# .. note:: Interestingly, the blocking ``to("cuda")`` actually performs the same asynchronous device casting operation
+#   (``cudaMemcpyAsync``) as the one with ``non_blocking=True`` with a synchronization point after each copy.
+#
+# Synergies
+# ~~~~~~~~~
+#
+#   .. _pinned_memory_synergies:
+#
+# Now that we have made the point that data transfer of tensors already in pinned memory to GPU is faster than from
+# pageable memory, and that we know that doing these transfers asynchronously is also faster than synchronously, we can
```
mikaylagawarecki, line 435: nit: 
```suggestion
# pageable memory, and we know that doing these transfers asynchronously is also faster than synchronously, we can
```

intermediate_source/pinmem_nonblock.py
```diff
@@ -0,0 +1,728 @@
...
+#
+#   .. _pinned_memory_resources:
+#
+# If you are dealing with issues with memory copies when using CUDA devices or want to learn more about
+# what was discussed in this tutorial, check the following references:
+#
+# - `CUDA toolkit memory management doc <https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html>`_;
+# - `CUDA pin-memory note <https://forums.developer.nvidia.com/t/pinned-memory/268474>`_;
+# - `How to Optimize Data Transfers in CUDA C/C++ <https://developer.nvidia.com/blog/how-optimize-data-transfers-cuda-cc/>`_;
+# - tensordict :meth:`~tensordict.TensorDictBase.to` method.
```
mikaylagawarecki, line 727: nit: 
this doesn't link it

mikaylagawarecki (approved):
Please address the nits though

PR #2931, opened 2024-06-14
Adding xeon_run_cpu.rst doc

ZailiWang:
This PR is to add a user guide for `torch.backends.xeon.run_cpu` script.
https://github.com/pytorch/pytorch/blob/main/torch/backends/xeon/run_cpu.py

The content is same with pytorch PR https://github.com/pytorch/pytorch/pull/128099
After some discussion we moved it to tutorials part.

@jingxu10 @jgong5 @svekars @kwen2501 @kurman 
Please help review it.

cc @sekyondaMeta @svekars @kit1980 @brycebortree

PR #2989, opened 2024-08-01
fix xeon_run_cpu.html link

ZailiWang:
Fixes #ISSUE_NUMBER

## Description
To correct the link in index.rst for xeon_run_cpu.html. (previous PR #2931)

Hi @svekars I'm not quite sure about the principles which docs should be put into the 2nd-tier `recipes` subfolder. If this tutorial should be moved to the subfolder instead of changing the link in index.rst, please let me know so that I'll update this PR.

cc @svekars @brycebortree

### DIFF
--- a/recipes_source/recipes_index.rst
+++ b/recipes_source/recipes_index.rst
@@ -278,7 +278,7 @@ Recipes are bite-sized, actionable examples of how to use specific PyTorch featu
    :header: CPU launcher script for optimal performance on Intel® Xeon
    :card_description: How to use launcher script for optimal runtime configurations on Intel® Xeon CPUs.
    :image: ../_static/img/thumbnails/cropped/profiler.png
-   :link: ../recipes/recipes/xeon_run_cpu.html
+   :link: ../recipes/xeon_run_cpu.html
    :tags: Model-Optimization
 
 .. customcarditem::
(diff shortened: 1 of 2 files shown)


svekars (approved):

svekars:
The change looks good to me. We can put under just `recipes/`

recipes_source/recipes_index.rst
```diff
@@ -278,7 +278,7 @@ Recipes are bite-sized, actionable examples of how to use specific PyTorch featu
    :header: CPU launcher script for optimal performance on Intel® Xeon
```
svekars, line 278: One thing I noticed that the in the card you call the tutorial "CPU launcher script for optimal performance on Intel® Xeon" and in the actual [page](https://docs-preview.pytorch.org/pytorch/tutorials/2989/recipes/xeon_run_cpu.html) it's "Optimizing PyTorch Inference with Intel® Xeon® Scalable Processors" - you might want to keep the same name to avoid the confusion. 
Screenshot 2024-08-01 at 10 53 13 AM

ZailiWang: Thanks for finding this. The title is aligned as "Optimizing CPU Performance on Intel® Xeon® with run_cpu Script"

PR #3002, opened 2024-08-17
fix: rm `use_cuda` param

shaoyuyoung:
this PR removes the `use_cuda` param as it originally shows below:
https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html#pytorch-profiler
image

Actually, the param `use_cuda` is deprecated
image
If this param is still in tutorials, it would mislead someone.  

cc @aaronenyeshi @chaekit @subramen


svekars (approved):
Thanks!

aaronenyeshi (approved):
LGTM!

PR #3012, opened 2024-08-26
Add weights_only=True to torch.load

### LINKED ISSUE #3007: [BUG] - FutureWarning in "Introduction to PyTorch - Save and Load the Model" tutorial
### Add Link

https://pytorch.org/tutorials/beginner/basics/saveloadrun_tutorial.html

### Describe the bug

Getting 2x FutureWarnings from lines `model.load_state_dict(torch.load('model_weights.pth'))` and `model = torch.load('model.pth')` in the Save and Load Model tutorial. I think this is what it wants changing:
> We recommend you start setting `weights_only=True`

The full warning I am seeing is:
>/var/lib/workspace/beginner_source/basics/saveloadrun_tutorial.py:37: FutureWarning:
You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.

### Describe your environment

Shows for me in the tutorial on my web browser. Also on MacOS, torch version 2.4.0.

cc @subramen @albanD

svekars:
Fixes #3007 
Fixing the issue and a bunch of other tutorials that use torch.load. In the generated output we have something like this:

Screenshot 2024-08-26 at 2 11 11 PM

We should fix all those warnings.

### DIFF
--- a/beginner_source/basics/saveloadrun_tutorial.py
+++ b/beginner_source/basics/saveloadrun_tutorial.py
@@ -34,7 +34,7 @@ torch.save(model.state_dict(), 'model_weights.pth')
 # using ``load_state_dict()`` method.
 
 model = models.vgg16() # we do not specify ``weights``, i.e. create untrained model
-model.load_state_dict(torch.load('model_weights.pth'))
+model.load_state_dict(torch.load('model_weights.pth', weights_only=True))
 model.eval()
 
 ###########################
@@ -52,7 +52,7 @@ torch.save(model, 'model.pth')
 ########################
 # We can then load the model like this:
 
-model = torch.load('model.pth')
+model = torch.load('model.pth', weights_only=False),
 
 ########################
 # .. note:: This approach uses Python `pickle <https://docs.python.org/3/library/pickle.html>`_ module when serializing the model, thus it relies on the actual class definition to be available when loading the model.
--- a/beginner_source/basics/saveloadrun_tutorial.py
+++ b/beginner_source/basics/saveloadrun_tutorial.py
@@ -32,6 +32,11 @@ torch.save(model.state_dict(), 'model_weights.pth')
 ##########################
 # To load model weights, you need to create an instance of the same model first, and then load the parameters
 # using ``load_state_dict()`` method.
+#
+# In the code below, we set ``weights_only=True`` to limit the
+# functions executed during unpickling to only those necessary for
+# loading weights. Using ``weights_only=True`` is considered
+# a best practice when loading weights.
 
 model = models.vgg16() # we do not specify ``weights``, i.e. create untrained model
 model.load_state_dict(torch.load('model_weights.pth', weights_only=True))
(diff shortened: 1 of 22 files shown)


intermediate_source/ddp_tutorial.rst
```diff
@@ -214,10 +214,10 @@ and elasticity support, please refer to `TorchElastic <https://pytorch.org/elast
         # configure map_location properly
         map_location = {'cuda:%d' % 0: 'cuda:%d' % rank}
         ddp_model.load_state_dict(
-            torch.load(CHECKPOINT_PATH, map_location=map_location))
+            torch.load(CHECKPOINT_PATH, map_location=map_location, weights_only=True))
 
         loss_fn = nn.MSELoss()
-        optimizer = optim.SGD(ddp_model.parameters(), lr=0.001)
+        optimizer = optim.SGD(ddp_model.parameters(),gr=0.001)
```
svekars, line 220:
```suggestion
        optimizer = optim.SGD(ddp_model.parameters(), lr=0.001)
```

beginner_source/basics/saveloadrun_tutorial.py
```diff
@@ -34,7 +34,7 @@
 # using ``load_state_dict()`` method.
 
 model = models.vgg16() # we do not specify ``weights``, i.e. create untrained model
-model.load_state_dict(torch.load('model_weights.pth'))
+model.load_state_dict(torch.load('model_weights.pth', weights_only=True))
```
mikaylagawarecki, line 37: nit: since this is the main landing page for save load, a comment along these lines  would be great 

"Setting `weights_only=True` restricts the functions that can be executed during unpickling to those required to load weights, using this flag is a best practice for loading weights."

beginner_source/basics/saveloadrun_tutorial.py
```diff
@@ -52,7 +52,7 @@
 ########################
 # We can then load the model like this:
 
-model = torch.load('model.pth')
+model = torch.load('model.pth', weights_only=False),
```
mikaylagawarecki, line 55: nit: I would add a comment here that as noted in https://pytorch.org/docs/main/notes/serialization.html#saving-and-loading-torch-nn-modules, saving `state_dict`s is the best practice and that we use `weights_only=False` here because this is loading a model, which is a legacy use case for `torch.save`

mikaylagawarecki (commented):
Thanks a lot for fixing these! I just have two quick nit comments around messaging for weights_only=True/False

svekars:
Thanks, @mikaylagawarecki! I added your notes - can you take a look?

mikaylagawarecki (approved):
Thanks!

PR #3013, opened 2024-08-26
Updated Doc for Intel XPU Profile

louie-tsai:
## Description
PyTorch Profiling changes for XPU.

Landing page: 
https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html
https://pytorch.org/tutorials/recipes/profile_with_itt.html

## Checklist

cc @gujinghui @EikanWang @fengyuan14 @guangyey


recipes_source/recipes/profiler_recipe.py
```diff
@@ -197,6 +198,46 @@
 #    Self CPU time total: 23.015ms
 #    Self CUDA time total: 11.666ms
 #
+######################################################################
+# Profiler can also be used to analyze performance of models executed on XPUs:
+
+model = models.resnet18().xpu()
```
dvrogozh, line 204: That's python script, not rst document. As such, mixing up explicit examples for cuda and xpu would give issues to people to try this example since they might not have both cuda and xpu on the system. So, they will start seeing errors.

Maybe it's better to follow the pattern to abstract device with `.to(device)`, hardcode `cuda` and let user modify for `xpu`?

guangyey: Yes, we should refactor this python script to be device-agnostic and let the user know it supports on cuda and xpu. And in the future, the other backend also can expand it.Yes, we should refactor this python script to be device-agnostic and let user know it supports on cuda and xpu. And in the future, the other backend also can expand it.

louie-tsai: some goods. Let me work on that and update PR later this week.

recipes_source/recipes/profiler_recipe.py
```diff
@@ -197,6 +198,46 @@
...
+
+######################################################################
+# (Note: the first use of XPU profiling may bring an extra overhead.)
+
+######################################################################
+# The resulting table output (omitting some columns):
+#
+# .. code-block:: sh
+#
+-------------------------------------------------------  ------------  ------------  ------------  ------------  ------------
```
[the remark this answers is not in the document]
dvrogozh: please, resolve this comment

recipes_source/recipes/profiler_recipe.py
```diff
@@ -161,17 +161,26 @@
 # Note the occurrence of ``aten::convolution`` twice with different input shapes.
 
 ######################################################################
-# Profiler can also be used to analyze performance of models executed on GPUs:
+# Profiler can also be used to analyze performance of models executed on GPUs and XPUs:
+# Users could switch between cpu, cuda and xpu
+device = 'cuda'
```
guangyey, line 166:
```suggestion
if torch.cuda.is_available():
    device = 'cuda'
elif torch.xpu.is_available():
    device = 'xpu'
else:
    device = 'cpu'
```
It is better to initialize `activities` via this if-else statement. And ensure this Python script can run successfully on CPU-only, cuda, and xpu device.

recipes_source/recipes/profiler_recipe.py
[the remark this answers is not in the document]
louie-tsai: thanks. addressed it accordingly.

recipes_source/recipes/profiler_recipe.py
```diff
@@ -160,17 +161,31 @@
...
-with profile(activities=[
-        ProfilerActivity.CPU, ProfilerActivity.CUDA], record_shapes=True) as prof:
+# Profiler can also be used to analyze performance of models executed on GPUs and XPUs:
+# Users could switch between cpu, cuda and xpu
+if torch.cuda.is_available():
+    device = 'cuda'
+elif torch.xpu.is_available():
+    device = 'xpu'
+else:
+    device = 'cpu'
```
dvrogozh, line 171: `cpu` was actually already demonstrated above. I think you can just terminate the script here:
```suggestion
    print('Neither CUDA nor XPU devices are available to demonstrate profiling on acceleration devices')
    import sys
    sys.exit(0)
```

dvrogozh (changes requested):

louie-tsai:
> Changes to profile_with_itt looks fine to me, but changes to profiler_recpie makes it really hard to read. Perhaps one should try to find an easier to use interface for collecting/analyzing profiler information from accelerators?

@malfet 
The profiler_recipe contents will also be populated into a html file : https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html
Hope that html file format addresses your concern for "hard to read".

malfet:
> The profiler_recipe contents will also be populated into a html file : https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html Hope that html file format addresses your concern for "hard to read".

Sorry, I was not talking about this one, but rather about the following code:
```python
if device == 'cuda':
   activities.append(ProfilerActivity.CUDA)
elif device == 'xpu':
   activities.append(ProfilerActivity.XPU)
```
As one can not compile PyTorch with CUDA and XPU at the same time, why introducing new enum rather than creating an alias(i.e. `ProfilerActivity.XPU == ProfilerActivity.CUDA`), so that users do not have to rewrite their programs when migrating from one accelerator to another?

jingxu10:
> > The profiler_recipe contents will also be populated into a html file : https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html Hope that html file format addresses your concern for "hard to read".
> 
> Sorry, I was not talking about this one, but rather about the following code:
> 
> ```python
> if device == 'cuda':
>    activities.append(ProfilerActivity.CUDA)
> elif device == 'xpu':
>    activities.append(ProfilerActivity.XPU)
> ```
> 
> As one can not compile PyTorch with CUDA and XPU at the same time, why introducing new enum rather than creating an alias(i.e. `ProfilerActivity.XPU == ProfilerActivity.CUDA`), so that users do not have to rewrite their programs when migrating from one accelerator to another?

Hi @malfet , it doesn't seem to be something can be fixed in this tutorial. It involves code changes. We will see how to address this concern in 2.6 with a separate PR to change profiler code.

dvrogozh:
> As one can not compile PyTorch with CUDA and XPU at the same time

@malfet : why? this should be technically possible if someone will install both CUDA and XPU stacks on the system. And might be convenient for some people (mostly for debug I guess) to build pytorch with both path enabled in a single build.

louie-tsai:
> > As one can not compile PyTorch with CUDA and XPU at the same time
> 
> @malfet : why? this should be technically possible if someone will install both CUDA and XPU stacks on the system. And might be convenient for some people (mostly for debug I guess) to build pytorch with both path enabled in a single build.
@malfet 
Any feedback?  Ex: some laptops also have both Intel iGPU and NVidia dGPU at the same time, so it might be common to have both XPU and CUDA.

recipes_source/recipes/profiler_recipe.py
```diff
@@ -160,17 +161,33 @@
...
+    import sys
+    sys.exit(0)
+
+activities = [ProfilerActivity.CPU]
+sort_by_keyword = device + "_time_total"
+
+if device == 'cuda':
+   activities.append(ProfilerActivity.CUDA)
+elif device == 'xpu':
+   activities.append(ProfilerActivity.XPU)
```
malfet, line 181: No need for this complex conditional, it's OK to ask profiler to collect activities even if device is missing
```suggestion
activities = [ProfilerActivity.CPU, ProfilerActivity.CUDA, ProfilerActivity.XPU]
sort_by_keyword = device + "_time_total"
```

recipes_source/recipes/profiler_recipe.py
```diff
@@ -197,6 +214,36 @@
...
+#                                     aten::_convolution      0.000us         0.00%       1.871ms      93.560us            20
+#                         aten::convolution_overrideable      1.871ms        72.89%       1.871ms      93.560us            20
+#                                               gen_conv      1.484ms        57.82%       1.484ms      74.216us            20
+#                                       aten::batch_norm      0.000us         0.00%     432.640us      21.632us            20
+#                           aten::_batch_norm_impl_index      0.000us         0.00%     432.640us      21.632us            20
+#                                aten::native_batch_norm      432.640us      16.85%     432.640us      21.632us            20
+#                                           conv_reorder      386.880us      15.07%     386.880us       6.448us            60
+#  -------------------------------------------------------   ------------  ------------  ------------  ------------  ------------
+#  Self CPU time total: 712.486ms
+#  Self XPU time total: 2.567ms
```
malfet, line 243: What value does this extra table brings to the user?

recipes_source/recipes/profiler_recipe.py
```diff
@@ -266,11 +313,20 @@
...
+# Users could switch between cpu, cuda and xpu
+device = 'cuda'
 
-model = models.resnet18().cuda()
-inputs = torch.randn(5, 3, 224, 224).cuda()
+activities = [ProfilerActivity.CPU]
+if device == 'cuda':
+   activities.append(ProfilerActivity.CUDA)
+elif device == 'xpu':
+   activities.append(ProfilerActivity.XPU)
```
malfet, line 324:
```suggestion
activities = [ProfilerActivity.CPU, ProfilerActivity.CUDA, ProfilerActivity.XPU]
```

recipes_source/recipes/profiler_recipe.py
```diff
@@ -282,20 +338,29 @@
...
 #    :scale: 25 %
 
+######################################################################
+# You can examine the sequence of profiled operators and XPU kernels
+# in Chrome trace viewer (``chrome://tracing``):
+#
+# .. image:: ../../_static/img/trace_xpu_img.png
+#    :scale: 25 %
+
+
```
malfet, line 348: IMO it does not add any additional info to the user
```suggestion
```

malfet: Just add `CUDA/XPU` (or Accelerator) kernels to the line 335

louie-tsai:
@malfet @jingxu10 @dvrogozh @guangyey 
please help to merge it if no concern for the change.

thanks!

recipes_source/recipes/profiler_recipe.py
louie-tsai: there are indeed just minor changes including some different operators and also XPU instead of GPU in this table.  just want people to understand how it might look like for the output.

louie-tsai:
@malfet 
updated accordingly. hope it works for you.

svekars:
To fix chekc spelling, please add these to the en-wordlist.txt:

* _batch_norm_impl_index
* convolution_overrideable 
* aten
* XPU

louie-tsai:
> To fix spelling, please add these to the en-wordlist.txt:
> 
> * _batch_norm_impl_index
> * convolution_overrideable
> * aten
> * XPU

addressed them accordingly. please also help to approve it if all look good to you.

malfet (approved):
Feels a bit verbose to me, but looks fine.

PR #3038, opened 2024-09-05
Remove sphinx theme from .gitmodules

svekars:
Because we import the theme [here](https://github.com/pytorch/tutorials/blob/main/.ci/docker/requirements.txt#L47), we don't really need it as a submodule. It also seem to generate some unneeded [paths](https://docs-preview.pytorch.org/pytorch/tutorials/3035/sitemap.xml) in the sitemap.


clee2000 (approved):
Seems reasonable to me

PR #3077, opened 2024-10-05
Fix logging issue for unsupported torch.compile devices

rghosh08:
**Fixes #137285**

## Description

This PR addresses the issue of insufficient logging when `torch.compile` is not supported on devices with CUDA capability below 7.0. Instead of exiting the program with a simple message, the code now logs detailed information about the device capabilities and reasons for the lack of support for `torch.compile`. It also ensures that the program continues to execute, showcasing the `TORCH_LOGS` logging system even when the function is not compiled.

### Changes made:
- Added detailed logging to explain why `torch.compile` is not supported on devices with lower CUDA capabilities.
- Introduced a fallback mechanism where the function runs uncompiled if the device doesn't support compilation.
- Consolidated the device selection logic for cleaner code.
- Ensured the program runs and demonstrates `TORCH_LOGS` regardless of device capabilities.
- Refactored code to adhere to PyTorch standards and PEP 8 linting guidelines.

---

## Checklist
- [x] The issue that is being fixed is referred to in the description (see above "Fixes #137285").
- [x] Only one issue is addressed in this pull request.
- [x] Labels from the issue that this PR is fixing are added to this pull request.
- [x] No unnecessary issues are included in this pull request.


Tharusha-Lekamge:
Small suggestion - Link the PR to the issue. Currently the issue has no linked PRs

svekars (commented):
I don't understand why everything is being deleted here.

mlazos:
@rghosh08 I don't think this PR makes sense, as logging is explicitly meant for `torch.compile` at the moment. If you run without `torch.compile` nothing of note is printed so it is kind of a useless tutorial. When I ran this on my machine (not compiling the function) and get:
```txt
INFO:__main__:CUDA Device Capability: (8, 0)
INFO:__main__:Device supports torch.compile.
INFO:__main__:Running the function without compilation...
===================Dynamo Tracing=========================
===================Traced Graph=========================
===================Fusion Decisions=========================
===================Output Code=========================
============================================
```

I also don't understand why so much of the tutorial aspect of the script is getting removed. Can you explain further?

PR #3090, opened 2024-10-13
replace decoder with self.decoder in PyTorch Chatbot example.

Ishan-Kumar2:
## Description
The decoder function needs to use self.decoder else it will throw a decoder not defined error.

## Checklist


albanD (approved):
Thanks!

PR #3144, opened 2024-11-04
Install dependencies globally instead of using conda

clee2000:
Stop using conda to install python and other dependnecies, and also don't use an conda environment.  We don't even have any packages that are exclusively on conda.  

* upgrade to ubuntu to 22.04 because 20.04 didn't have python3.10
* make link checker ignore .jenkins/build.sh


.ci/docker/common/install_pip_requirements.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -ex
+
+python3.10 -m venv /opt/venv/tutorials
+source /opt/venv/tutorials/bin/activate
+pip install --upgrade pip
+pip install -r ./requirements.txt
+chown ci-user:ci-user /opt/venv -R
```
malfet, line 9: Why do we need a venv here? It's our docker container, why not install dependencies system-wide?

malfet (approved):
LGTM, but consider getting rid of venv as well:
 ubuntu-22.04 has python-3.10 as default system compiler, just install everything system wide

.ci/docker/common/install_pip_requirements.sh
clee2000: Removed

Issue #3113, opened 2024-10-24
💡 [REQUEST] - Update tutorials with device-generic APIs


albanD:
### 🚀 Describe the improvement or the new tutorial

We should use the latest device-generic APIs when they come out in 2.6 in all tutorials to improve readability.

### Existing tutorials on this topic

https://pytorch.org/tutorials/beginner/basics/buildmodel_tutorial is an example of one we should update. There is most likely more.

### Additional context

cc @guangyey that might be a good follow up to consider before 2.6

guangyey:
I am delighted to take on this task. I couldn't self-assign, please assign it to me.

shink:
I'd love to contribute as well, but do we have a list of which APIs should be replaced and with what?

AndyDralle:
Just a note from a user-perspective: It's very confusing to have the tutorials updated before the features are actually available. 
I installed Pytorch yesterday, in order to try out https://pytorch.org/tutorials/beginner/transfer_learning_tutorial.html. By chance, this happened to be directly after the documentation was updated, I suppose. 

Anyway, pip gave me torch 2.5.1, which means that the tutorial code doesn't actually work for me. And the problem isn't google-able, since it's so new, and the documentation doesn't hint at this 2.6 version requirement either. I only found this explanation, because I happened to look the tutorial page up on GitHub, where I noticed this very recent commit.

PR #3170, opened 2024-11-27
[3/N] Refine beginner tutorial by accelerator api

guangyey:
For https://github.com/pytorch/tutorials/issues/3113

# Motivation
* __->__ #3170
* #3168
* #3167

As titled, we aim to use accelerator API to refine the beginner model tutorial to make it device-agnostic.

## Checklist


beginner_source/chatbot_tutorial.py
```diff
@@ -1318,16 +1320,16 @@ def evaluateInput(encoder, decoder, searcher, voc):
     encoder_optimizer.load_state_dict(encoder_optimizer_sd)
     decoder_optimizer.load_state_dict(decoder_optimizer_sd)
 
-# If you have CUDA, configure CUDA to call
+# If you have accelerator, configure it to call
```
albanD, line 1323:
```suggestion
# If you have an accelerator, configure it to call
```

albanD (commented):
Sounds good to me.
Will let @svekars take a final look.

beginner_source/chatbot_tutorial.py
guangyey: Thanks @albanD

PR #3188, opened 2024-12-17
Fix the DCGAN C++ shape warning

LiAuTraver:
The issue in example repo has been fixed but on website it did not update. I found it today while following the [tutorials](https://pytorch.org/tutorials/advanced/cpp_frontend.html).

Issue see [here](https://github.com/pytorch/examples/issues/819#issue-686390601), main code repo PR: [here](https://github.com/pytorch/examples/pull/1207#issue-2061218185). 

This PR adopted the solution mentioned in the issue and tested with the latest version libtorch(current 2.5.1) with cuda 12.6.

cc @albanD @jbschlosser


albanD (approved):
Sounds fair !
A squeeze(1) would also be ok here if we need the broadcasting to happen in the loss

PR #3204, opened 2024-12-27
Mention Python agnosticism in custom ops tutorial

janeyx99:
In 2.6, we can confidently tell people to use the py_limited_api to build python agnostic wheels.
Including the instructions within the existing custom op tutorial.

### DIFF
--- a/advanced_source/cpp_custom_ops.rst
+++ b/advanced_source/cpp_custom_ops.rst
@@ -66,10 +66,12 @@ If you need to compile CUDA code (for example, ``.cu`` files), then instead use
 Please see `extension-cpp <https://github.com/pytorch/extension-cpp>`_ for an
 example for how this is set up.
 
-In PyTorch 2.6 and later, if your custom library adheres to the `CPython stable
-Limited API <https://docs.python.org/3/c-api/stable.html>`_ or avoids CPython
-entirely, you can build one python agnostic wheel against a minimum supported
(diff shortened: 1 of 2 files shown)


advanced_source/cpp_custom_ops.rst
```diff
@@ -63,9 +63,45 @@ Using ``cpp_extension`` is as simple as writing the following ``setup.py``:
 
 If you need to compile CUDA code (for example, ``.cu`` files), then instead use
 `torch.utils.cpp_extension.CUDAExtension <https://pytorch.org/docs/stable/cpp_extension.html#torch.utils.cpp_extension.CUDAExtension>`_.
-Please see how
-`extension-cpp <https://github.com/pytorch/extension-cpp>`_ for an example for
-how this is set up.
+Please see `extension-cpp <https://github.com/pytorch/extension-cpp>`_ for an
+example for how this is set up.
+
+In PyTorch 2.6 and later, if your custom library adheres to the `CPython stable
```
albanD, line 69: I think it would be good to add an extra sentence here to be very explicit about what this bring: 
```suggestion
Starting with PyTorch 2.6, you can now build a single wheel for all CPython version (like you would do for pure python packages). 
In particular, if your custom library adheres to the `CPython stable
```

albanD (approved):
Small nit but SGTM otherwise!

PR #3221, opened 2025-01-08
Revert prefactor m in autograd documentation

### LINKED ISSUE #2168: Missing parameter in math formula
Reading [Deep Learning with PyTorch: A 60 Minute Blitz](https://github.com/pytorch/tutorials/blob/master/beginner_source/blitz/autograd_tutorial.py), located at line 191-209

$$
J^{T}\cdot \vec{v}=\left(\begin{array}{ccc}
       \frac{\partial y_{1}}{\partial x_{1}} & \cdots & \frac{\partial y_{m}}{\partial x_{1}}\\
       \vdots & \ddots & \vdots\\
       \frac{\partial y_{1}}{\partial x_{n}} & \cdots & \frac{\partial y_{m}}{\partial x_{n}}
       \end{array}\right)\left(\begin{array}{c}
       \frac{\partial l}{\partial y_{1}}\\
       \vdots\\
       \frac{\partial l}{\partial y_{m}}
       \end{array}\right)=\left(\begin{array}{c}
       \frac{\partial l}{\partial x_{1}}\\
       \vdots\\
       \frac{\partial l}{\partial x_{n}}
       \end{array}\right)
$$

* It looks like we need add a `m` as multiplier in right side of the fomula, as we have `m` items to accumulate.

cc @suraj813

asideofcode-dev:
Fixes #2168

## Description

The prefactor m was incorrectly added to the vector-Jacobian product formula in the autograd tutorial as part of https://github.com/pytorch/tutorials/pull/3181. This change was based on the mistaken assumption that m-scaling was necessary to account for multiple terms in the summation. However, the chain rule and vector-Jacobian product formula already correctly aggregate contributions from all intermediate variables without requiring explicit scaling by m.

## Checklist


albanD (approved):
My bad! Thanks for taking the time to send a fix!

asideofcode-dev:
@albanD No worries! Thanks for approving. I think some workflows need approval before moving forward... not 100% sure how the approval process works but I'll leave the rest in your hands :) Thanks again for the prompt approval

albanD:
I'll let @svekars check that part!

PR #3233, opened 2025-01-16
Update spellcheck to run only on changed files

svekars:
.rst, .md, and .py file changes are included for testing purposes only and will not be merged.


malfet (approved):
Sure, though unless it's slowest job, it probably would be better to run spell checking on all files, as those kinds of changes increase the chances that changed file will be skipped or complex PR will fail the lint due to bugs in the script

PR #3251, opened 2025-01-23
[testing]

clee2000:
Testing for a change in the theme


huydhn (approved):
Good catch!

svekars:
Need to now figure out why fonts look different compared to when we install with the "-e" option.

PR #3257, opened 2025-01-24
change torch.cuda.amp.GradScaler to torch.GradScaler("cuda")

0ssamaak0:
## Description
using `torch.cuda.amp.GradScaler` will result in a warning
```
FutureWarning: `torch.cuda.amp.GradScaler(args...)` is deprecated. Please use `torch.amp.GradScaler('cuda', args...)` instead.
```
So it should be changed according to [Automatic Mixed Precision package Documentation](https://pytorch.org/docs/stable/amp.html#gradient-scaling) `torch.cuda.amp.GradScaler` to `torch.GradScaler("cuda")` to match

## Checklist


recipes_source/recipes/amp_recipe.py
```diff
@@ -150,7 +150,7 @@ def make_model(in_size, out_size, num_layers):
 # The same ``GradScaler`` instance should be used for the entire convergence run.
 # If you perform multiple convergence runs in the same script, each run should use
 # a dedicated fresh ``GradScaler`` instance. ``GradScaler`` instances are lightweight.
-scaler = torch.cuda.amp.GradScaler()
+scaler = torch.GradScaler("cuda")
```
spzala, line 153: @0ssamaak0 thanks for the PR. The deprecated warning says to use `torch.amp.GradScaler('cuda', args...)` so why are you changing to `torch.GradScaler("cuda")` ?  i.e. why did you remove `amp`

recipes_source/recipes/amp_recipe.py
```diff
@@ -182,7 +182,7 @@ def make_model(in_size, out_size, num_layers):
 
 net = make_model(in_size, out_size, num_layers)
 opt = torch.optim.SGD(net.parameters(), lr=0.001)
-scaler = torch.cuda.amp.GradScaler(enabled=use_amp)
+scaler = torch.GradScaler("cuda" ,enabled=use_amp)
```
spzala, line 185: Same as above.

spzala (commented):
Please see comments inline. Thanks!

recipes_source/recipes/amp_recipe.py
0ssamaak0: I followed this
image

[Source](https://pytorch.org/docs/stable/amp.html#gradient-scaling)

recipes_source/recipes/amp_recipe.py
0ssamaak0: I followed this
image

[Source](https://pytorch.org/docs/stable/amp.html#gradient-scaling)

recipes_source/recipes/amp_recipe.py
spzala: Thanks for the quick reply. `GradScaler` can be used on its own but because you are modifying `amp` recipe, I would suggest to use `torch.amp.GradScaler` as it's also mentioned in the code - https://github.com/pytorch/pytorch/blob/main/torch/cuda/amp/grad_scaler.py#L12 
Also, if you run the recipe, you will notice the warning message suggesting to use `torch.amp.GradScaler`

recipes_source/recipes/amp_recipe.py
spzala: Thanks for the quick reply. `GradScaler` can be used on its own but because you are modifying `amp` recipe, I would suggest to use `torch.amp.GradScaler` as it's also mentioned in the code - https://github.com/pytorch/pytorch/blob/main/torch/cuda/amp/grad_scaler.py#L12 
Also, if you run the recipe, you will notice the warning message suggesting to use `torch.amp.GradScaler`

0ssamaak0: I have changed the code accordingly.

Thank you

spzala: Thanks!

malfet (approved):
Looks good for me, thank you for the update

PR #3324, opened 2025-04-15
Run tutorials on g5

svekars:
Why we're making this change
-------------------------------
PR authors are frustrated with `torch.compile` tutorials, which aren't supported on our older workers. As more tutorials adopt `torch.compile`, users must constantly update `metadata.json` to ensure their tutorials run on compatible workers.

This PR moves these tutorials to G5 instances, providing a consistent environment that supports `torch.compile` out of the box, eliminating the need for manual configuration and improving the overall developer experience. This change should also help to make the whole tutorial build faster.

Cons: Less and less tutorials are actually runnable on Colab, which still serves only T4 machines in their free tier


malfet (approved):
LGTM, but perhaps it would be good to undo the logic that dispatches different tests to different nodes

PR #3336, opened 2025-04-21
Adjust torch.compile() best practices

punkeel:
## Description

1. Add best practice to prefer `mod.compile` over `torch.compile(mod)`, which avoids `_orig_` naming problems. Repro steps:
- opt_mod = torch.compile(mod)
- train opt_mod
- save checkpoint

In another script, potentially on a machine that does NOT support `torch.compile`: load checkpoint.

This fails with an error, because the checkpoint on `opt_mod` got its params renamed by `torch.compile`:
```
RuntimeError: Error(s) in loading state_dict for VQVAE:
	Missing key(s) in state_dict: "embedding.weight", "encoder.encoder.net.0.weight", "encoder.encoder.net.0.bias", ...
	Unexpected key(s) in state_dict: "_orig_mod.embedding.weight", "_orig_mod.encoder.encoder.net.0.weight", "_orig_mod.encoder.encoder.net.0.bias", ...
```

2. Add best practice to use, or at least try, `fullgraph=True`. This doesn't always work, but we should encourage it.

Note: I'm not a PyTorch expert, these are just based on footguns I've encountered over the past week.

## Checklist

### DIFF
--- a/intermediate_source/torch_compile_tutorial.py
+++ b/intermediate_source/torch_compile_tutorial.py
@@ -101,8 +101,11 @@ class MyModule(torch.nn.Module):
         return torch.nn.functional.relu(self.lin(x))
 
 mod = MyModule()
-opt_mod = torch.compile(mod)
-print(opt_mod(t))
+mod.compile()
+print(mod(t))
+## or:
+# opt_mod = torch.compile(mod)
+# print(opt_mod(t))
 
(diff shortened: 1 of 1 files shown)


williamwen42 (approved):

punkeel:
Thank you both & sorry for the formatting quirks!

PR #3346, opened 2025-04-30
Fix missing import for Dataset in custom dataset tutorial

AbhiLegend:
Fixes #152518

## Description
<This PR addresses the missing import in the "Creating a Custom Dataset" section of the PyTorch tutorials.

The class `CustomImageDataset` inherits from `Dataset`, but the `Dataset` class was not imported in that specific code block. This causes a `NameError` if the user runs the code independently.

Added the following line above the custom class definition:
```python
from torch.utils.data import Dataset>

## Checklist
<!--- Make sure to add `x` to all items in the following checklist: -->

### DIFF
--- a/beginner_source/basics/data_tutorial.py
+++ b/beginner_source/basics/data_tutorial.py
@@ -121,6 +121,7 @@ plt.show()
 import os
 import pandas as pd
 from torchvision.io import read_image
+from torch.utils.data import Dataset  # ✅ FIXED
 
 class CustomImageDataset(Dataset):
     def __init__(self, annotations_file, img_dir, transform=None, target_transform=None):

### CODE REVIEW

**Comment by AbhiLegend** [comment]
@pytorch/tutorials-team Could someone please review this PR?

File: beginner_source/basics/data_tutorial.py [hunk 1]
```diff
@@ -121,6 +121,7 @@
 import os
 import pandas as pd
 from torchvision.io import read_image
+from torch.utils.data import Dataset  # ✅ FIXED
```
svekars, line 124: We are already importing Dataset on line 49.

AbhiLegend: Oh Ok.

PR #3351, opened 2025-05-02
[test] call sphinx once per tutorial file

clee2000:
there is ~5 min overhead per file due to sphinx stuff

adds up to maybe 2x time
https://github.com/pytorch/tutorials/actions/runs/14911776461/usage 6 hrs
vs
https://github.com/pytorch/tutorials/actions/runs/15032936259/usage 12 hr

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,6 @@ cleanup.sh
 
 # pyspelling
 dictionary.dic
+
+# CI related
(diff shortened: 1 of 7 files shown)


clee2000:
Doing https://github.com/pytorch/tutorials/pull/3355 instead since its faster

PR #3369, opened 2025-05-30
fix broken URL

partev:
fix broken URL:
https://docs.ray.io/en/master/ray-core/using-ray-with-gpus.html#fractional-gpus ->
https://docs.ray.io/en/latest/ray-core/scheduling/accelerators.html#fractional-accelerators

based on the archived copy of the original link here: https://web.archive.org/web/20220313114338/https://docs.ray.io/en/master/ray-core/using-ray-with-gpus.html#fractional-gpus

it looks like the new link is referring to the same content

Fixes #ISSUE_NUMBER

## Checklist

cc @svekars @sekyondaMeta @AlannaBurke


svekars (approved):
Thanks!

PR #3389, opened 2025-06-09
Visualizing gradients tutorial (issue #3186)

### LINKED ISSUE #3186: Writing a gradient tutorial, focused on leaf vs non leaf tensors.
There is no tutorial that specifically talks about requires_grad, retain_grad, and leaf tensor/ non-leaf tensors and how they interact with each other. Can I write a tutorial specifically talking about this topic? This will be useful when gradients are used in unusual places, as is the case for the deep dream algorithm. 

cc: @albanD

j-silv:
Fixes #3186

## Description

Add draft for visualizing gradients tutorial. [Link is here](https://docs-preview.pytorch.org/pytorch/tutorials/3389/advanced/visualizing_gradients_tutorial.html) but the content is old and the files need to be re-built.

## Checklist


advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,426 @@
+"""
+Visualizing Gradients
+=====================
+
+**Author**: `Justin Silver <https://github.com/j-silv>`_
+
+By performance and efficiency reasons, PyTorch does not save the
```
sekyondaMeta: Change the By to For

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,426 @@
...
+# False         True                False             no-op                                sets ``retains_grad`` to True
+# False         True                True              no-op                                no-op
+# ============  ==================  ================  ===================================  =============================
+
+
+######################################################################
+# References
+# ----------
+# 
+# https://docs.pytorch.org/tutorials/beginner/basics/autogradqs_tutorial
```
sekyondaMeta: Nit but can you change these to hyperlinks

sekyondaMeta:
Generally seems to be headed in the right direction in terms of tone and organization from my perspective. 
Can you add perquisite knowledge for this.

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,426 @@
...
+
+# prints all True because tensors are non-leaf nodes
+print(f"{z.requires_grad=}")      
+print(f"{y_pred.requires_grad=}") 
+print(f"{loss.requires_grad=}")  
+
+
+######################################################################
+# A useful heuristic to remember is that whenever a tensor is a non-leaf,
+# it **has** to have ``requires_grad=True``, otherwise back-propagation
```
soulitzer: non-leaf by definition has requires_grad=True

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,426 @@
...
+# 
+# In the context of a generic DAG then, a *leaf* is simply a node which is
+# at the input (beginning) of the graph, and *non-leaf* nodes are
+# everything else.
+# 
+# To start the generation of the computational graph which can be used for
+# gradient calculation, we need to pass in the ``requires_grad=True``
+# parameter to the tensor constructors. That is because by default,
+# PyTorch is not tracking gradients on any created tensors. To verify
+# this, try removing the parameter above and then run back-propagation:
```
soulitzer: I think you mean, try not setting requires_grad?

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,426 @@
+"""
+Visualizing Gradients
+=====================
+
+**Author**: `Justin Silver <https://github.com/j-silv>`_
+
+By performance and efficiency reasons, PyTorch does not save the
+intermediate gradients when running back-propagation. To visualize the
+gradients of these internal layer tensor, we have to explicitly tell
+PyTorch to retain those values with the ``retain_grad`` parameter.
```
soulitzer: might be too low level to include in the first section?

advanced_source/visualizing_gradients_tutorial.py
soulitzer: Don't think we need any reasoning here: "By performance and efficiency reasons,"  these gradients aren't saved because we don't need them.

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,426 @@
...
+# ::
+# 
+#    >>> x.retain_grad()
+#    RuntimeError: can't retain_grad on Tensor that has requires_grad=False
+# 
+
+
+######################################################################
+# In summary, using ``retain_grad()`` and ``retains_grad`` only make sense
+# for non-leaf nodes, since the ``grad`` attribute has to be populated for
```
soulitzer: "will already" be populated

soulitzer (commented):
Thanks for the working on this tutorial. Overall I'd say though that this section can be much shorter.

> By the end of this tutorial, you will be able to:

> Differentiate between *leaf* and *non-leaf* tensors

have a diagram from https://github.com/szagoruyko/pytorchviz, point to the leafs 

>  Know when to use\ ``retain_grad`` vs. ``require_grad`

"use requires_grad for leaf, use retain_grad for non-leaf"

j-silv:
Thank you for the comments, they were really helpful. Let me know if you think the first section is still too long. 

Concerning the "visualizing gradients" section with an actual example, I'm not sure if I'm going about retaining the gradients for intermediate tensors correctly. My thought process was to use a forward hook, call `retain_grad()` on the output tensor of that module, and then store that output tensor in a list. Later, after calling `loss.backward()`, I could then pluck out the `grad` attribute of that tensor and plot it.

Initially I tried using a backward pass hook like `register_full_backward_hook()` but this didn't work because the `ResNet` model performs some inplace operations (i.e. ReLU and one += addition) and PyTorch complains about it:

_RuntimeError: Output 0 of BackwardHookFunctionBackward is a view and is being modified inplace. This view was created inside a custom Function (or because an input was returned as-is) and the autograd logic to handle view+inplace would override the custom backward associated with the custom Function, leading to incorrect gradients. This behavior is forbidden. You can fix this by cloning the output of the custom Function._

I know that I can plot the gradients for the parameters by just looping through the `named_parameters()` but I would like to also plot the gradients for the intermediate tensors. 

If anyone sees a problem with my method let me know. The current state of the code isn't doing what I expected so I still have to debug it.

**EDIT 1:** I stumbled upon this [issue](https://github.com/pytorch/pytorch/issues/61519). Perhaps it's better to switch to using tensor hooks as suggested by alban, instead of storing the outputs through a forward pass and then later accessing their .grad 

**EDIT 2:** I decided to not use ResNet but instead a simplified fully connected network as explained in the BatchNorm paper. It is purely for educative purposes, but it actually shows the results I was expecting. With the ResNet implementation, I believe that the residual connections and ReLU non-linearity are muddying the negative effect on the gradients if they don't have BatchNorm. I'll push an updated PR sometime today.

sekyondaMeta:
@soulitzer what are your thoughts on these new updates?

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,593 @@
...
+
+print(model_bn.layers[0])
+print(model_nobn.layers[0])
+
+
+######################################################################
+# Because we are using a ``nn.Module`` instead of individual tensors for
+# our forward pass, we need another method to access the intermediate
+# gradients. This is done by `registering a
+# hook <https://www.digitalocean.com/community/tutorials/pytorch-hooks-gradient-clipping-debugging>`__.
```
soulitzer: Cool tutorial, but not totally accurate in its explanations sometimes

Maybe link https://docs.pytorch.org/docs/stable/notes/autograd.html#backward-hooks-execution instead?

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,593 @@
...
+# We can verify that batch normalization is only being applied to one of
+# the models by probing one of the internal layers:
+# 
+
+print(model_bn.layers[0])
+print(model_nobn.layers[0])
+
+
+######################################################################
+# Because we are using a ``nn.Module`` instead of individual tensors for
```
soulitzer: -> Because we wrapped up the logic and state of our model in a nn.Module

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,593 @@
...
+# our forward pass, we need another method to access the intermediate
+# gradients. This is done by `registering a
```
soulitzer: This is only if we want to avoid modifying nn.Module code directly. I mean we can always directly just attach hooks in forward.

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,593 @@
...
+# hook <https://www.digitalocean.com/community/tutorials/pytorch-hooks-gradient-clipping-debugging>`__.
+# 
+# .. warning::
+# 
+#    Note that using backward pass hooks to probe an intermediate nodes gradient is preferred over using `retain_grad()`.
+#    It avoids the memory retention overhead if gradients aren't needed after backpropagation.
+#    It also lets you modify and/or clamp gradients during the backward pass, so they don't vanish or explode.
+#    However, if in-place operations are performed, you cannot use the backward pass hook
```
soulitzer: This only applies to module backward hooks and not to backward hooks attached tensors.

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,593 @@
...
+######################################################################
+# Because we are using a ``nn.Module`` instead of individual tensors for
+# our forward pass, we need another method to access the intermediate
+# gradients. This is done by `registering a
+# hook <https://www.digitalocean.com/community/tutorials/pytorch-hooks-gradient-clipping-debugging>`__.
+# 
+# .. warning::
+# 
+#    Note that using backward pass hooks to probe an intermediate nodes gradient is preferred over using `retain_grad()`.
+#    It avoids the memory retention overhead if gradients aren't needed after backpropagation.
```
soulitzer: Well you'd just have to clear the .grads yourself

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,593 @@
...
+
+######################################################################
+# Because we are using a ``nn.Module`` instead of individual tensors for
+# our forward pass, we need another method to access the intermediate
+# gradients. This is done by `registering a
+# hook <https://www.digitalocean.com/community/tutorials/pytorch-hooks-gradient-clipping-debugging>`__.
+# 
+# .. warning::
+# 
+#    Note that using backward pass hooks to probe an intermediate nodes gradient is preferred over using `retain_grad()`.
```
soulitzer:
```suggestion
#    Note that using backward pass hooks to probe an intermediate node's gradient is preferred over using `retain_grad()`.
```

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,593 @@
...
+# 
+# 1. Any ``nn.Module`` that has ``nn.Parameter`` will have
+#    ``requires_grad=True`` for its parameters (see
+#    `here <https://docs.pytorch.org/tutorials/beginner/basics/quickstart_tutorial.html#creating-models>`__)
+# 2. Locally disabling gradient computation with context managers (see
+#    `here <https://docs.pytorch.org/docs/stable/notes/autograd.html#locally-disabling-gradient-computation>`__)
+# 
+# In summary, ``requires_grad`` tells autograd which tensors need to have
+# their gradients calculated for backpropagation to work. This is
+# different from which gradients have to be stored inside the tensor,
```
soulitzer: -> different from which tensor's gradients field is actually populated

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,593 @@
+"""
+Visualizing Gradients
+=====================
+
+**Author:** `Justin Silver <https://github.com/j-silv>`__
+
+When training neural networks with PyTorch, it’s possible to ignore some
+of the library’s internal mechanisms. For example, running
+backpropagation requires a simple call to ``backward()``. This tutorial
```
soulitzer: Cool intro , but I think we should drop and just begin with what the tutorial does

> When training neural networks with PyTorch, it’s possible to ignore some
of the library’s internal mechanisms. For example, running
backpropagation requires a simple call to ``backward()``.

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,593 @@
...
+dives into how those gradients are calculated and stored in two
```
soulitzer: Maybe something more like "This tutorial explains the subtleties of requires_grad, retains_grad, leaf and non-leaf using a simple example..."

since I feel like that is the highlight of the tutorial rather than "dives into how those gradients are calculated"

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,593 @@
...
+different kinds of PyTorch tensors: leaf vs. non-leaf. It will also
```
soulitzer: "This tutorial will then use the retains_grad API to extract and visualize gradients across the a networks layers. This can be useful to debug issues like..."

Transition by explicitly mentioning retains_grad since part 1 explains how retains_grad works.

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,593 @@
...
+cover how we can extract and visualize gradients at any layer in the
+network’s computational graph. By inspecting how information flows from
+the end of the network to the parameters we want to optimize, we can
+debug issues that occur during training such as `vanishing or exploding
+gradients <https://arxiv.org/abs/1211.5063>`__.
+
+By the end of this tutorial, you will be able to:
```
soulitzer: This section can be removed if you explicitly mention what we want to highlight in the previous section.

advanced_source/visualizing_gradients_tutorial.py
```diff
@@ -0,0 +1,593 @@
...
+
+# prints False because tensor is the result of an operation with at
+# least one input having requires_grad=True
+print(f"{z.is_leaf=}")      
+
+
+######################################################################
+# The distinction between leaf and non-leaf determines whether the
+# tensor’s gradient will be stored in the ``grad`` property after the
+# backward pass, and thus be usable for gradient descent optimization.
```
soulitzer:
```suggestion
# backward pass, and thus be usable for gradient descent.
```

advanced_source/visualizing_gradients_tutorial.py
soulitzer: Although I guess in order to access the .grads you'd need to keep the intermediate forward outputs alive...

That seems pretty bad. Maybe we shouldn't use retains_grad for this tutorial LOL (we could say we're just using retains_grad for illustration purposes, but people probably want something pasteable they can use for their own models)

j-silv: [An alternative way](https://github.com/pytorch/pytorch/issues/61519#issuecomment-883524237) (which seems to be PyTorch's official recommendation) is to attach a forward hook to each module and then during the backward pass, attach a backward tensor hook to the output.

As far as I can tell, this is the only other sure-way to access intermediate gradients without modifying a `nn.Module` object. This is because any `nn.Module` with in-place operations will not allow any backward pass hook at the module level.

If we use @albanD 's suggestion though, is that better than keeping the intermediate forward outputs alive with `retain_grad()`? Is it just as inefficient?

soulitzer: I think we should use Alban's suggestion here. It should be more efficient than keeping the intermediates alive for sure!

j-silv:
@soulitzer here's my latest draft. Let me know if you would like anything changed!

j-silv:
@soulitzer sorry to bug you as I know you are busy! When you get a chance can you review these changes? I'd love to get the tutorial merged :)

soulitzer (commented):
Sorry for the delay, and thanks for your patience addressing the comments!
Overall, I think this is quite good! 
My only comment would be that it should actually be two separate tutorials, since now the second part is kind of unrelated to the first now. 
Would you mind splitting it? Sorry for the churn here.

j-silv:
> Sorry for the delay, and thanks for your patience addressing the comments! Overall, I think this is quite good! My only comment would be that it should actually be two separate tutorials, since now the second part is kind of unrelated to the first now. Would you mind splitting it? Sorry for the churn here.

No worries I can do that!

What difficulty do you think there should be for each tutorial? Also should I open up a separate PR for one of the splits?

soulitzer:
> What difficulty do you think there should be for each tutorial? Also should I open up a separate PR for one of the splits?

Maybe beginner for both? WDYT

> Also should I open up a separate PR for one of the splits?

Yes, thanks!

j-silv:
> Maybe beginner for both? WDYT

I took a look through most of the tutorials on the PyTorch docs and I think that beginner is reasonable. I could maybe see intermediate for the gradients one, but I'm on the fence. @sekyondaMeta do you think beginner for both?

soulitzer:
>  could maybe see intermediate for the gradients one, but I'm on the fence.

Either is fine to me!

soulitzer (approved):
LGTM!

j-silv:
Fixed lintrunner errors, should be good now

PR #3391, opened 2025-06-10
Update cpp custom ops docs with sycl extension and xpu support

ZhaoqiongZ:
Update cpp custom ops docs with sycl extension and xpu support

### DIFF
--- a/advanced_source/cpp_custom_ops.rst
+++ b/advanced_source/cpp_custom_ops.rst
@@ -1,7 +1,7 @@
 .. _cpp-custom-ops-tutorial:
 
-Custom C++ and CUDA Operators
-=============================
+Custom C++ and CUDA/SYCL Operators
+==================================
 
 **Author:** `Richard Zou <https://github.com/zou3519>`_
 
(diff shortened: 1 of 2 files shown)


advanced_source/cpp_custom_ops.rst
```diff
@@ -249,6 +254,70 @@ in a separate ``TORCH_LIBRARY_IMPL`` block:
     m.impl("mymuladd", &mymuladd_cuda);
   }
 
+If you also have a SYCL implementation of ``myaddmul``, you can also register it
```
dvrogozh:
```suggestion
If you also have a SYCL implementation of ``myaddmul``, you can register it
```

dvrogozh (changes requested):
@ZhaoqiongZ : this tutorial references https://github.com/pytorch/extension-cpp, right? Since you are updating it with https://github.com/pytorch/extension-cpp/pull/114, we need it to be merged first. I think it makes sense to add this to PR description and mark it as draft till it can actually be merged:
```
Requires: https://github.com/pytorch/extension-cpp/pull/114
```

zou3519 (changes requested):
I feel like a separate tutorial focused on xpu extensions, instead of trying to shoehorn xpu into this tutorial, would make things easier for a user to grok.

dvrogozh:
> I feel like a separate tutorial focused on xpu extensions, instead of trying to shoehorn xpu into this tutorial, would make things easier for a user to grok.

@zou3519, I disagree on that. By creating separate tutorial for SYCL extension we will effectively create duplicated tutorial as its significant part is dedicated to extension setup and configuration which is generic across backends with the differences coming when user wants to program device specific kernels. Trying to remove common part and refer to the existing C++/CUDA tutorial would fraction instructions and make instructions difficult to understand. Note that C++ part is also a need as this is typical case to target C++/CUDA or C++/SYCL at the same time. Further, a number of projects which currently rely on the API will want to consider to support more device options meaning that they will want to add another option (SYCL) to the existing code (C++/CUDA) and combined tutorial makes more sense in my mind. I suggest to modify existing C++/CUDA tuturial and clearly highlight that there are multiple device specific options to create kernels with custom ops. So tutorial will have a common part with C++ kernels and sections for CUDA, SYCL or other device specific backends shall they appear. Will you be ok with such proposal?

dvrogozh:
@albanD : we are trying to cover sycl extension support in pytorch CPP extension API by tutorials/examples and there are 2 different proposals on the plate:
* Extend existing tutorials/examples
* Create new

Could you, please, help provide your opinion on the matter? Who else should we invite for the discussion?

albanD:
I would expect this to be a separate tutorial indeed as most things are different and link to it from the main landing page.

advanced_source/cpp_custom_ops.rst
```diff
@@ -1,7 +1,7 @@
 .. _cpp-custom-ops-tutorial:
 
-Custom C++ and CUDA Operators
-=============================
+Custom C++ and CUDA/SYCL Operators
```
malfet, line 3: May be `Custom native extensions`?

advanced_source/cpp_custom_ops.rst
```diff
@@ -10,25 +10,30 @@ Custom C++ and CUDA Operators
     .. grid-item-card:: :octicon:`mortar-board;1em;` What you will learn
        :class-card: card-prerequisites
 
-       * How to integrate custom operators written in C++/CUDA with PyTorch
+       * How to integrate custom operators written in C++/CUDA/SYCL with PyTorch
```
malfet, line 13: Why don't we mention HIP/Metal in this list?

malfet (changes requested):
I think we need some sort of vendor neutral language here or make sure that all supported extension languages are mentioned in this tutorial

advanced_source/cpp_custom_ops.rst
ZhaoqiongZ: For HIP, they have claim in tutorial https://docs.pytorch.org/tutorials/advanced/cpp_custom_ops.html#cpp-custom-ops-tutorial that the code for CUDA will also work on AMD ROCm with no additional modifications.
For Metal, not sure if the feature cpp extension ready or not.

advanced_source/cpp_custom_ops.rst
ZhaoqiongZ: Yes we can use Custom native extension to replace the listed launguages and to avoid update from time to time,
But I do think in that way we should also maintain a clear table listed all the devices/launguages support this feature, right?

ZhaoqiongZ:
> I would expect this to be a separate tutorial indeed as most things are different and link to it from the main landing page.

I will create a seperate tutorial for Sycl extension usage

ZhaoqiongZ:
Hi @dvrogozh  @albanD  @malfet  @svekars , I created a standalone tutorial for sycl #3470 
Could you help review it?

PR #3400, opened 2025-06-18
[export] backed_size_oblivious tutorial

pianpwk:
Fixes #ISSUE_NUMBER

## Description
Adds an export tutorial section for avoiding 0/1 specialization

## Checklist

### DIFF
--- a/intermediate_source/torch_export_tutorial.py
+++ b/intermediate_source/torch_export_tutorial.py
@@ -489,6 +489,7 @@ except Exception:
 # specify 0/1 sample inputs when you'd like your program to hardcode them, and non-0/1 sample inputs when dynamic behavior is desirable. See what happens
 # at runtime when we export this linear layer:
 
+torch._logging.set_logs(dynamic=0)
 ep = export(
     torch.nn.Linear(4, 3),
     (torch.randn(1, 4),),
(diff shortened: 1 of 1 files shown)


intermediate_source/torch_export_tutorial.py
```diff
@@ -591,6 +592,30 @@ def forward(self, x, y):
     "bool_val": None,
 }
 
+######################################################################
+# (experimental) Avoiding 0/1 specialization
```
angelayi, line 596: should this be added after the section on 0/1 specialization? or move the section on 0/1 specialization to right before here?

intermediate_source/torch_export_tutorial.py
```diff
@@ -591,6 +592,30 @@ def forward(self, x, y):
...
+# ^^^^^^^^^^^^^^^^^^
+#
+# Export provides an experimental option to avoid specializing on size 0/1 sample inputs. Users can turn on `torch.fx.experimental._config.backed_size_oblivious = True` to enable this behavior.
```
angelayi, line 599:
```suggestion
# Export provides an experimental option to avoid specializing on size 0/1 sample inputs. Users can turn on ``torch.fx.experimental._config.backed_size_oblivious = True`` to enable this behavior.
```

intermediate_source/torch_export_tutorial.py
```diff
@@ -591,6 +592,30 @@ def forward(self, x, y):
...
+
+class Foo(torch.nn.Module):
+    def forward(self, x, y):
+        return x + y  # nothing special about size 0/1 here
+
+x = torch.randn(0, 1)
+y = torch.randn(1)
+dynamic_shapes = {"x": (Dim.AUTO, Dim.AUTO), "y": (Dim.AUTO,)}
+with torch.fx.experimental._config.patch(backed_size_oblivious=True):
+    ep = export(Foo(), (x, y), dynamic_shapes=dynamic_shapes)
```
angelayi, line 615: print ep?

intermediate_source/torch_export_tutorial.py
```diff
@@ -591,6 +592,30 @@ def forward(self, x, y):
...
+ep.module()(torch.randn(8, 1), torch.randn(1))
+ep.module()(torch.randn(5, 6), torch.randn(6))
+
```
angelayi, line 618: i wonder if it would be good to add an example where the behavior is weir after setting backed_size_oblivious. your example here is a little simple.

PR #3403, opened 2025-06-23
get rid of flask tutorials

msaroufim:
txt:flask ./_templates/layout.html:  var div = '<div class="admonition note"><p class="admonition-title">Note</p><p><i class="fa fa-flask" 
aria-hidden="true">&nbsp</i> This tutorial describes a prototype feature. Prototype features are typically not available as part of binary distributions like PyPI or 
Conda, except sometimes behind run-time flags, and are at an early stage for feedback and testing.</p></div>' ./.devcontainer/requirements.txt:flask 
./beginner_source/introyt/captumyt.py:  conda install pytorch torchvision captum flask-compress matplotlib=3.3.4 -c pytorch


.github/workflows/link_checkPR.yml
```diff
@@ -43,6 +49,10 @@ jobs:
       - name: Skip Message
         if: steps.skip-label.outputs.result == 'true'
         run: echo "Link check was skipped due to the presence of the 'skip-link-check' label."
+
+      - name: No Files to Check
+        if: steps.skip-label.outputs.result == 'false' && steps.changed-files.outputs.any_changed == 'false'
+        run: echo "No relevant files were changed in this PR that require link checking."
```
janeyx99: wait is this needed in this PR?

janeyx99 (approved):
lgtm other than the github workflows change

.github/workflows/link_checkPR.yml
msaroufim: Ah sorry miss from previous PR. It is indeed to ensure lint checking doesn't break on file deletions but we should merge it not here

PR #3404, opened 2025-06-23
delete seq2seq ts tutorial

msaroufim:
This is an outdated tutorial mostly talking about torchscript


janeyx99 (approved):
huh, surprising there are no references to it in the repo

PR #3410, opened 2025-06-24
[DOC] Set strict export explicitly for API change

CuiYifeng:
Align to https://github.com/pytorch/ao/pull/2430. The workflow of PT2E CPU/XPU has been broken by API change of `torch.export.export` and `torch.export.export_for_training`. This PR is to fix documents with setting `strict=True` explicitly.

cc @gujinghui @EikanWang @fengyuan14 @guangyey


CuiYifeng:
@ZhiweiYan-96 Please help review, thanks.

guangyey (approved):

PR #3419, opened 2025-06-26
Remove TorchScript part from the Saving and Loading tutorial

svekars:
Remove TorchScript part from the Saving and Loading tutorial

### DIFF
--- a/beginner_source/saving_loading_models.py
+++ b/beginner_source/saving_loading_models.py
@@ -227,42 +227,6 @@ functions to be familiar with:
 # normalization layers to evaluation mode before running inference.
 # Failing to do this will yield inconsistent inference results.
 #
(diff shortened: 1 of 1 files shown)


mikaylagawarecki:
cc @angelayi on whether there's anything we should add re export serialization in this tutorial

angelayi:
@mikaylagawarecki thanks for the bump! 
@svekars would it be possible to update this section to using export? It's very similar to TorchScript as you dont need to redefine the model class in order to run the model. Here are the docs for doing so: https://docs.pytorch.org/docs/main/export.html#serialization. Or I can do it in a separate PR!

svekars:
@angelayi  I added - can you take a look?

malfet (approved):

PR #3427, opened 2025-07-01
Update tutorials image to CUDA-12.6

malfet:
We should have update a while back when we've changed the default on PyPI


svekars (approved):

svekars:
@malfet thanks! Where do we track changes like this?

malfet:
> @malfet thanks! Where do we track changes like this?

I don't know...

PR #3431, opened 2025-07-07
Add custom class pt2 tutorial

angelayi:
Porting over part of https://docs.google.com/document/d/1xUfRsqyQp3CUqpHaPjG3APxx9gHEKjQ3ISYi5jtKpuk/edit?tab=t.0


advanced_source/custom_class_pt2.rst
```diff
@@ -0,0 +1,215 @@
...
+
+    ...
+
+That’s it! Now we can create a module that uses this object and run it with ``torch.compile`` or ``torch.export``:
+
+.. code-block::python
+
+    import torch
+
+    torch.ops.load_library("//caffe2/test:test_torchbind_cpp_impl")
```
svekars: is this internal stuff?

advanced_source/custom_class_pt2.rst
```diff
@@ -0,0 +1,215 @@
...
+
+1. Operators on real objects can be time consuming e.g. the custom object
+    might be reading from the network or loading data from the disk.
+2. We don’t want to mutate the real custom object or create side-effects to the environment while tracing.
+3. It cannot support dynamic shapes.
+
+However, it may be difficult for users to write a fake class: the original class
+uses some third-party library that determines the output shape of the methods,
+or is complicated and written by others. Besides, users may not care about the
+limitations listed above. In this case, please reach out to us!
```
ydwu4: Can also mention users can define x.tracing_mode(), which returns "real" so as to skip fakification.

ydwu4:
Thanks for porting this tutorial! Looks good to me.

ydwu4 (approved):

advanced_source/custom_class_pt2.rst
```diff
@@ -0,0 +1,215 @@
+Supporting Custom C++ Classes in PyTorch 2
+==========================================
+
+This tutorial is a follow-on to the
+:doc:`custom C++ classes <torch_script_custom_classes>` tutorial, and
+introduces additional steps that are needed to support custom C++ classes in
+PyTorch 2.
```
zou3519: torch.compile/torch.export. I don't think users understand the difference between 2 being a version and pt2 being the codename for torch.compile/torch.export.

ditto for the title.

advanced_source/custom_class_pt2.rst
zou3519: probably, I hope the OSS buck/bazel target for torch isn't caffe2 lol

advanced_source/custom_class_pt2.rst
```diff
@@ -0,0 +1,215 @@
...
+
```
zou3519, line 15: Do we have a sense of the stability of this? Does it always work now? If not, some warning about the rough edges around this (this is a "prototype" feature, right? or "unstable" in the current classification) would help

advanced_source/custom_class_pt2.rst
zou3519: So it would be nice if we had some sample code someone could download that had the C++ extension set up with it. Then it would be clearer exactly what needs to be loaded (via load_library, or an import statement). But I'm happy to ship some version of this as-is without that because it already helps alot

advanced_source/custom_class_pt2.rst
```diff
@@ -0,0 +1,215 @@
...
+    mod = ForEachAdd()
+    tq = empty_tensor_queue()
+    qlen = 10
+    for i in range(qlen):
+    tq.push(torch.zeros(1))
+
+    ep = torch.export.export(mod, (tq, torch.ones(1)), strict=False)
+
+Why do we need to make a Fake Class?
+------------------------------------
```
zou3519, line 248: We should try to list out all the caveats around tensor aliasing. Namely, we're gonna assume that tensors inside and outside a torchbind object are different. If you do things like mutate one of them, there will be undefined behavior.

PR #3437, opened 2025-07-08
Fix formatting to follow template for Profiler Recipe

sraikund16:
Fixes #ISSUE_NUMBER

## Description
Fixes the formatting to be the same as the template for tutorials 

## Checklist

cc @aaronenyeshi @chaekit


recipes_source/recipes/profiler_recipe.py
```diff
@@ -5,31 +5,31 @@
...
-use profiler to analyze model performance.
+# In this recipe, we will use a simple Resnet model to demonstrate how to
+# use profiler to analyze model performance.
 
-Prerequisites
----------------
-- ``torch >= 1.9``
+# Prerequisites
+# ---------------
+# - ``torch >= 1.9``
```
svekars: We should really set a higher version here probably

sraikund16: Ok let me change to 2.3.0 since we need that for one of the features

PR #3432, opened 2025-07-08
Deprecating pytorch quantization and FX tutorials

jerryzh168:
Summary:
Next step for https://github.com/pytorch/pytorch/issues/157591

PR #3438, opened 2025-07-08
Remove "Quantization Recipe"

andrewor14:
Refers to old `torch.quantization` APIs.


andrewor14:
Actually this is addressed in https://github.com/pytorch/tutorials/pull/3432

PR #3447, opened 2025-07-10
zero the gradients after updating weights

bintang-aswam:
Manually zero the gradients after updating weights by using machine epsilon for standard float (64-bit).

Fixes #ISSUE_NUMBER

## Checklist

### DIFF
--- a/beginner_source/examples_autograd/polynomial_autograd.py
+++ b/beginner_source/examples_autograd/polynomial_autograd.py
@@ -40,7 +40,7 @@ c = torch.randn((), dtype=dtype, requires_grad=True)
 d = torch.randn((), dtype=dtype, requires_grad=True)
 
 learning_rate = 1e-6
-for t in range(2000):
+for t in range(int(1/(learning_rate))):
     # Forward pass: compute predicted y using operations on Tensors.
     y_pred = a + b * x + c * x ** 2 + d * x ** 3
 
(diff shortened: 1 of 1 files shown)


albanD (commented):
Would you be able to share details on why do you want to do this change.
Setting them to None is equivalent to setting them to exactly 0. which is best no?

bintang-aswam:
> Would you be able to share details on why do you want to do this change. Setting them to None is equivalent to setting them to exactly 0. which is best no?
answer:
@albanD, In other words, setting them to `weight.grad = None` is equivalent to setting them to `weight.grad.zero_()`. However, there are few conceptual difference in the context of numerically zeroing between **zeroing the gradients** vs **tensor-multiplication of loss by machine-epsilon** as follows:
| Indicator                 | Tensor-Multiplication of Loss by Machine Epsilon                                          | Zeroing the Gradients                                                              |
|---------------------------|-------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------|
| Goal                      | to scale the loss to near-zero values                                                     | to clear up the accumulated gradients before the next backward pass                |
| Effect on Backpropagation | to generate gradients extremely small                                                     | to ensure gradients from previous steps don't interfere                            |
| Typical Use Case          | used to simulate vanishing gradients                                                      | mostly used for standard practice within training loops                            |
| Numerical Impact          | to test sensitivity of model to small perturbations including to test numerical stability | just reset gradients to None or literally zero (which is numerically meaningless). |

bintang-aswam:
> Would you be able to share details on why do you want to do this change. Setting them to None is equivalent to setting them to exactly 0. which is best no?

answer:
@albanD, In other words, setting them to `weight.grad = None` is equivalent to setting them to `weight.grad.zero_()`. However, there are few conceptual difference in the context of numerically zeroing between **zeroing the gradients** vs **tensor-multiplication of loss by machine-epsilon** as follows:
image

albanD:
Thanks for sharing these details.
I'm still not sure why this particular tutorial should be updated?
Maybe you want to do momentum-like updates here? If so, we can add a momentum term there?

PR #3461, opened 2025-07-14
Fix Parametrizations tutorial

svekars:
Fixes https://github.com/pytorch/tutorials/actions/runs/16273680500/job/45947334322?pr=3416#step:9:2088


intermediate_source/parametrizations.py
```diff
@@ -132,7 +132,7 @@ def forward(self, X):
 # Inspecting a parametrized module
 # --------------------------------
 #
-# When a module is parametrized, we find that the module has changed in three ways:
+# When a module is parametrized, we find that the module has changed in the following ways:
```
svekars:
```suggestion
# When a module is parametrized, we find that the module has changed in three ways:
```

PR #3462, opened 2025-07-14
Fix DQN w RNN tutorial

### LINKED ISSUE #3450: RL Tutorials failing against RC 2.8
It looks like some tutorials (RL) are failing against the RC 2.8 test with the error:
```
ModuleNotFoundError: No module named 'tensordict'
```
Possibly need an update to [tensordict in the requirements.txt](https://github.com/pytorch/tutorials/blob/main/.ci/docker/requirements.txt#L33)

You can find the RC branch test here: https://github.com/pytorch/tutorials/pull/3416

cc: @vmoens

svekars:
Fixes #3450
Fixes https://github.com/pytorch/tutorials/actions/runs/16273680500/job/45947334373?pr=3416#step:9:8137

### DIFF
--- a/.ci/docker/requirements.txt
+++ b/.ci/docker/requirements.txt
@@ -29,8 +29,8 @@ tensorboard
 jinja2==3.1.3
 pytorch-lightning
 torchx
-torchrl==0.7.2
-tensordict==0.7.2
+torchrl==0.9.2
+tensordict==0.9.1
 # For ax_multiobjective_nas_tutorial.py
 ax-platform>=0.4.0,<0.5.0
 nbformat>=5.9.2


svekars:
@vmoens ^^

intermediate_source/dqn_with_rnn_tutorial.py
```diff
@@ -389,7 +392,9 @@
 #   For the sake of efficiency, we're only running a few thousands iterations
 #   here. In a real setting, the total number of frames should be set to 1M.
 #
-collector = SyncDataCollector(env, stoch_policy, frames_per_batch=50, total_frames=200, device=device)
+collector = SyncDataCollector(
```
svekars, line 395:
```suggestion

collector = SyncDataCollector(
```

intermediate_source/dqn_with_rnn_tutorial.py
```diff
@@ -342,7 +342,10 @@
 # will return a new instance of the LSTM (with shared weights) that will
 # assume that the input data is sequential in nature.
 #
-policy = Seq(feature, lstm.set_recurrent_mode(True), mlp, qval)
+from torchrl.modules import set_recurrent_mode
+
+with set_recurrent_mode(True):
```
vmoens: This is to be set during the loss computation. 

The policy module given to the collector and to the loss will be the same. The decorator is to be set to True whenever a recurrent call is required (typically within the loss).

svekars:
> The decorator should be used when the policy is executed in the collector, not during instantiation of the policy object

Thanks, @vmoens. I just attempted that.

intermediate_source/dqn_with_rnn_tutorial.py
svekars: Can you check the update, @vmoens ?

vmoens:
Anything I can do for this or is it solved now?

PR #3467, opened 2025-07-16
[CI][lint] Add lintrunner + spaces, tabs, newlines linters

clee2000:
* Add lint job for lintrunner
* Add lintrunner.toml
* Add newlines, spaces, tabs lines
  * These download the linter from pytorch/pytorch from main as of right now.  Not sure if I should do this or just copy the linters over
  * Big exclusion lists so that the job will be green, then we can make the changes and remove them from the exclusion list later


svekars (approved):
thank you!

PR #3468, opened 2025-07-16
[ez][BE] Apply lintrunner to non tutorials files

clee2000:
Removes the files from the exclusion lists and run `lintrunner --all-files -a` to apply the changes

This is only files at the root level and in `.github` and `.jenkins` folders

The linters should just be whitespace changes


clee2000:
nvm i think it was transient

svekars (approved):

PR #3472, opened 2025-07-18
Update tutorial of device mesh to use fsdp2

fduwjj:
## Description
Since we are deprecating FSDP1 we need to update the place where FSDP1 is used to use FSDP2 in the tutorial of DeviceMesh.

## Checklist


wz337 (approved):
LGTM! Thanks for updating!

PR #3474, opened 2025-07-18
Upgrade torchrl and tensordict requirements

vmoens:
@svekars I think we should upgrade torchrl and tensordict - 0.7 is quite old for both!


vmoens:
Looks like tests are passing! 🥳

PR #3494, opened 2025-07-28
Update index.rst - missing cheatsheet?

doomdagadiggiedahdah:
This commit is less to suggest the linked cheatsheet but instead to point out there is no ptcheat.html in the repo and wasn't sure if there was a suggested one already.

I checked in these repos for "cheat" and nothing came up: Pytorch, examples, tutorials, torchrec.

Fixes #ISSUE_NUMBER

## Description
linked to another cheatsheet

## Checklist


svekars:
thanks, we will fix this. Please next time file an issue instead.

PR #3513, opened 2025-08-06
Update redirects.py

### LINKED ISSUE #3512: Redirect for prototype/ -> unstable/
### 🚀 Describe the improvement or the new tutorial

When I search ["flight recorder pytorch" on Google](https://www.google.com/search?q=pytorch+flight+recorder&sca_esv=56a8724cb68766c6&ei=_7yTaKLqN4ra5NoP38nhqAg&oq=pytorch+flight+recorder&gs_lp=Egxnd3Mtd2l6LXNlcnAiF3B5dG9yY2ggZmxpZ2h0IHJlY29yZGVyKgIIADIIEAAYgAQYsAMyCRAAGLADGAgYHjILEAAYsAMYCBgKGB4yDhAAGIAEGLADGIYDGIoFMg4QABiABBiwAxiGAxiKBTIOEAAYgAQYsAMYhgMYigUyCBAAGLADGO8FMggQABiwAxjvBTILEAAYgAQYsAMYogRIngtQAFgAcAF4AJABAJgBAKABAKoBALgBAcgBAJgCAaACA5gDAIgGAZAGCZIHATGgBwCyBwC4BwDCBwMyLTHIBwM&sclient=gws-wiz-serp)

The top link is https://docs.pytorch.org/tutorials/prototype/flight_recorder_tutorial.html

whereas now these tutorials are under https://docs.pytorch.org/tutorials/unstable/flight_recorder_tutorial.html

Can we add a redirect? I knew that the tutorial should be there since i actively work on PyTorch, but for new users this will be confusing and also when `prototype/` is in URL path, the site doesn't have any CSS and is scary-looking

_No response_

### Additional context

_No response_

sekyondaMeta:
Fixes #3512

### DIFF
--- a/redirects.py
+++ b/redirects.py
@@ -13,6 +13,7 @@ redirects = {
     "intermediate/dynamic_quantization_bert_tutorial.html": "../index.html",
     "intermediate/quantized_transfer_learning_tutorial.html": "../index.html",
     "prototype/graph_mode_dynamic_bert_tutorial.html": "../index.html",
+    "prototype/flight_recorder_tutorial.html": "https://docs.pytorch.org/tutorials/unstable/flight_recorder_tutorial.html"
     "prototype/fx_graph_mode_ptq_dynamic.html": "../index.html",
     "prototype/fx_graph_mode_ptq_static.html": "../index.html",
     "prototype/fx_graph_mode_quant_guide.html": "../index.html",
--- a/redirects.py
+++ b/redirects.py
@@ -13,7 +13,7 @@ redirects = {
     "intermediate/dynamic_quantization_bert_tutorial.html": "../index.html",
     "intermediate/quantized_transfer_learning_tutorial.html": "../index.html",
     "prototype/graph_mode_dynamic_bert_tutorial.html": "../index.html",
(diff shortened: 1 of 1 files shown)


AlannaBurke:
LGTM! 👍

redirects.py
```diff
@@ -13,6 +13,7 @@
     "intermediate/dynamic_quantization_bert_tutorial.html": "../index.html",
     "intermediate/quantized_transfer_learning_tutorial.html": "../index.html",
     "prototype/graph_mode_dynamic_bert_tutorial.html": "../index.html",
+    "prototype/flight_recorder_tutorial.html": "https://docs.pytorch.org/tutorials/unstable/flight_recorder_tutorial.html"
```
H-Huang: The same issue applies to the other tutorials in the URL `prototype/` URL path (e.g. https://docs.pytorch.org/tutorials/prototype/context_parallel.html)

Though, when I search context parallel on Google it doesn't show this path, but maybe a more systemic fix is to redirect everything under prototype?

sekyondaMeta: I think part of the problem is that not everything previously under prototype moved to unstable. It looks like some things no longer exist, some moved to other folders and some moved to unstable. So it will be tough to try and catch all and send it the correct way

PR #3453, opened 2025-07-10
Remove references to TorchScript in docs

gmagogsfm:
* Removes references to TorchScript
* Replace it with a deprecation warning and redirect to torch.export when appropriate

PR #3546, opened 2025-09-08
Add custom class tutorial back

angelayi:
Reverts part of https://github.com/pytorch/tutorials/pull/3453 which removed the Custom Class tutorial

Custom classes are a generic concept in PyTorch. Although it was developed for TorchScript, it is also supported in Pytorch 2.0 as many users (like TensorRT and internal Ads frameworks) rely on it. Therefore, we should not delete this tutorial. 

Additionally the [Custom Class x PT2 tutorial](https://docs.pytorch.org/tutorials/advanced/custom_class_pt2.html) also references this tutorial, although CI did not seem to catch the broken link (cc @svekars)

As part of revert, I removed references to TS.

### DIFF
--- a/advanced_source/custom_class_pt2.rst
+++ b/advanced_source/custom_class_pt2.rst
@@ -3,7 +3,7 @@ Supporting Custom C++ Classes in torch.compile/torch.export
 
 
 This tutorial is a follow-on to the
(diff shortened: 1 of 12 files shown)


svekars (approved):
SGTM - fix lint?

PR #3566, opened 2025-09-16
[ONNX] Update PyTorch version and clean up tutorial

justinchuby:
Updated prerequisite version of PyTorch and removed optimization section because optimization is now default.

@svekars @titaiwangms @xadupre


titaiwangms:
@svekars

PR #3571, opened 2025-09-16
Update pytorch_sphinx_theme2

svekars:
Updated the PyTorch Sphinx theme version in requirements.


AlannaBurke:
Just for future reference, why do we have to call the theme this way, with the hash?

svekars:
> Just for future reference, why do we have to call the theme this way, with the hash?

Because we pin to a specific commit.

sekyondaMeta (approved):

PR #3603, opened 2025-09-30
Fix async cp recipe

Saiteja64:
## Description
set checkpoint_future properly in pinned mem tutorial


ekr0 (approved):
I tested this works, the recipe no longer hangs on join.

PR #3608, opened 2025-10-05
Update semi_structured_sparse.py

catswe:
In my environment with an A100, I only see the speedup when enabling the following line

SparseSemiStructuredTensor._FORCE_CUTLASS = True

### DIFF
--- a/advanced_source/semi_structured_sparse.py
+++ b/advanced_source/semi_structured_sparse.py
@@ -55,6 +55,9 @@ import torch
 from torch.sparse import to_sparse_semi_structured, SparseSemiStructuredTensor
 from torch.utils.benchmark import Timer
 
+# the following line may need to be enabled to see a speedup
+# SparseSemiStructuredTensor._FORCE_CUTLASS = True
+
 # mask Linear weight to be 2:4 sparse
 mask = torch.Tensor([0, 0, 1, 1]).tile((3072, 2560)).cuda().bool()
 linear = torch.nn.Linear(10240, 3072).half().cuda().eval()


catswe:
If I run without forcing CUTLASS and use CUSPARSELT instead, the speedup can't be observed. 

```python
import torch
from torch.sparse import to_sparse_semi_structured, SparseSemiStructuredTensor
from torch.utils.benchmark import Timer
import torch
print("torch:", torch.__version__, "cuda:", torch.version.cuda)
print("gpu:", torch.cuda.get_device_name(0), "cc:", torch.cuda.get_device_capability(0))

# mask Linear weight to be 2:4 sparse
mask = torch.Tensor([0, 0, 1, 1]).tile((3072, 2560)).cuda().bool()
linear = torch.nn.Linear(10240, 3072).half().cuda().eval()
linear.weight = torch.nn.Parameter(mask * linear.weight)
print(linear.weight)

x = torch.rand(3072, 10240).half().cuda()

with torch.inference_mode():
    dense_output = linear(x)
    dense_t = Timer(stmt="linear(x)",
                    globals={"linear": linear,
                             "x": x}).blocked_autorange().median * 1e3

    # accelerate via SparseSemiStructuredTensor
    linear.weight = torch.nn.Parameter(to_sparse_semi_structured(linear.weight))
    print(linear.weight)

    sparse_output = linear(x)
    sparse_t = Timer(stmt="linear(x)",
                    globals={"linear": linear,
                             "x": x}).blocked_autorange().median * 1e3

    # sparse and dense matmul are numerically equivalent
    # On an A100 80GB, we see: `Dense: 0.870ms Sparse: 0.630ms | Speedup: 1.382x`
    assert torch.allclose(sparse_output, dense_output, atol=1e-3)
    print(f"Dense: {dense_t:.3f}ms Sparse: {sparse_t:.3f}ms | Speedup: {(dense_t / sparse_t):.3f}x")
```

```
torch: 2.8.0+cu126 cuda: 12.6
gpu: NVIDIA A100-SXM4-40GB cc: (8, 0)
Parameter containing:
tensor([[-0.0000,  0.0000,  0.0036,  ..., -0.0000,  0.0042,  0.0057],
        [-0.0000,  0.0000, -0.0010,  ...,  0.0000,  0.0083, -0.0043],
        [-0.0000, -0.0000, -0.0054,  ...,  0.0000, -0.0007, -0.0020],
        ...,
        [ 0.0000, -0.0000,  0.0007,  ..., -0.0000,  0.0033,  0.0074],
        [ 0.0000, -0.0000, -0.0079,  ..., -0.0000,  0.0071, -0.0024],
        [-0.0000, -0.0000, -0.0018,  ..., -0.0000, -0.0079, -0.0042]],
       device='cuda:0', dtype=torch.float16, requires_grad=True)
SparseSemiStructuredTensorCUSPARSELT(shape=torch.Size([3072, 10240]))
Dense: 0.803ms Sparse: 1.057ms | Speedup: 0.760x
```

PR #3609, opened 2025-10-06
Control Flow Operator Tutorial

bohnstingl:
## Description
This PR aims to add a tutorial of the recent control flow operators.

## Checklist

cc @ydwu4

### DIFF
--- a/intermediate_source/torch_controlflow_tutorial.py
+++ b/intermediate_source/torch_controlflow_tutorial.py
@@ -356,12 +356,25 @@ class LSTM_scan(torch.nn.Module):
             c_new = f * c + i * g
             h_new = o * torch.tanh(c_new)
             
-            # return (h_new, c_new.clone()), h_new.clone()
-            return (h_new, c_new + 0.), h_new + 0.
+            return (h_new, c_new.clone()), h_new.clone()
         
         carry, outs = scan(lstm_combine, init, xs, dim=0)
-        
         return carry, outs
+    
+class LSTM_scan_cell(torch.nn.Module):
+    def __init__(self, state_dict):
+        super(LSTM_scan_cell, self).__init__()
+        self.lstm_cell = torch.nn.LSTMCell(input_size, hidden_size)
+        self.lstm_cell.load_state_dict(state_dict)
+        
+    def forward(self, init: torch.Tensor, xs: torch.Tensor):
+        def lstm_combine(carry, x):
+            hx, cx = self.lstm_cell(x, carry)
+            return (hx, cx.clone()), hx.clone()
+        
+        carry, outs = scan(lstm_combine, init, xs, dim=0)
+        return carry, outs
+    
 print('='*80)
 print('Example: RNN with scan')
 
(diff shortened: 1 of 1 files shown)


intermediate_source/torch_controlflow_tutorial.py
```diff
@@ -0,0 +1,504 @@
...
+**Authors:** Yidi Wu, Thomas Ortner, Richard Zou, Edward Yang, Adnan Akhundov, Horace He and Yanan Cao
+
+This tutorial introduces the PyTorch Control Flow Operators: 
+``cond``, ``while_loop``, ``scan``, ``associative_scan``, and ``map``. 
+These operators enable data-dependent control flow to be expressed in a functional, 
+differentiable, and exportable manner. 
+
+The tutorial is split into three parts:
+
+Part 1: Basic Inference Examples
```
ydwu4, line 14: Thanks for writing this! For part 1 and part 2, I feel this tutorial does not add much value compared to what we have in the doc string. For part 3, users probably know how to do autograd in PyTorch and it feels to me just a demo that we could do autograd.

Instead, maybe we should just focus on Part 2,  explore the killer examples in depth and show the benefits of using control flow operators in that particular setting e.g. using scan for chunked loss, rnn and associative scan for SSMs and try to get convincing e2e benchmarking results.

bohnstingl: Thank you @ydwu4 for looking at it.

About part 1, to me it still feels instructive and my impression is that not too many people are aware of the control flow operators, hence showing some basic usecases may make sense?

I guess we can remove part 3 entirely and then expand more thoroughly in part 2. The RNN and SSM examples are already there, but I can definitely expand more on them. I will also add the chunked loss computation that you mentioned. 

What I thought would be interesting to show also are the cases where the control flow ops are actually needed in order to have a compilable model, e.g., show that without them there would be graph breaks. Early stopping based on data-dependent conditions would be another example, aka while_loop. WDYT?

ydwu4: Yeah, sounds good. For chunked loss, I'm doing some benchmarks against liger so you can focus on rnn/SSMs. Graph breaks are not the fundamental reasons why we need control flow operators unless you cannot have graph breaks e.g. in export. It doesn't necessarily have worse performance when the model has graph breaks (at least we need to prove it makes perf worse in important use cases). IMO it's important to show control flow operators' values from perf/usability side compared with eager, torch.compile with native control and with hand-tuned kernels.

bohnstingl: But what about early stopping? Wouldn't you need data-dependent operations that get compiled somehow?

ydwu4: can compile the body and have multiple torch.compiles in your program where you want to accelerate.

bohnstingl:
@ydwu4, I've removed the part 3 and enhanced the `scan` example in part 2. However, the performance is not as expected.
In particular, one problem arises when comparing scan to the for-loop approach. Although the compile time of `scan` is clearly much faster, the runtime is actually slower.

I would be more than happy if you can help me investigate that. This is also linked to the dedicated benchmark scripts that I have started [here](https://github.com/pytorch/pytorch/pull/164859).

PR #3619, opened 2025-10-22
Remove unnecessary character from beginner's autograd example

sonnyding1:
## Description
In the "Learning PyTorch with Examples" tutorial, the autograd code block started with `rimport`. This PR fixed the typo by removing `r`.

### DIFF
--- a/beginner_source/examples_autograd/polynomial_autograd.py
+++ b/beginner_source/examples_autograd/polynomial_autograd.py
@@ -1,4 +1,4 @@
-r"""
+"""
 PyTorch: Tensors and autograd
 -------------------------------
 


sekyondaMeta (approved):

sonnyding1:
> r prefix in Python means raw string, i.e. one where does not need to escape `\` I.e.

Thank you for the heads up, then it should be the problem of the `includenodoc` directive not parsing the docstring correctly. It seems `includenodoc` used to be defined in a file `custom_directives.py`, as mentioned in #550, but the file was removed, and the definition of `includenodoc` is moved outside of this repository.

Do you know where the includenodoc directive is maintained now? I'd be happy to create an issue and a pull request over there to fix this bug.

svekars:
I guess we can update the regex here: https://github.com/pytorch/pytorch_sphinx_theme/blob/pytorch_sphinx_theme2/pytorch_sphinx_theme2/custom_directives.py#L39 to accommodate for raw strings. Not sure why it worked before with the old theme. `includenodoc` only used in that one file.

sonnyding1:
Thank you, I'll close this PR and submit another over there.

PR #3657, opened 2025-11-24
Fix retain_grad() documentation for leaf vs non-leaf tensors

### LINKED ISSUE #3652: retain_grad() documentation misleading about non-leaf tensors
The documentation says:

"If you call retain_grad() on a non-leaf node, it results in a no-op."

This is misleading or incomplete.
retain_grad() on a non-leaf tensor with requires_grad=True does not result in a no-op — it correctly retains gradients.

Only calling retain_grad() on:

a leaf tensor is a no-op

a tensor with requires_grad=False throws an error

patrocinio:
The documentation incorrectly stated that calling retain_grad() on a non-leaf node results in a no-op. This is misleading because:

- retain_grad() on a non-leaf tensor with requires_grad=True correctly retains gradients (not a no-op)
- retain_grad() on a leaf tensor is a no-op (already retains by default)
- retain_grad() on a tensor with requires_grad=False throws an error

Updated the documentation and summary table to accurately reflect this behavior.

Fixes #3652

## Checklist

cc @jbschlosser

### DIFF
--- a/intermediate_source/torch_compile_tutorial.py
+++ b/intermediate_source/torch_compile_tutorial.py
@@ -163,7 +163,7 @@ def timed(fn):
     result = fn()
     end.record()
     torch.cuda.synchronize()
-    return result, start.elapsed_time(end) / 1024
+    return result, start.elapsed_time(end) / 1000
 
 
 inp = torch.randn(4096, 4096).cuda()


svekars:
cc: @j-silv

soulitzer (approved):
Thanks!

PR #3697, opened 2025-12-11
Tutorial for DebugMode

yushangdi:
A tutorial for DebugMode

cc @pianpwk


recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+"""
+
+######################################################################
+# Overview
+# --------
+#
+# ``DebugMode`` (:class:`torch.utils._debug_mode.DebugMode`) is a
+# ``TorchDispatchMode`` that intercepts PyTorch runtime calls and emits a
+# hierarchical log of operations. It is particularly useful when you need to
+# understand *what* actually ran, both in eager mode and under ``torch.compile``
```
Lucaskabela: nit: "*what* actually runs" 

I would prefer present tense here as you run DebugMode as a context manager actively

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+# or when you need to pinpoint numerical divergence between two runs.
+#
+# Key capabilities:
+#
+# * **Runtime logging** – Records dispatched operations and TorchInductor compiled
+#   Triton kernels.
+# * **Tensor hashing** – Attaches deterministic hashes to inputs/outputs so you
```
Lucaskabela: nit: `to enable diffing runs to locate numerical divergences`

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+#   can diff runs and locate numerical divergences.
+# * **Dispatch hooks** – Lets you register custom hooks to annotate each call
```
Lucaskabela: `Allows registration of custom hooks to annotate calls`

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+    a = torch.compile(f)(x)
+
+print("Triton in DebugMode logs:")
+print(debug_mode.debug_string())
+
+######################################################################
+# Numerical debugging with tensor hashes
+# --------------------------------------
+#
+# If you have some numerical divergence between modes, you can use DebugMode to find where the
```
Lucaskabela: remove `some`

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+# numerical divergence originates.
+# In the example below, you can see that all tensor hashes are the same for eager mode and compiled mode.
+# If any hash looks different, then that's where the numerical divergence is coming from.
```
Lucaskabela: nit: `If any hash is different, `

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+        DebugMode._annotate("Foo")
+        return self.l1(x)
+
+mod = Foo()
+with DebugMode(record_nn_module=True) as debug_mode:
+    DebugMode._annotate("forward")
+    mod(x)
+
+print("DebugMode output with annotation:")
+print(debug_mode.debug_string())
```
Lucaskabela: This looks pretty good! Can we add a closing paragraph/some closing thoughts as well?

Lucaskabela (changes requested):

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+
+######################################################################
+# Overview
+# --------
+#
+# ``DebugMode`` (:class:`torch.utils._debug_mode.DebugMode`) is a
+# ``TorchDispatchMode`` that intercepts PyTorch runtime calls and emits a
+# hierarchical log of operations. It is particularly useful when you need to
+# understand *what* actually ran, both in eager mode and under ``torch.compile``
+# or when you need to pinpoint numerical divergence between two runs.
```
pianpwk, line 32: can we call this out as a prototype/WIP somewhere lol

recipes_source/debug_mode_tutorial.py
pianpwk: hm why "deterministic"? if we say this we should mention somewhere collectives aren't deterministic (yet).

recipes_source/debug_mode_tutorial.py
pianpwk: `+ with arbitrary information`

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+        record_output=True,
+        record_stack_trace=True,
+        record_ids=True,
+    ) as debug_mode,
+    DebugMode.log_tensor_hashes(
+        hash_fn=["norm", "hash_tensor"],
+        hash_inputs=True,
+    ),
+):
+    result = run_once()
```
pianpwk, line 94: - record_output should be on by default now
- record_stack_trace I'm a bit wary of advertising, since it's not ideal for compile right now, but I think that's fine
- do you think we need ["norm", "hash_tensor"]? I think just norm (not specifying) might be fine now; this might be information overload. But happy either way.

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+######################################################################
+# Getting more metadata
+# -----------
+#
+# For most investigations, you'll want to enable stack traces, tensor IDs, and tensor hashing.
+# These features provide metadata to correlate operations back to model code.
+#
+# ``DebugMode.log_tensor_hashes`` decorates the log with hashes for every call.
+# The ``hash_tensor`` hash function uses ``torch.hash_tensor``, which returns 0 for tensors whose
+# elements are all the same. The ``norm`` hash function uses ``norm`` with ``p=1``.
```
pianpwk, line 79: I would call out that with both these functions, tensor closeness in numerics -> hash closeness, so it's rather interpretable (especially norm, IMO)

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+
+######################################################################
+# Interpreting the log
+# --------------------
+#
+# Each line follows ``op(args) -> outputs``. When ``record_ids`` is enabled,
+# tensors are suffixed with ``$<id>`` and DTensors are labeled ``dt``.
+#
+# Indentation generally reflects the dynamic call stack, but it's not guaranteed to be the same
+# as the call stack at runtime, especially for DTensor calls.
```
pianpwk: hmm I don't know if we need to mention this, indentation is mainly a concern with nn.Module & DTensors, and we show any DTensor examples

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+# If you have some numerical divergence between modes, you can use DebugMode to find where the
+# numerical divergence originates.
+# In the example below, you can see that all tensor hashes are the same for eager mode and compiled mode.
+# If any hash looks different, then that's where the numerical divergence is coming from.
+
+def run_model(model, data, *, compile_with=None):
+    if compile_with is not None:
+        model = torch.compile(model, backend=compile_with)
+    with DebugMode(record_output=True) as dm, DebugMode.log_tensor_hashes(
+        hash_fn=["norm"],
```
pianpwk: don't need to specify hash_fn

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+        return torch.relu(x).mm(x.T)
+
+inputs = (torch.randn(4, 4),)
+dm_eager, _ = run_model(Toy(), inputs)
+dm_compiled, _ = run_model(Toy(), inputs, compile_with="aot_eager")
+
+print("Eager mode:")
+print(dm_eager.debug_string())
+print("Compiled aot_eager mode:")
+print(dm_compiled.debug_string())
```
pianpwk, line 160: this doesn't diverge right?

Not sure what the expected output here is, but I think it's hard to find an example with divergence.

I think it's fine to present this eager <-> compiled version where the output is bitwise equivalent, but maybe the intermediate ops are different (e.g. addmm vs linear), so logs aren't 1-1, and comment on that.

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+######################################################################
+# Custom dispatch hooks
+# ---------------------
+#
+# Hooks allow you to annotate each call with custom metadata such as GPU memory usage. ``log_hook`` returns a mapping
+# that is rendered inline with the debug string.
+
+MB = 1024 * 1024.0
```
pianpwk, line 220: We should call out somewhere that with all context managers: `DebugMode(), log_tensor_hashes(), dispatch_hooks()`, you can wrap them around any particular region (except inside compiled regions). So the wrappers aren't global.

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+        self.l1 = torch.nn.Linear(8, 8)
+
+    def forward(self, x):
+        DebugMode._annotate("Foo")
+        return self.l1(x)
+
+mod = Foo()
+with DebugMode(record_nn_module=True) as debug_mode:
+    DebugMode._annotate("forward")
+    mod(x)
```
pianpwk: lol I don't think we should mention this; we haven't been testing or using it at all

pianpwk:
overall looks great!

recipes_source/debug_mode_tutorial.py
yushangdi: hmm, i think the "except inside compiled regions" part can be confusing, so I'd prefer to leave this one out.

recipes_source/debug_mode_tutorial.py
yushangdi: Added!

recipes_source/debug_mode_tutorial.py
yushangdi: added a note

recipes_source/debug_mode_tutorial.py
yushangdi: removed reocrd_output, I added a note that `record_stack_trace` only works for eager now. And I removed `"hash_tensor"` from the code, but I still mention it in the text.

recipes_source/debug_mode_tutorial.py
yushangdi: added

recipes_source/debug_mode_tutorial.py
yushangdi: removed

recipes_source/debug_mode_tutorial.py
yushangdi: removed.

recipes_source/debug_mode_tutorial.py
yushangdi: I added an example where I provide a wrong decomposition and the logs diverge. cc @Lucaskabela also commented on this

recipes_source/debug_mode_tutorial.py
yushangdi: removed.

svekars (changes requested):
Please include into a relevant toctree in one of the related .rst file. And please don't merge until the doc team does the review. We typically merge closer to the release date.

yushangdi:
> Please include into a relevant toctree in one of the related .rst file.

sure, added now

svekars:
Nightly Preview is here: https://docs-preview.pytorch.org/pytorch/tutorials-nightly-preview/3697/recipes/debug_mode_tutorial.html

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,265 @@
+# -*- coding: utf-8 -*-
+
+"""
+DebugMode: Recording Dispatched Operations and Numerical Debugging
+=================================================================
+
+**Authors:** Pian Pawakapan, Shangdi Yu
+"""
+
+######################################################################
```
AlannaBurke, line 24: Please include the prerequisites as well as what's going to be covered in the tutorial. See template: https://github.com/pytorch/tutorials/blob/main/beginner_source/template_tutorial.py

recipes_source/debug_mode_tutorial.py
```diff
@@ -0,0 +1,265 @@
...
+    _ = mod(inp)
+
+print("DebugMode output with stack traces and module boundaries:")
+print(debug_mode.debug_string(show_stack_trace=True))
+
+######################################################################
+# Conclusion
+# ----------
+#
+# DebugMode gives you a lightweight, runtime-only view of what PyTorch actually
```
AlannaBurke: Can you edit this to be more along the lines of "this is what we covered in this tutorial?"

yushangdi: sure, updated

recipes_source/debug_mode_tutorial.py
yushangdi: sure, added

Lucaskabela (approved):

PR #3698, opened 2025-12-11
Fix gallery pattern regex?

clee2000:
Regex was generated by chatgpt, I haven't figured out lookaheads/behinds yet

also briefly checked with hyperparameter tuning tutorial (didn't run the entire thing so idk about the rest of the tutorial generation but I did check that it probably didn't run anything else)

shangdi reported the old pattern wasn't working, but this one seems to


yushangdi:
can confirm this works for me!

PR #3700, opened 2025-12-16
Adding Glossary

AlannaBurke:
## Description
Term list: glossary.rst
JS is used to load the terms for the tooltips

Screenshot 2025-12-15 at 10 01 31 PM (3)

Started with a limited list of terms and updated a handful of pages. Once we're happy with it i'll add more terms and update the rest of the files. 

## Checklist

### DIFF
--- a/.ci/docker/requirements.txt
+++ b/.ci/docker/requirements.txt
@@ -17,8 +17,7 @@ pandocfilters==1.5.1
 markdown==3.8.2
 
 # PyTorch Theme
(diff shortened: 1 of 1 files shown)


advanced_source/cpp_export.rst
```diff
@@ -1,3 +1,3 @@
 .. warning::
-    TorchScript is deprecated, please use
```
svekars, line 2: this is a redirect page - why are we adding a term here?

PR #3702, opened 2025-12-17
Adds a GitHub Action to notify when preview link is live

AlannaBurke:
## Description
Adds a GitHub Action to notify when preview link is live. Waits up to 40 minutes. I did test this locally to the extent that i could.

## Checklist

### DIFF
--- a/beginner_source/onnx/export_control_flow_model_to_onnx_tutorial.py
+++ b/beginner_source/onnx/export_control_flow_model_to_onnx_tutorial.py
@@ -28,12 +28,11 @@ Export a model with control flow to ONNX
 #
 # - How to refactor the model to use :func:`torch.cond` for exporting.
 # - How to export a model with control flow logic to ONNX.
-# - How to optimize the exported model using the ONNX optimizer.
 #
 # Prerequisites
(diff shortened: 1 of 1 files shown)


AlannaBurke:
Looks like there is an unrelated error, i took a stab at fixing it here but i haven't gotten it to build yet: https://github.com/pytorch/tutorials/pull/3704

conf.py
```diff
@@ -182,7 +182,7 @@ def wrapper(*args, **kwargs):
         "# https://docs.pytorch.org/tutorials/beginner/colab\n"
         "%matplotlib inline"
     ),
-    "ignore_pattern": r"_torch_export_nightly_tutorial.py",
```
svekars, line 185: not related to this PR, right?

lychee.toml
```diff
@@ -1 +1 @@
-exclude_path = [".jenkins/build.sh", "_static/img/", "_static/images/"]
```
svekars, line 1: we don't need this.

conf.py
AlannaBurke: I just added this to see if it would pass

PR #3717, opened 2026-01-04
Clarify explanation of requires_grad in PyTorch

### LINKED ISSUE #3716: 💡 [REQUEST] - Clarification of requires_grad about beginner/nn_tutorial.html
### 🚀 Describe the improvement or the new tutorial

It was challenging for me to initially grasp why `requires_grad` was done *after* weights, but *in the same line* as bias under https://docs.pytorch.org/tutorials/beginner/nn_tutorial.html#neural-net-from-scratch-without-torch-nn

At first glance, the code looks inconsistent:

1. `weights` initialization is split into two lines.
2. `bias` initialization is done in one line.

**The Logic Gap**
The tutorial currently explains *that* we do it, but not exactly *why* the distinction exists between these two specific variables.

* **The Bias** is created using a factory function (`torch.zeros`) with no subsequent mathematical operations. It is born as a "Leaf Node" (a source parameter).
* **The Weights** involve a mathematical operation (`/ math.sqrt(...)`). If we set `requires_grad=True` inside `torch.randn()`, PyTorch records the division as a computational step. The resulting `weights` variable becomes a **non-leaf node** (a calculated outcome), which the optimizer cannot update.

**Proposed Improvement**
I propose modifying the comment block to explicitly mention that `requires_grad` must be deferred until *after* the initialization math is complete to preserve the tensor as a trainable parameter (Leaf Node).

* https://docs.pytorch.org/tutorials/beginner/nn_tutorial.html
* https://docs.pytorch.org/tutorials/beginner/nn_tutorial.html#neural-net-from-scratch-without-torch-nn

### Additional context

Image

FlightVin:
Fixes #3716

## Description
It was challenging for me to initially grasp why `requires_grad` was done after `weights` initialization, but in the same line as `bias`.

The existing explanation ("we don't want that step included in the gradient") is technically correct but omits the practical consequence: **Leaf Node status.**

If `requires_grad=True` is set before the initialization math (the division by `sqrt(n)`), the `weights` tensor records that operation and becomes a calculated output (non-leaf node) rather than a source parameter. This makes it impossible for optimizers to update it.

This PR clarifies that we set `requires_grad` *after* the math to ensure the tensor remains a trainable **Leaf Node**.

## Checklist

P.S. Leaving the 3rd point unchecked since there are no label as of yet in the issue. 

cc @svekars @sekyondaMeta @AlannaBurke @albanD @jbschlosser


FlightVin:
_random_: Should I also link [this](https://docs.pytorch.org/tutorials/beginner/understanding_leaf_vs_nonleaf_tutorial.html)?

albanD (approved):
Sounds fair. Thanks!

sekyondaMeta:
@FlightVin thanks for this update. Can you fix the lint issues from the failing lint check.

FlightVin:
Forgot this wasn't work lol, I'll request reviews from the UI, sorry for the tag :sob:

AlannaBurke (approved):

PR #3597, opened 2025-09-27
Switch from sin(x) to exp(x) as the target function in polynomial_autograd 

leomilano:
Fixes #3596

## Description
This PR  branch uses 'exp(x)' instead of 'sin(x)' as the 'y' (target) variable. The reason for this is faster and clearer convergence, since exp(x) is well approximated by a Taylor expansion around the origin. Also, all the derivatives are exp(0)=1, so the polynomial coefficients are trivially '1/n! '(for the 'n-th' term).

I am also improving the information output to the user inside the optimization loop.

Finally, I am making the top doc string a raw string, to avoid a python3 warning about it.

## Checklist

- [x] The issue that is being fixed is referred in the description (see above "Fixes #ISSUE_NUMBER")
- [x] Only one issue is addressed in this pull request
- [x] Labels from the issue that this PR is fixing are added to this pull request
- [x] No unnecessary issues are included into this pull request.

PR #3736, opened 2026-01-20
Fix/polynomial autograd post merge

leomilano:
Fixes  suggestion in #3597 

## Description
After closing #3597 , it was pointed out by a community member that while the changes made sense in isolation, they broke the flow of the tutorial overall. In particular, the tutorial develops a number of examples around sin(x).  So, this is mostly a one-liner to revert to sin(x) instead of exp(x).

As agreed in the discussion in #3597 , I also suggest in comments experimentation with exp(x) or other functions with faster polynomial convergence. 

The learning rate had to be reduced for this by a factor 10 in order to converge with autograd. 

I also fixed a trivial documentation typo. 

## Checklist

cc @albanD @jbschlosser

### DIFF
--- a/beginner_source/examples_autograd/polynomial_autograd.py
+++ b/beginner_source/examples_autograd/polynomial_autograd.py
@@ -3,7 +3,10 @@ PyTorch: Tensors and autograd
 -------------------------------
 
 A third order polynomial, trained to predict :math:`y=\sin(x)` from :math:`-\pi`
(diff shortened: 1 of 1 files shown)


AlannaBurke:
looks good from a docs point of view, have someone approve the content.

leomilano:
Thank you both!  I wonder if @albanD or @jbschlosser could review. Thanks!

PR #3756, opened 2026-01-30
fix: replace torch.functional import with preferred torch.nn.functional

### LINKED ISSUE #3755: [BUG] - Incorrect import: torch.functional used instead of torch.nn.functional
### Add Link

https://github.com/pytorch/tutorials/blob/main/beginner_source/introyt/modelsyt_tutorial.py

### Describe the bug

The LeNet example currently imports `torch.functional` and aliases it as `F`:

    import torch.functional as F

However, `torch.functional` does not provide neural network functions such as
`relu` or `max_pool2d`. This causes runtime errors when calling:

    F.relu(...)
    F.max_pool2d(...)

The correct import should be:

    import torch.nn.functional as F

This aligns with PyTorch’s functional API and matches official documentation
and tutorials.

### Describe your environment

- Google Collab

cc @albanD @jbschlosser

vinayaksodar:
Fixes #3755 

## Description
replace invalid torch.functional import with torch.nn.functionalv

## Checklist

cc @zou3519 @albanD @jbschlosser

### DIFF
--- a/beginner_source/introyt/modelsyt_tutorial.py
+++ b/beginner_source/introyt/modelsyt_tutorial.py
@@ -144,7 +144,7 @@ print(y)
 # We saw convolutional layers in action in LeNet5 in an earlier video:
 # 
 
-import torch.functional as F
+import torch.nn.functional as F
 
 
 class LeNet(torch.nn.Module):


zou3519:
@svekars is there CI on this? I would have expected that we caught this before. If the answer is that there is no CI on this, then I would check this out locally and see if it actually runs. NB: I'm going to be sporadic for a couple of days

PR #3763, opened 2026-02-10
Ray Data + Ray Train tutorial

crypdick:
Adds a new tutorial showcasing streaming data loading using Ray Data with Ray Train. Will move out of draft once I get reviews from the Data and Train teams.

cc @pcmoritz @robertnishihara @matthewdeng @richardliaw @akshay-anyscale

### DIFF
--- a/beginner_source/distributed_training_with_ray_tutorial.py
+++ b/beginner_source/distributed_training_with_ray_tutorial.py
@@ -107,7 +107,6 @@ print(f"Dataset schema:\n{train_ds.schema()}")
 # This means that the dataset has one column called ``text`` and it is a string.
 #
 # Inspect raw data
-#
 # ~~~~~~~~~~~~~~~~
 #
 # Use ``take(n)`` to fetch a small number of rows for inspection.


beginner_source/distributed_training_with_ray_tutorial.py
```diff
@@ -0,0 +1,719 @@
...
+#
+#    {'train_loss': 7.0646, 'val_loss': 7.6051, 'epoch': 4,
+#     'epoch_time_sec': 12.34, 'epoch_tokens': 20480, 'tokens_per_sec': 1759.8}
+#
+# The per-worker logs show training loss, validation loss, and throughput
+# metrics for each epoch. With random weights and only a few steps, expect
+# a high loss (~10–11).
+
+###############################################################################
+# Checkpointing
```
crypdick, line 491: debating whether I should slim this section down, given that I'm not checkpointing in the tutorial

beginner_source/distributed_training_with_ray_tutorial.py
```diff
@@ -0,0 +1,719 @@
...
+
+This tutorial shows how to distribute PyTorch training across multiple GPUs
+using Ray Train and Ray Data for scalable, production-ready model training.
+
+.. grid:: 2
+
+    .. grid-item-card:: :octicon:`mortar-board;1em;` You will learn how to:
+       :class-card: card-prerequisites
+
+       * Pre-train a GPT-2 (~124M-parameter) language model using PyTorch
```
svekars, line 15: should we provide a link on how to do it?

beginner_source/distributed_training_with_ray_tutorial.py
```diff
@@ -0,0 +1,719 @@
...
+       * Save and load distributed checkpoints.
+       * Scale from a single node to a multi-node cluster with minimal code changes.
+       * Optimize cost and performance with heterogeneous clusters.
+       * Monitor training with the Ray dashboard.
+
+    .. grid-item-card:: :octicon:`list-unordered;1em;` Prerequisites
+       :class-card: card-prerequisites
+
+       * PyTorch v2.9+.
+       * Ray Train (``ray[train]``) v2.52.1+.
```
svekars, line 28: In a Colab, scenario, will users need to install it? If yes, can add a note about that.

crypdick: Yep, I did so here: https://github.com/pytorch/tutorials/pull/3763/changes#diff-c45dd3d8b102a536c769a715a4694109a6e60923daf9fc3ba02fba8eda8d2943R49

beginner_source/distributed_training_with_ray_tutorial.py
crypdick: What do you mean?

beginner_source/distributed_training_with_ray_tutorial.py
justinvyu: Let's remove this section and point to the checkpointing user guide in Ray Docs. That way we can direct people to new features and avoid showing outdated apis (ex: the TorchTrainer.restore() API is deprecated).

beginner_source/distributed_training_with_ray_tutorial.py
```diff
@@ -0,0 +1,719 @@
...
+            "val_loss": round(avg_val_loss, 4),
+            "epoch": epoch,
+            "epoch_time_sec": round(epoch_elapsed, 2),
+            "epoch_tokens": train_tokens,
+            "tokens_per_sec": round(train_tokens / max(train_elapsed, 1e-6), 2),
+        }
+        ray.train.report(
+            metrics=metrics,
+            checkpoint=None,  # If we were checkpointing, we'd pass a Checkpoint here
+        )
```
justinvyu: we can just print if we're not checkpointing to simplify the script. these metrics don't get saved anywhere automatically. the user should report to wandb or something if they want to keep these.

beginner_source/distributed_training_with_ray_tutorial.py
```diff
@@ -0,0 +1,719 @@
...
+#        run_config=RunConfig(
+#            # Shared storage accessible from all nodes
+#            storage_path="s3://my-bucket/ray-checkpoints",
+#            checkpoint_config=CheckpointConfig(num_to_keep=2),
+#        ),
+#    )
+#
+# Ray Train automatically:
+#
+# * Launches workers across all available nodes.
```
justinvyu:
```suggestion
# * Launches workers across all available nodes, bringing up new nodes if needed in an autoscaling Ray cluster.
```

beginner_source/distributed_training_with_ray_tutorial.py
```diff
@@ -0,0 +1,719 @@
...
+
+    if num_chunks == 0:
+        return {"input_ids": [], "labels": []}
+
+    tokens_array = np.array(all_tokens, dtype=np.int64).reshape(num_chunks, chunk_len)
+    input_ids = tokens_array[:, :-1]
+    labels = tokens_array[:, 1:]
+    return {
+        "input_ids": input_ids,
+        "labels": labels,
```
justinvyu: nit: no need to output the shifted version from the map batches, you can output just the input ids and then take a shifted view of the same tensor in the training code to get the labels.

I think the HF model might also do the label shifting internally already if you only pass input ids? should double check that

beginner_source/distributed_training_with_ray_tutorial.py
```diff
@@ -0,0 +1,719 @@
...
+# * Learned how heterogeneous clusters let you run data preprocessing
+#   on CPU nodes and training on GPU nodes for cost and performance
+#   optimization.
+# * Learned about Ray Train's **fault tolerance** mechanisms for
+#   production training jobs.
+# * Monitored training with the Ray dashboard.
+#
+# Ray Train handles the complexity of distributed systems, gradient
+# synchronization, and resource allocation so that you can focus on
+# your model and data.
```
justinvyu:
```suggestion
```

beginner_source/distributed_training_with_ray_tutorial.py
crypdick: good call, and I'll double check the shifting behavior

beginner_source/distributed_training_with_ray_tutorial.py
crypdick: Makes sense @justinvyu . I rewrote it without code, but kept the section because I wanted to discuss some of the nice checkpointing features. Lmk what you think.

beginner_source/distributed_training_with_ray_tutorial.py
crypdick: You were right, the labels are already shifted internally, so no need to shift. I've fixed the bug. Nice catch

beginner_source/distributed_training_with_ray_tutorial.py
crypdick: I've enabled checkpointing now

beginner_source/distributed_training_with_ray_tutorial.py
```diff
@@ -0,0 +1,668 @@
...
+# .. code-block:: text
+#
+#    Column  Type
+#    ------  ----
+#    text    string
+#
+# This means that the dataset has one column called ``text`` and it is a string.
+#
+# Inspect raw data
+#
```
svekars: There is an issue with rendering this heading: https://docs-preview.pytorch.org/pytorch/tutorials/3763/beginner/distributed_training_with_ray_tutorial.html#load-the-dataset-with-ray-data
```suggestion
```

beginner_source/distributed_training_with_ray_tutorial.py
```diff
@@ -0,0 +1,668 @@
...
+# Each row is a dictionary with the column names as keys.
+
+print("--- Raw data sample ---")
+sample = train_ds.take(2)
+for i, row in enumerate(sample):
+    text_preview = (row["text"][:120] + "...") if len(row["text"]) > 120 else row["text"]
+    print(f"  Row {i}: {text_preview!r}")
+
+###############################################################################
+# You'll see output like:
```
svekars:
```suggestion
# You'll see output like this:
```

PR #3769, opened 2026-02-13
[Build] Pin onnx-related packages

titaiwangms:
Fix https://github.com/pytorch/tutorials/actions/runs/21965226194/job/63555988496?pr=3761#step:9:4434

## Description

Pin ONNX-related packages: onnx and onnxscript to avoid potential regression (for example, 0.6.1 onnxscript)

## Checklist


titaiwangms:
cc @justinchuby Do you think we need to specify onnx-ir?

justinchuby:
> cc @justinchuby Do you think we need to specify onnx-ir?

I think this is ok. Thanks!

PR #3770, opened 2026-02-16
Fix outdated torch_logs tutorial by removing CUDA device check

ManasVardhan:
## Fix

The `torch_logs.py` tutorial was wrapped in a CUDA device capability check:

```python
if torch.cuda.get_device_capability() < (7, 0):
    print("Skipping because torch.compile is not supported on this device.")
else:
    # ... entire tutorial ...
```

When built on CI without a compatible GPU, this caused the rendered docs to show only **"Skipping because torch.compile is not supported on this device."** instead of actual compilation log output.

## Changes

- Removed the CUDA device capability check that gated the entire tutorial
- Use dynamic device selection (`cuda` if available, otherwise `cpu`)
- `torch.compile` works on CPU, so the tutorial now produces meaningful log output regardless of build environment

Fixes pytorch/pytorch#137285

cc @PaliC @mlazos


ManasVardhan:
I've signed the CLA. Could you re-run the CLA check? Thanks!

mlazos:
Hi can you just also allow this on cpu - torch.compile is not supported on cuda devices lower than that version and users can hit strange errors if it isn't gated.

ManasVardhan:
Hey, good call! Updated the gate so it only skips on CUDA devices with capability < 7.0. If there's no CUDA or if the device is good enough, it just picks the right device (CPU or CUDA) and runs normally. Should work fine on CPU-only machines now too.

ManasVardhan:
Hey @mlazos, friendly ping -- I pushed the fix you requested (allowing CPU alongside the CUDA gate) back on Mar 3. Would you mind taking another look when you get a chance? Thanks!

PR #3792, opened 2026-03-10
Add torch.compile escape hatches to programming model doc

ydwu4:
As titled.

cc @williamwen42 @msaroufim @anijain2305 @angelayi

### DIFF
--- a/intermediate_source/torch_compile_tutorial.py
+++ b/intermediate_source/torch_compile_tutorial.py
@@ -398,6 +398,99 @@ bar_fixed(inp1, -inp2)
 # for tips on how to work around graph breaks.
 
 ######################################################################
+# Escape Hatches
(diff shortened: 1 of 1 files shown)


williamwen42 (commented):
Sorry for the late review! I think this should go into the primary programming model docs, not the tutorial.

e.g. mention `leaf_function` as an escape hatch here: https://docs.pytorch.org/docs/stable/user_guide/torch_compiler/compile/programming_model.fullgraph_true.html#strategy-2-pure-functions-can-always-be-compiled-via-an-escape-hatch

And then write a new page for leaf_function such as https://docs.pytorch.org/docs/stable/user_guide/torch_compiler/compile/programming_model.dynamo_nonstrict_trace.html

PR #3796, opened 2026-03-13
Fix invalid escape sequence in saving_loading_models tutorial

### LINKED ISSUE #3790: Feedback about Saving and Loading Models
There is the following issue on this page: https://docs.pytorch.org/tutorials/beginner/saving_loading_models.html#saving-and-loading-models
Great page for beginners, to understand how to save and load torch models...

pranay-3108:
Fixes #3790

While running `saving_loading_models.py`, Python raised a SyntaxWarning due to an invalid escape sequence in the documentation string

this PR removes the unnecessary escape character to avoid the warning in newer Python versions. The change does not affect the tutorial functionality

Fixes #ISSUE_NUMBER

## Checklist

### DIFF
--- a/beginner_source/saving_loading_models.py
+++ b/beginner_source/saving_loading_models.py
@@ -18,7 +18,7 @@ functions to be familiar with:
    objects can be saved using this function.
 
 2) `torch.load <https://pytorch.org/docs/stable/torch.html?highlight=torch%20load#torch.load>`__:
-   Uses `pickle <https://docs.python.org/3/library/pickle.html>`__\ ’s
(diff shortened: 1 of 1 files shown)


pranay-3108:
thanks for pointing that out.

I might be wrong, but while looking at the logs it seems like the failure is happening during the dataset download in jenkins/download_data.py`, 
where the MNIST file checksum doesn’t match the expected value

Since this PR only removes an invalid escape sequence in `saving_loading_models.py`, 
I don’t think it should affect that part 
Just sharing my small observation in case it helps

if I misunderstood something please tell me it improve me

PR #3826, opened 2026-04-13
Update CIFAR10 tutorial device selection for CUDA, MPS, and CPU

amit-chaubey:
## Description
This PR updates `beginner_source/blitz/cifar10_tutorial.py` to make device selection explicit across CUDA, MPS, and CPU.

### Changes
- Select device in this order:
  - `cuda:0` if CUDA is available
  - `mps` if Apple MPS backend is available
  - `cpu` otherwise
- Update nearby tutorial comments to refer to generic `device` usage instead of CUDA-only wording.

### Why
The tutorial currently frames this section as CUDA-specific. This change keeps the same tutorial flow while making guidance clearer for Apple Silicon and CPU-only users.

## Checklist
- [ ] The issue that is being fixed is referred in the description (no exact matching open issue linked for this change)

cc @subramen @albanD @jbschlosser

### DIFF
--- a/beginner_source/blitz/cifar10_tutorial.py
+++ b/beginner_source/blitz/cifar10_tutorial.py
@@ -299,20 +299,24 @@ for classname, correct_count in correct_pred.items():
 # Just like how you transfer a Tensor onto the GPU, you transfer the neural
 # net onto the GPU.
 #
-# Let's first define our device as the first visible cuda device if we have
-# CUDA available:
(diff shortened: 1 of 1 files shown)


svekars:
@claude

amit-chaubey:
Thanks! @claude Updated to use torch.accelerator (same one-liner style as quickstart_tutorial.py), adjusted the surrounding prose so it stays accurate when device is "cpu", and kept num_workers=0 for direct script runs on macOS/Windows. Pushed in the latest commit.

PR #3838, opened 2026-04-23
Fix: #3833 Feedback about Multi GPU training with DDP

### LINKED ISSUE #3833: Feedback about Multi GPU training with DDP
There is the following issue on this page: https://docs.pytorch.org/tutorials/beginner/ddp_series_multigpu.html

At the end of the page, it says "Here's what the code looks like: " but it does not actually show the full code. The full code would be helpful to have, since otherwise it is somewhat unclear what code needs to be run before the call to mp.spawn(), and what can be run after.

samithcsachi:
Fixes #3833 Feedback about Multi GPU training with DDP

## Description

At the end of the page, it says "Here's what the code looks like: " but it does not actually show the full code. The code was not visible because there was no proper spacing. Have corrected the spacing. 

## Checklist

- [x ] The issue that is being fixed is referred in the description (see above "Fixes #3833")
- [x ] Only one issue is addressed in this pull request
- [x ] Labels from the issue that this PR is fixing are added to this pull request

cc @svekars @sekyondaMeta @AlannaBurke

### DIFF
--- a/beginner_source/ddp_series_multigpu.rst
+++ b/beginner_source/ddp_series_multigpu.rst
@@ -202,7 +202,6 @@ Running the distributed training job
 Here's what the code looks like:
 
 .. code-block:: python
-
    
    def main(rank, world_size, total_epochs, save_every):
       ddp_setup(rank, world_size)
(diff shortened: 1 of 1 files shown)


wconstab:
looks fixed to me:
image

@svekars can you approve the workflows so this can land? it doesn't let me.

wconstab (approved):

wconstab:
@samithcsachi can you fix the lint issues and resubmit?

PR #3848, opened 2026-05-01
Fix torch_logs tutorial to support CPU fallback and CUDA capability check

Sip4818:
### Problem
The current tutorial wraps the entire example in a CUDA capability check:

```python
if torch.cuda.get_device_capability() < (7, 0):
    print("Skipping because torch.compile is not supported on this device.")
else:
    # tutorial code
```
This causes the tutorial to output only a "Skipping..." message on:

- CPU-only environments  
- CI builds without compatible GPUs  

As a result, users do not see any actual `torch.compile` logging output.

---

### Solution
- Added device selection logic:
  - Use CUDA if available and supported (compute capability >= 7.0)
  - Otherwise, fall back to CPU  
- Ensured the tutorial runs regardless of the environment  
- Preserved a warning message when CUDA is available but unsupported  

---

### Result
The tutorial now:

- Runs on CPU-only systems  
- Runs on supported CUDA devices  
- Produces meaningful logging output in all environments  

Addresses pytorch/pytorch#137285


Sip4818:
CLA signed. Happy to make any changes if needed.

sekyondaMeta:
@claude

Sip4818:
Fixed lint issues (whitespace and formatting). Ready for re-review.

mlazos (approved):

Sip4818:
Rebased onto latest main. The CI checks haven't triggered - @mlazos could you please take a look?  Thanks.

PR #3860, opened 2026-05-05
Revise Captum setup and remove visualization code

### LINKED ISSUE #3859: Fix broken Captum tutorial — remove retired Insights section, update requirements
## Description

The [Model Understanding with Captum](https://docs.pytorch.org/tutorials/beginner/introyt/captumyt.html) tutorial has critical breakage: the Captum Insights section uses APIs that were fully removed from Captum in Feb 2026. The tutorial also has stale dependency requirements.

### Critical — broken code

- **Remove Captum Insights section** (lines ~400-499): `AttributionVisualizer`, `Batch`, `ImageFeature`, and `visualizer.render()` were deprecated in Captum v0.8.0 (Feb 2025) and fully removed from the codebase (Feb 2026). This entire section will crash on current Captum.
- **Remove `from captum.insights import AttributionVisualizer, Batch`**
- **Remove `from captum.insights.attr_vis.features import ImageFeature`**

### Deprecated APIs

- **`transforms.ToTensor()` → `v2.ToImage() + v2.ToDtype(torch.float32, scale=True)`**: `ToTensor()` is deprecated since torchvision 0.18.

### Stale requirements

- **Python version**: `3.6` → `3.9` or higher (Python 3.6 is EOL since Dec 2021; PyTorch 2.x requires 3.9+)
- **PyTorch version**: Remove `1.2 or higher` floor — it's ancient
- **TorchVision version**: Remove `0.6 or higher` floor
- **Flask dependency**: Remove entirely — Flask was only needed for Captum Insights
- **Matplotlib pin**: Remove `matplotlib==3.3.4` / `matplotlib=3.3.4` pin from both pip and conda install commands. The pin was for a Captum bug that was fixed long ago.
- **Conda install**: Switch to pip-only (Captum v0.8.0 was the last conda release on the PyTorch channel)

### Nice to have

- Consider replacing the removed Insights section with an **LLM Prompt Attribution** example — this is the headline feature of Captum v0.8.0 and would modernize the tutorial to reflect current AI trends.

## Files

- [`beginner_source/introyt/captumyt.py`](https://github.com/pytorch/tutorials/blob/main/beginner_source/introyt/captumyt.py)

cc @subramen

payal101:
Updated environment requirements and installation instructions for Captum. Removed outdated visualization section and related code.

Version for pytorch not written (instead written reccomended latest version)

Fixes #3859

## Checklist

- [ x] The issue that is being fixed is referred in the description 
- [x ] Only one issue is addressed in this pull request
- [ x] Labels from the issue that this PR is fixing are added to this pull request

### DIFF
--- a/beginner_source/introyt/captumyt.py
+++ b/beginner_source/introyt/captumyt.py
@@ -97,23 +97,15 @@ Installation
 
 Before you get started, you need to have a Python environment with:
 
--  Python version 3.6 or higher
--  For the Captum Insights example, Flask 1.1 or higher and Flask-Compress
-   (the latest version is recommended)
--  PyTorch version 1.2 or higher (the latest version is recommended)
--  TorchVision version 0.6 or higher (the latest version is recommended)
+-  Python version 3.9 or higher
+-  PyTorch version  (the latest version is recommended)
+-  TorchVision version (the latest version is recommended)
 -  Captum (the latest version is recommended)
--  Matplotlib version 3.3.4, since Captum currently uses a Matplotlib
-   function whose arguments have been renamed in later versions
 
-To install Captum in an Anaconda or pip virtual environment, use the
-appropriate command for your environment below:
 
-With ``conda``:
+To install Captum in  pip virtual environment, use the
+command for your environment below:
 
-.. code-block:: sh
-
-    conda install pytorch torchvision captum flask-compress matplotlib=3.3.4 -c pytorch
 
 With ``pip``:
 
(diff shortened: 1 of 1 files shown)


gtsitsik:
I think there may be a few issues in the current diff.

The tutorial currently imports `torchvision.transforms as transforms`, so the existing `transforms.ToTensor()` is using the v1 transform API. My understanding is that the deprecation warning applies to `torchvision.transforms.v2.ToTensor`, not necessarily to `torchvision.transforms.ToTensor`.

Also, unless `v2` is imported elsewhere, `v2.ToImage()` and `v2.ToDtype(...)` would be undefined here.

Finally, I think the replacement should likely not use `+` between transforms. If the intent is to use the v2 replacement, then according to [the documentation](https://docs.pytorch.org/vision/main/generated/torchvision.transforms.v2.ToTensor.html) this should likely be written as:

```python
v2.Compose([
    v2.ToImage(),
    v2.ToDtype(torch.float32, scale=True),
])

svekars:
the issue is assigned to [thesteve0](https://github.com/thesteve0)

PR #3897, opened 2026-05-13
fix description of equation in polynomial_autograd.py example

### LINKED ISSUE #3632: Feedback about PyTorch: Tensors and autograd
There is the following issue on this page: https://docs.pytorch.org/tutorials/beginner/examples_autograd/polynomial_autograd.html

The paratext and previous/subsequent excercises imply this should learn sin(x), but the y= is generated as torch.exp(x), with the exponential's Taylor expantion.

telamonian:
fixes #3632

## Description
In the polynomial_autograd.py example, the equation that a + b*x + c*x^2 + d*x^3 converges to changed in a recent commit, from y=sin(x) to y=e^x. However, they didn't change the text description of the equation. This commit fixes that.

## Checklist

### DIFF
--- a/beginner_source/examples_autograd/polynomial_autograd.py
+++ b/beginner_source/examples_autograd/polynomial_autograd.py
@@ -2,8 +2,8 @@ r"""
 PyTorch: Tensors and autograd
 -------------------------------
 
-A third order polynomial, trained to predict :math:`y=\sin(x)` from :math:`-\pi`
-to :math:`\pi` by minimizing squared Euclidean distance.
+A third order polynomial, trained to predict :math:`y=e^x` from :math:`-1`
+to :math:`1` by minimizing squared Euclidean distance.
 
 This implementation computes the forward pass using operations on PyTorch
 Tensors, and uses PyTorch autograd to compute gradients.
--- a/beginner_source/examples_autograd/polynomial_autograd.py
+++ b/beginner_source/examples_autograd/polynomial_autograd.py
@@ -14,7 +14,6 @@ Tensor that has ``x.requires_grad=True`` then ``x.grad`` is another Tensor
 holding the gradient of ``x`` with respect to some scalar value.
 """
 import torch
-import math
 
 # We want to be able to train our model on an `accelerator <https://pytorch.org/docs/stable/torch.html#accelerators>`__
 # such as CUDA, MPS, MTIA, or XPU. If the current accelerator is available, we will use it. Otherwise, we use the CPU.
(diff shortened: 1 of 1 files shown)


telamonian:
After I saw #3899 I have now also removed the same unused lines of code that @gtsitsik removed in his version of this same PR.

Also, I made up some code for plotting the fit/visualizing the "goodness" of fit:

```python

def fit(x):
    return a.item() + b.item() * x + c.item() * x**2 + d.item() * x**3

y_pred = [fit(_x) for _x in x]

fig, ax = plt.subplots(figsize=(8, 5))
ax.plot(x, y_pred, label="predicted")
ax.plot(x, y, label="real")
ax.set_xlabel("x")
ax.set_ylabel("$e^x$")
ax.legend()
plt.show()
```

Untitled

Should I add this plotting code to the tutorial, or do you guys want to keep `matplotlib` out of the tutorials?

beginner_source/examples_autograd/polynomial_autograd.py
```diff
@@ -39,7 +38,6 @@
 c = torch.randn((), dtype=dtype, requires_grad=True)
 d = torch.randn((), dtype=dtype, requires_grad=True)
 
-initial_loss = 1.
 learning_rate = 1e-5
```
martinptm, line 41:
```suggestion
learning_rate = 1e-6
```

beginner_source/examples_autograd/polynomial_autograd.py
martinptm, line 30:
```suggestion
y = torch.sin(x)
```

beginner_source/examples_autograd/polynomial_autograd.py
```diff
@@ -14,7 +14,6 @@
 holding the gradient of ``x`` with respect to some scalar value.
 """
 import torch
-import math
```
martinptm, line 17: Keep import to use `math.pi` below.

beginner_source/examples_autograd/polynomial_autograd.py
martinptm, line 29:
```suggestion
x = torch.linspace(-math.pi, math.pi, 2000, dtype=dtype)
```

beginner_source/examples_autograd/polynomial_autograd.py
```diff
@@ -2,8 +2,8 @@
 PyTorch: Tensors and autograd
 -------------------------------
 
-A third order polynomial, trained to predict :math:`y=\sin(x)` from :math:`-\pi`
```
martinptm, line 5: Keep original description.

martinptm (commented):
Hello @telamonian, I have just stumbled across the mismatch in the tutorial that you address here. It does break the flow of the tutorial and I appreciate that you already prepared a fix.
However, in the context of the previous tutorial sections, I believe it would be more consistent to keep fitting a sin-curve in this section as well, i.e. fix the implementation to match the description instead of the other way around. 
If this is done, it is important for the stability of the fit to keep the lower `learning_rate` too.

gtsitsik:
@martinptm you may want to check the discussion in #3899 as well.

PR #3899, opened 2026-05-13
Fix inconsistencies in polynomial autograd tutorial

### LINKED ISSUE #3632: Feedback about PyTorch: Tensors and autograd

gtsitsik:
## Changes

- Use sine instead of exponential to match the tutorial text.
- Update the learning rate and iteration count to match `beginner_source/examples_tensor/polynomial_tensor.py` which is exactly the same tutorial but with manual gradient computations.
- Correct the description of the `x.grad` attribute.
- Remove redundant lines of code.
- Fix typo. 

## Note
PR #3663 and PR #3897 recommended updating the tutorial text to show the exponential function instead of updating the code to use a sine. However, I recommend using sine everywhere since both the numpy and non-autograd pytorch polynomial tutorials use a sine. 

Fixes #3632 

## Checklist

### DIFF
--- a/beginner_source/examples_autograd/polynomial_autograd.py
+++ b/beginner_source/examples_autograd/polynomial_autograd.py
@@ -10,11 +10,12 @@ Tensors, and uses PyTorch autograd to compute gradients.
 
 
 A PyTorch Tensor represents a node in a computational graph. If ``x`` is a
-Tensor that has ``x.requires_grad=True`` then ``x.grad`` is another Tensor
-holding the gradient of ``x`` with respect to some scalar value.
+Tensor with ``x.requires_grad=True``, then after backpropagating from a scalar
+``loss``, ``x.grad`` is another Tensor storing the gradient of that scalar loss
+with respect to ``x``.
 """
+
 import torch
-import math
 
 # We want to be able to train our model on an `accelerator <https://pytorch.org/docs/stable/torch.html#accelerators>`__
 # such as CUDA, MPS, MTIA, or XPU. If the current accelerator is available, we will use it. Otherwise, we use the CPU.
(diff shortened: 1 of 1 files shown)


telamonian:
Dueling fixes (full disclosure I'm the author of #3897)!

@gtsitsik I don't know if you saw #3597, the PR that originally (and fairly recently) changed the tutorial from using sin(x) to e^x, but they gave a pretty reason for doing so. Basically, the trained Taylor expansion given in the tutorial is able to converge nicely to e^x, but does not converge that nicely to sin(x). You can see the problem in the following plots:

Untitled

Untitled-1

The problem is fundamentally a mathematical one; a third order polynomial is never a good fit for sin(x), while at least over the domain [-1, 1] a third order polynomial is a good fit for e^x.

Therefore I propose that Pytorch keeps the tutorial using e^x while also keeping @gtsitsik's fixes of typos and removal of unused code. I've updated #3897 with these fixes.

gtsitsik:
@telamonian by the way I noticed your PR after I created mine.

Regarding using an exponential to get a better approximation, I don't think it really matters for the purposes of this tutorial. But I also don't mind if it remains an exponential. 

However, I do think that the 3 tutorials should be consistent with each other as they seem to be aiming to demonstrate the same thing, but via different approches. That is, 

- manual gradient calculation using numpy 
- manual gradient calculation using pytorch 
- automatic gradient calculation using pytorch 

All 3 should either use a sine or an exponential because otherwise they will be inconsistent with each other and their pedagogical quality becomes slightly off.

telamonian:
> All 3 should either use a sine or an exponential because otherwise they will be inconsistent with each other

That I can definitely agree with. But it doesn't all have to be changed in one big PR. I think the best thing to do for now would be to pull in #3897, close this PR, and then fix the equations/descriptions for the other examples in future PRs.

But we'll have to wait and see what any of the actual maintainers think

gtsitsik:
@telamonian that is fine with me. However, in my PR I include another fix that you didn't include in yours. Specifically, the tutorial incorrectly claims that `grad.x` holds the gradient of `x` with respect to some scalar (the loss), when in reality it is the gradient of the loss with respect to `x`.

PR #3924, opened 2026-06-17
Fix torch_logs tutorial to support CPU-only environments (#137285)

### LINKED ISSUE #137285: Docs are little bit outdated for torch logs
### 📚 The doc issue

At the end of setup here I assume the output should be some log describing the program rather than saying that the device does not support torch.compile.

https://pytorch.org/tutorials/recipes/torch_logs.html#setup

### Suggest a potential alternative/fix

Run the program in the tutorial and update the doc :)

yyouretoast:
Fixes https://github.com/pytorch/pytorch/issues/137285

The `torch_logs` tutorial was gated behind a hard CUDA capability check (`get_device_capability() < (7, 0)`) that printed a skipping message and exited. This caused the tutorial to be skipped during CPU-only documentation builds, resulting in the page rendering as "skipped" on the official PyTorch website.

Since `torch.compile` is fully supported on CPU (using the C++ Inductor backend) and generates meaningful compilation logging outputs, this PR refactors the tutorial to allow CPU execution. It dynamically selects CUDA if available and the compute capability is `>= 7.0` (to avoid compile crashes on older GPUs), otherwise falling back to CPU.

### Changes
- Replaced the hard CUDA check with dynamic device selection:
  ```python
  device = "cuda" if torch.cuda.is_available() and torch.cuda.get_device_capability() >= (7, 0) else "cpu"


yyouretoast:
Hi! Just bumping this PR. 

Could a maintainer please approve the `pytorch_tutorial_build_manager` workflow run so the checks can build? Let me know if any changes or additional testing are needed for this CPU-only fallback fix. Thank you!

PR #3950, opened 2026-08-09
Fix torch_logs tutorial to run on CPU when CUDA unavailable

### LINKED ISSUE #137285: Docs are little bit outdated for torch logs

ajaykesarwani:
Fixes pytorch/pytorch#137285

The `torch_logs` tutorial was skipping execution on machines without a
modern CUDA GPU (including the Sphinx doc build server), resulting in
"Skipping because torch.compile is not supported on this device."
being shown in the documentation instead of actual log output.

**Changes:**
- Removed the hard `torch.cuda.get_device_capability()` check
- Added `device = "cuda" if torch.cuda.is_available() else "cpu"` fallback
- Fixed indentation of code that was inside the old `else:` block
- Updated the `.. note::` to mention that CUDA is recommended but not required


ajaykesarwani:
Closing this! I didn't notice the linked issue had actually been closed over a month ago and already fixed in another PR. Sorry for the noise!

